PR #3, opened 2021-09-22
Additional repository setup

alexander-demichev:
Add `make unit|images|push` tasks, dockerfiles and a task for verifying vendor content

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+IMG ?= controller:latest
+
 all: build
 
 verify: fmt vet 
@@ -34,3 +36,13 @@ vendor:
 	go mod tidy
 	go mod vendor
 	go mod verify
+
+# Build the docker image
+.PHONY: image
+image:
+	docker build -t ${IMG} .
+
+# Push the docker image
+.PHONY: push
+push:
+	docker push ${IMG}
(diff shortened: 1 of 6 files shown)


elmiko (commented):
this generally looks good to me, i'm not so familiar with the envtest stuff though.
/lgtm

alexander-demichev:
@elmiko I copied envtest setup from cccmo

lobziik:
Hm... @alexander-demichev can you please explain, why you want two different dockerfiles there?

Makefile
```diff
@@ -1,16 +1,23 @@
+IMG ?= controller:latest
+
 all: build
 
-verify: fmt vet 
+verify-%:
+	make $*
+	./hack/verify-diff.sh
```
JoelSpeed, line 7: Just a note that in the CI, I think we need to run `verify-fmt` and `verify-vet` to make sure that once those tools have run, there's no diff left over

elmiko:
> Hm... @alexander-demichev can you please explain, why you want two different dockerfiles there?

iiuc, the `Dockerfile` allows community build, whereas `Dockerfile.rhel` is restricted to people with RHEL subscription or our CI systems.

alexander-demichev:
@elmiko you are right, dockerfiles reference different images. One is publicly available for anyone outside of red hat, second is rhel based

PR #14, opened 2021-11-02
Unit tests for isCAPIFeatureGateEnabled

asalkeld:


alexander-demichev (approved):
/approve

alexander-demichev (approved):
/lgtm

elmiko (commented):
i think the only confusing thing here is using the zero value for the `wantErr`, but it's not a blocker for me.
/lgtm

PR #18, opened 2021-11-23
Remove go-get-tool.sh and use tools module

ardaguclu:
go-get-tool.sh runs `go get` command to install tools binaries which
will be deprecated and this PR instead uses tools module for
binaries to be compatible with the upstream operator.


ardaguclu:
cc: @asalkeld

alexander-demichev (commented):
Thanks for the PR! You might want to include vendor directory for tools because the binaries won't build in CI container if we try it. We already had this issue here https://github.com/openshift/cluster-api/pull/146

ardaguclu:
> Thanks for the PR! You might want to include vendor directory for tools because the binaries won't build in CI container if we try it. We already had this issue here [openshift/cluster-api#146](https://github.com/openshift/cluster-api/pull/146)

My first attempt was to vendor tools but since I did not know you have already discussed it, I removed. I'll update this patch with hack/tools vendoring. Thanks @alexander-demichev

Makefile
```diff
@@ -41,11 +44,11 @@ lint: $(GOLANGCI_LINT)
...
 # Download golangci-lint locally if necessary
-$(GOLANGCI_LINT):
-	$(PROJECT_DIR)/hack/go-get-tool.sh go-get-tool $(GOLANGCI_LINT) github.com/golangci/golangci-lint/cmd/golangci-lint@v1.41.1
+$(GOLANGCI_LINT): $(TOOLS_DIR)/go.mod # Build golangci-lint from tools folder.
+	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
 
-$(KUSTOMIZE):
-	$(PROJECT_DIR)/hack/go-get-tool.sh go-get-tool $(KUSTOMIZE) sigs.k8s.io/kustomize/kustomize/v3@v3.9.4
+$(KUSTOMIZE): $(TOOLS_DIR)/go.mod
+	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/kustomize sigs.k8s.io/kustomize/kustomize/v3
```
alexander-demichev, line 51:
```suggestion
	cd $(TOOLS_DIR); go build -tags=tools -mod=readonly -o $(BIN_DIR)/kustomize sigs.k8s.io/kustomize/kustomize/v3
```

Makefile
```diff
@@ -41,11 +44,11 @@ lint: $(GOLANGCI_LINT)
 	( GOLANGCI_LINT_CACHE=$(PROJECT_DIR)/.cache $(GOLANGCI_LINT) run )
 
 # Download golangci-lint locally if necessary
-$(GOLANGCI_LINT):
-	$(PROJECT_DIR)/hack/go-get-tool.sh go-get-tool $(GOLANGCI_LINT) github.com/golangci/golangci-lint/cmd/golangci-lint@v1.41.1
+$(GOLANGCI_LINT): $(TOOLS_DIR)/go.mod # Build golangci-lint from tools folder.
+	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
```
alexander-demichev, line 48:
```suggestion
	cd $(TOOLS_DIR); go build -tags=tools -mod=readonly -o $(BIN_DIR)/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
```
I've just noticed that setting `-mod=readonly` works too, so maybe we can avoid vendoring

alexander-demichev (approved):
/approve

asalkeld (commented):
/lgtm

PR #19, opened 2021-11-23
Extract provider versions from json file in import-assets

ardaguclu:
This PR extract provider versions from file instead hardcoded in import-assets utility. It would be better to manage resources in json file. In addition to that, although it is not being used, providerFilter can be used if import-assets is executed directly by user to filter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,5 +29,5 @@ This command does 2 main things:
    b. place provider rbac resources in /manifests
    c. place all other resources in /assets/providers as configmaps (to be consumed by capi-operator)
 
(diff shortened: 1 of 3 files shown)


ardaguclu:
cc: @alexander-demichev

alexander-demichev (approved):
/approve

ardaguclu:
@asalkeld could you please have a look at?, thanks.

asalkeld:
can you also update the readme?

ardaguclu:
> can you also update the readme?

Done

PR #20, opened 2021-11-24
Add tech preview annotation to manifests

alexander-demichev:
Add tech preview annotation to manifests, after this PR https://github.com/openshift/cluster-version-operator/pull/694/files#diff-7a023c513e7d3dd1938909a860915f4541530097d5d25a4d743f3eabaad6fc95R227is merged, CVO will be responsible for creating manifests on tech preview clusters.


manifests/image-references
```diff
@@ -6,6 +6,10 @@ spec:
     from:
       kind: DockerImage
       name: registry.ci.openshift.org/openshift:cluster-capi-operator
+  - name: cluster-api
+    from:
+      kind: DockerImage
+      name: registry.ci.openshift.org/openshift:cluster-api
```
JoelSpeed, line 12: This image is definitely in payload right?

JoelSpeed (commented):
I expect there will be other things we can use this feature for that we are currently using the operator for, is the plan to extract those in a separate PR?

manifests/image-references
alexander-demichev: Looks like no, I thought it is at this moment

alexander-demichev: Now I remember, Clayton asked to call it `cluster-capi-controllers`. If `images` task passes in CI then it's fine.

alexander-demichev:
@JoelSpeed yes, I will do it in a separate PR, after CVO PR merges

manifests/image-references
```diff
@@ -6,6 +6,10 @@ spec:
     from:
       kind: DockerImage
       name: registry.ci.openshift.org/openshift:cluster-capi-operator
+  - name: cluster-capi-controllers
```
JoelSpeed, line 9: I don't see cluster-capi-controllers within the OCP build data, I would expect it needs to be here before we can merge an image reference: https://github.com/openshift/ocp-build-data/tree/openshift-4.10/images 

While this may work for CI, merging without proper build data could break the nightly build process

alexander-demichev: https://github.com/openshift/ocp-build-data/blob/openshift-4.10/images/ose-cluster-api.yml, there is a config for capi

alexander-demichev: I removed new image references from this PR, need to figure out what is wrong.

wking:
Seems like you're missing an annotation on at least one manifest, based on [this run][1]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-capi-operator/20/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1467934131437441024/artifacts/e2e-aws-capi-techpreview/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-b797587c5-7wcwb_cluster-version-operator.log | grep -o 'error running apply.*' | sort | uniq
error running apply for role "openshift-cluster-api/capg-leader-election-role" (187 of 823): namespaces "openshift-cluster-api" not found
error running apply for serviceaccount "openshift-cluster-api/capz-manager" (177 of 823): namespaces "openshift-cluster-api" not found
```

Or maybe you need to keep the namespace regardless of tech-preview-ness to support those resources for some non-tech-preview component?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-capi-operator/20/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1467934131437441024

jiajliu:
cc @jiajliu

alexander-demichev:
@wking I fixed all annotations, thanks. I think namespace should not be present in non-techpreview clusters, this might bring some confusion for users.

alexander-demichev:
/hold 
techpreview e2e is failing, I will post a fix in another PR

PR #22, opened 2021-12-07
Disable operator reconciliation

alexander-demichev:
This PR temporarily disables operator reconciliation. After CVO started managing all manifests, we need to refactor the operator. With the current implementation, it tries to install AWS provider and some CRDs are missing, this causes tests to fail.


elmiko (commented):
we talked about this during our architecture meeting, i think it makes sense for now.
/approve

PR #24, opened 2021-12-15
Revert "Merge pull request #20 from alexander-demichev/annotations"

dgoodwin:
This reverts commit 3d1c8939f88b4cbbf0d6f7abf638dc6f2f7a682f, reversing
changes made to d4e3c4fcbbb7bcf216d2a6b2269a3bc109a1ae8f.

The original pr in #20 looks to have broken payload delivery for the org, so per policy we need to revert as quickly as possible. Ideally you can then revert the revert, and layer in the fixed commits. However in this case the fix might be to the tests in the origin repo.

You can see whats taking down payloads here: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.10-e2e-aws-ovn-upgrade/1470831487283630080

```
fail [github.com/openshift/origin/test/extended/operators/operators.go:151]: unexpected operator status versions cluster-api:
[]v1.OperandVersion{v1.OperandVersion{Name:"operator", Version:"4.10.0-0.ci-2021-12-14-120025"}}
Expected
    <string>: 4.10.0-0.ci-2021-12-14-120025
to equal
    <string>: 4.10.0-0.ci-2021-12-14-185120
```

Essentially we have tests that assert all operators hanging on the ClusterVersion must reach the correct upgrade level, and I suspect that annotation you added is to explicitly not do that, so presumably the test needs an update.

I notice you don't have any upgrade e2e jobs on this repo so that is likely how it slipped in. You may not want upgrade support so that is reasonable, but it might be good to add one just to catch this kind of thing as you are part of the payload, and that does tie you into upgrades even if your desired behavior is a little different.

### DIFF
--- a/hack/import-assets/providers.go
+++ b/hack/import-assets/providers.go
@@ -153,7 +153,7 @@ func (p *provider) writeRBACComponentsToManifests(objs []unstructured.Unstructur
 		return err
 	}
 
-	fName := strings.ToLower("0000_30_cluster-api_" + p.providerTypeName() + "-" + p.name + "_03_rbac.yaml")
+	fName := strings.ToLower("0000_30_cluster-api-" + p.providerTypeName() + "-" + p.name + "_03_rbac.yaml")
 	return os.WriteFile(path.Join(manifestsPath, fName), ensureNewLine(combined), 0600)
 }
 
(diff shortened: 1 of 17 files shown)


alexander-demichev (approved):
/lgtm
/approve

PR #26, opened 2022-01-04
Bug 2037061: Add service account names to aws and gcp credentials request manifest

elmiko:
For enabling short-lived credentials in aws and gcp clusters using workload
 identity, we need ccoctl tool to know the kubernetes service account
 names from the credentials request manifest so that it can create
 aws and gcp service accounts that can be impersonated only by specific
 kubernetes service accounts.

ref:
https://github.com/openshift/machine-api-operator/commit/1f3a083fe50c0e80106ca336878725f142173aba


akhil-rane (commented):
/lgtm

Fedosin (approved):
/approve

PR #32, opened 2022-03-03
Add CAPA to supported providers

alexander-demichev:
- Add CAPA to provider-list.yaml
- Update CAPI version in provider-list.yaml
- Filter out credentials secret that comes with provider manifests because our credentials are managed by CCO


Fedosin (approved):
/approve
/lgtm

PR #34, opened 2022-03-08
Refactor import-assets

alexander-demichev:
Instead of iterating through objects multiple times, we can do it in one go.


elmiko (commented):
i like how you've brought the object processing together, i don't have any strong thoughts about changes
/lgtm

lobziik:
tests are reporting about bunch of alerts and crushes from cluster-capi-operator, @alexander-demichev can you take a look please, it might be connected with this changes

JoelSpeed (commented):
/approve

PR #35, opened 2022-03-09
Add cluster controller

alexander-demichev:
Add cluster controller that creates and reconciles cluster objects.


pkg/controllers/cluster/cluster_test.go
```diff
@@ -0,0 +1,68 @@
...
+		Expect(cluster.Spec.InfrastructureRef.APIVersion).To(Equal(infraGVK))
+		Expect(cluster.Spec.InfrastructureRef.Kind).To(Equal(infraClusterKind))
+		Expect(cluster.Spec.InfrastructureRef.Name).To(Equal(r.clusterName))
+		Expect(cluster.Spec.InfrastructureRef.Namespace).To(Equal(controllers.DefaultManagedNamespace))
+
+		Expect(cl.Delete(ctx, cluster)).To(Succeed())
+		Eventually(
+			apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(cluster.DeepCopy()), &awsv1.AWSCluster{})),
+			timeout,
+		).Should(BeTrue())
```
JoelSpeed, line 53: This block wont work, for eventually functions to work, you need to pass in an uncalled function for the eventually to call, you have passed is a called function so the value can never change
```suggestion
		Eventually(func() bool {
			return apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(cluster.DeepCopy()), &awsv1.AWSCluster{}))
		}, timeout).Should(BeTrue())
```

cmd/cluster-controller/main.go
```diff
@@ -0,0 +1,150 @@
...
+		os.Exit(1)
+	}
+
+	supportedProviders, err := util.ReadProvidersFile(*providerFile)
+	if err != nil {
+		klog.Error(err, "unable to get providers from file", "name", *providerFile)
+		os.Exit(1)
+	}
+
+	if err = (&cluster.ClusterReconciler{
```
JoelSpeed, line 113: Don't shadow the previous err declaration here
```suggestion
	if err := (&cluster.ClusterReconciler{
```

pkg/controllers/cluster/aws_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+		Expect(awsCluster.Annotations).To(HaveKey(clusterv1.ManagedByAnnotation))
+		Expect(awsCluster.Spec.Region).To(Equal(awsPlatformStatus.Region))
+		Expect(awsCluster.Status.Ready).To(BeTrue())
+
+		Expect(cl.Delete(ctx, awsCluster)).To(Succeed())
+		Eventually(
+			apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(awsCluster.DeepCopy()), &awsv1.AWSCluster{})),
+			timeout,
+		).Should(BeTrue())
```
JoelSpeed, line 58: You need to wrap the assertion in a function that Eventually can call
```suggestion
		Eventually(func() bool {
			return apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(awsCluster.DeepCopy()), &awsv1.AWSCluster{})),
		}, timeout).Should(BeTrue())
```

pkg/controllers/cluster/cluster.go
```diff
@@ -0,0 +1,128 @@
...
+		klog.Infof("Platform type %v is not supported. Skipping cluster reconciliation...", platformType)
+		if err := r.SetStatusAvailable(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	klog.Infof("Reconciling %v infrastucture cluster", platformType)
+	// Reconcile infrastructure cluster based on platform type.
+	infraClusterKind := ""
```
JoelSpeed, line 76: Nit: var infraClusterKind string

alexander-demichev:
@JoelSpeed all fixed

lobziik:
e2e-aws-capi-techpreview reporting bunch of alerts and crashes from cluster-capi-operator, @alexander-demichev can you please take a look, it might be connected with this changes.

go.mod
```diff
@@ -15,20 +15,22 @@ require (
 	k8s.io/apimachinery v0.23.0
 	k8s.io/client-go v0.23.0
 	k8s.io/component-base v0.23.0
-	k8s.io/klog/v2 v2.30.0
+	k8s.io/klog/v2 v2.40.1
 	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
+	sigs.k8s.io/cluster-api v1.1.1
 	sigs.k8s.io/cluster-api-operator v0.0.0-20220221190000-3f8fd689a585
+	sigs.k8s.io/cluster-api-provider-aws v1.3.0
```
lobziik, line 22: Should there be our fork?

go.mod
```diff
@@ -15,20 +15,22 @@ require (
 	k8s.io/apimachinery v0.23.0
 	k8s.io/client-go v0.23.0
 	k8s.io/component-base v0.23.0
-	k8s.io/klog/v2 v2.30.0
+	k8s.io/klog/v2 v2.40.1
 	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
+	sigs.k8s.io/cluster-api v1.1.1
```
lobziik, line 20: Should there be our fork?

lobziik:
Have some concerns which i feel need to be communcated:
1) Separate binary for this cluster controller. I don't feel that this is necessary here. Basically we have two separate binaries which looks into same resources. It might be combined into one because of performance matter. Is there any strong reasoning for having it separately?
2) Dependency on providers - it's kinda self descriptive. I would like to avoid this, because if causes drastic raise of maintenance effort then.

cmd/cluster-capi-operator/main.go
```diff
@@ -138,6 +140,32 @@ func main() {
 		klog.Error(err, "unable to create controller", "controller", "ClusterOperator")
 		os.Exit(1)
 	}
+
+	if err := (&cluster.ClusterReconciler{
+		ClusterOperatorStatusClient: operatorstatus.ClusterOperatorStatusClient{
+			Client:         mgr.GetClient(),
+			Recorder:       mgr.GetEventRecorderFor("cluster-capi-operator-cluster-controller"),
```
lobziik, line 147: Can we make this name more specific, like `cluster-resource-controller`

PR #36, opened 2022-03-11
Set a release branch for manifests

alexander-demichev:
Set a release branch for manifests, this should help with backports later and we can always see where manifests are coming from.


Fedosin (approved):
/approve

PR #38, opened 2022-03-15
Add network interfaces permissions for aws

alexander-demichev:
CAPA attaches security groups to network interfaces after instance creation, this requires new permissions. In MAPI we specify security groups on instance creation that's we don't have these permissions for MAPI.


JoelSpeed (commented):
/approve

PR #42, opened 2022-03-18
Add kubeconfig controller

alexander-demichev:
Add controller that creates a kubeconfig from operators service account. This kubeconfig is tied to cluster object will be used by CAPI to link nodes and machines.

### DIFF
--- a/pkg/controllers/kubeconfig/generate.go
+++ b/pkg/controllers/kubeconfig/generate.go
@@ -1,0 +1,59 @@
+package kubeconfig
+
+import (
+	"errors"
+
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/openshift/cluster-capi-operator/pkg/controllers"
+)
+
+type kubeconfigOptions struct {
+	token            []byte
+	caCert           []byte
+	apiServerEnpoint string
+	clusterName      string
+}
+
+func generateKubeconfig(options kubeconfigOptions) (*api.Config, error) {
+	if len(options.token) == 0 {
+		return nil, errors.New("token can't be empty")
+	}
+
+	if len(options.caCert) == 0 {
+		return nil, errors.New("ca cert can't be empty")
+	}
+
+	if options.apiServerEnpoint == "" {
+		return nil, errors.New("api server endpoint can't be empty")
+	}
+
+	if options.clusterName == "" {
+		return nil, errors.New("cluster name can't be empty")
+	}
+
+	userName := "cluster-capi-operator"
+	kubeconfig := &api.Config{
+		Clusters: map[string]*api.Cluster{
+			options.clusterName: {
+				Server:                   options.apiServerEnpoint,
+				CertificateAuthorityData: options.caCert,
+			},
+		},
+		Contexts: map[string]*api.Context{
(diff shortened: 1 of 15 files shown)


pkg/controllers/kubeconfig/kubeconfig_test.go
```diff
@@ -0,0 +1,111 @@
...
+			Expect(test.CleanupAndWait(ctx, cl, serviceAccount, serviceAccountSecret, kubeconfigSecret)).To(Succeed())
+		})
+
+		It("should create a kubeconfig secret when it doesn't exist", func() {
+			Expect(r.reconcileKubeconfig(ctx)).To(Succeed())
+		})
+
+		It("should reconcile existing kubeconfig secret when it doesn't exist", func() {
+			Expect(r.reconcileKubeconfig(ctx)).To(Succeed())
+			Expect(r.reconcileKubeconfig(ctx)).To(Succeed())
```
elmiko, line 78: these two tests look confusing to me. in the first test i assume that the `reconcileKubeconfig` is creating the kubeconfig?

in the second tests though, does the first call create the kubeconfig and the second call reconcile it?

elmiko (commented):
this generally looks good to me, i just have a question about one of the tests.

pkg/controllers/kubeconfig/kubeconfig_test.go
alexander-demichev: yes, it first creates and next time reconciles it, AfterEach() block contains an assertion that kubeconfig is present

pkg/test/cleanup.go
```diff
@@ -0,0 +1,67 @@
...
+		errs = append(errs, errors.Wrapf(err, "key %s, %s is not being deleted from the testenv client cache", o.GetObjectKind().GroupVersionKind().String(), key))
+	}
+	return kerrors.NewAggregate(errs)
+}
+
+// cleanup deletes all the given objects.
+func cleanup(ctx context.Context, cl client.Client, objs ...client.Object) error {
+	errs := []error{}
+	for _, o := range objs {
+		err := cl.Delete(ctx, o)
```
lobziik, line 60: just note, it wont work if deployment or ds have finalizers

lobziik, line 60: just note, it wont work if deployment or ds have finalizers

alexander-demichev: we don't set any finalizers at the moment but yes, they should be handled ideally

pkg/controllers/kubeconfig/generate.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"errors"
+
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/openshift/cluster-capi-operator/pkg/controllers"
+)
+
+// token := string(tokenSecret.Data["token"]) // DECODE
```
lobziik, line 11: is this needed?

lobziik:
code itself
/lgtm

But i'm heavily concerned about having necessity to do such manipulations.

Fedosin (approved):
/approve

PR #44, opened 2022-03-28
Set control plane initialized condition

alexander-demichev:
Set control plane initialized condition to true, we should hardcode it before we get proper control plane management using CAPI.


JoelSpeed (commented):
/approve

PR #45, opened 2022-03-31
Use kustomize to compile assets

alexander-demicev:
Instead of requiring to have compiled components in provider forks, we can compile them here using kustomize in our script.


alexander-demichev:
/override ci/prow/e2e-aws-serial 
these failures are not related to tech preview components

PR #47, opened 2022-04-22
Add GCP and Azure to supported providers

alexander-demichev:
Add GCP, Azure to supported providers and deploy upstream operator using CVO.


manifests/image-references
```diff
@@ -14,3 +14,19 @@ spec:
...
+      kind: DockerImage
+      name: registry.ci.openshift.org/openshift:aws-cluster-api-controllers
+  - name: azure-cluster-api-controllers
+    from:
+      kind: DockerImage
+      name: registry.ci.openshift.org/openshift:azure-cluster-api-controllers
+  - name: gcp-cluster-api-controllers
+    from:
+      kind: DockerImage
+      name: registry.ci.openshift.org/openshift:gcp-cluster-api-controllers
```
JoelSpeed, line 32: Are these all in payload?

hack/assets/providercustomizations.go
```diff
@@ -0,0 +1,102 @@
...
+		gcpCustomizations(obj)
+	}
+}
+
+func operatorCustomizations(obj *unstructured.Unstructured) {
+	switch obj.GetKind() {
+	case "Deployment":
+		deployment := &appsv1.Deployment{}
+		if err := scheme.Convert(obj, deployment, nil); err != nil {
+			panic(err)
```
JoelSpeed, line 24: Why a panic? Is this always a programming error?

hack/assets/customizations.go
```diff
@@ -178,33 +187,35 @@ func findWebhookServiceSecretName(objs []unstructured.Unstructured) map[string]s
...
-	if isOperator {
-		// Modify manager container command to match openshift Dockerfile
-		for i := range deployment.Spec.Template.Spec.Containers {
-			container := &deployment.Spec.Template.Spec.Containers[i]
-			if container.Name == "manager" {
-				container.Command = []string{"./bin/cluster-api-operator"}
+		container.Resources.Limits = corev1.ResourceList{}
+		// Remove all image references if they are external, they will be substituted operator later
+		if !strings.HasPrefix(container.Image, "registry.ci.openshift.org") {
+			container.Image = "example.com/image:tag"
```
JoelSpeed, line 208: Can we insert something that doesn't look like an image? To make it more obvious this needs substituting?

hack/assets/operator.go
```diff
@@ -44,8 +44,21 @@ func importCAPIOperator() error {
 		return err
 	}
 
-	// Write all other components(deployments, services, secret, etc)
-	return writeAllOtherOperatorComponents(resourceMap[otherKey])
+	// Write deployment to manifests, it will be managed by CVO
+	deploymentFileName := fmt.Sprintf("%s%s_11_deployment.%s.yaml", manifestPrefix, "capi-operator", "upstream")
+	err = writeComponentsToManifests(deploymentFileName, resourceMap[deploymentKey])
+	if err != nil {
```
JoelSpeed, line 50: Nit
```suggestion
	if err := writeComponentsToManifests(deploymentFileName, resourceMap[deploymentKey]); err != nil {
```

hack/assets/operator.go
```diff
@@ -44,8 +44,21 @@ func importCAPIOperator() error {
...
+		return err
+	}
+
+	// Write CRD components to manifests, they will be managed by CVO
+	serviceFileName := fmt.Sprintf("%s%s_02_service.%s.yaml", manifestPrefix, "capi-operator", "upstream")
+	err = writeComponentsToManifests(serviceFileName, resourceMap[serviceKey])
+	if err != nil {
```
JoelSpeed, line 57: Nit
```suggestion
	if err := writeComponentsToManifests(serviceFileName, resourceMap[serviceKey]); err != nil {
``

File: pkg/controllers/cluster/azure.go [hunk 6]
```diff
@@ -0,0 +1,48 @@
...
+	if azurePlatformStatus == nil {
+		return errors.New("AzurePlatformStatus can't be nil")
+	}
+
+	azureCluster := &azurev1.AzureCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      r.clusterName,
+			Namespace: controllers.DefaultManagedNamespace,
+			Annotations: map[string]string{
+				clusterv1.ManagedByAnnotation: "",
```
JoelSpeed, line 28: Remind me where we landed on having a value for this annotation?

pkg/controllers/cluster/azure_test.go
```diff
@@ -0,0 +1,64 @@
...
+	It("should create a cluster with expected spec and status", func() {
+		Expect(r.reconcileAzureCluster(ctx, azurePlatformStatus)).To(Succeed())
+	})
+
+	It("should reconcile created cluster with expected spec and status", func() {
+		Expect(r.reconcileAzureCluster(ctx, azurePlatformStatus)).To(Succeed())
+		Expect(cl.Get(ctx, client.ObjectKey{
+			Name:      azureCluster.Name,
+			Namespace: azureCluster.Namespace,
+		}, azureCluster)).To(Succeed())
```
JoelSpeed, line 61: Is this not the same thing that's being checked in the after each? Not sure what this test case is adding

pkg/controllers/cluster/gcp_test.go
```diff
@@ -0,0 +1,72 @@
...
+	It("should create a cluster with expected spec and status", func() {
+		Expect(r.reconcileGCPCluster(ctx, gcpPlatformStatus)).To(Succeed())
+	})
+
+	It("should reconcile created cluster with expected spec and status", func() {
+		Expect(r.reconcileGCPCluster(ctx, gcpPlatformStatus)).To(Succeed())
+		Expect(cl.Get(ctx, client.ObjectKey{
+			Name:      gcpCluster.Name,
+			Namespace: gcpCluster.Namespace,
+		}, gcpCluster)).To(Succeed())
```
JoelSpeed, line 69: Same as above, how does this differ from the after each? Do we need to have a test that changes the spec/status and check that it gets put back?

pkg/controllers/cluster/azure.go
alexander-demichev: I don't think we discussed it

hack/assets/providercustomizations.go
alexander-demichev: Yes, this is something that should never happen

pkg/controllers/cluster/azure.go
JoelSpeed: I would have expected the owner to put a value in here no? So that you can see from the "api" who to talk to when it breaks, or so that you can tell that one specific operator is owning it and not another

pkg/controllers/cluster/azure_test.go
alexander-demichev: Yes, I modified the cluster object below and reconciled it again to check that it's back to the original state.

alexander-demichev (commented):
/approve

PR #48, opened 2022-04-22
Update cluster conditions only if they have been changed

Fedosin:
Currently we always send a request to update cluster conditions, even if they haven't changed. It leads to creation of multiple fake events, that the cluster status was updated.

This commit sends these requests only if there are real changes in cluster conditions.


mdbooth (approved):
/lgtm

We call SyncStatus() if any condition has changed. Although we check in a loop, we will only emit the log message once, when the change occurs. The 2 affected functions:

* don't change anything other than conditions
* fetch the object being synced

so there's no risk that a caller is expecting this method to write other changes to the CO as a side-effect.

mdbooth:
Lets rebase this on top of #49 when it merges.

alexander-demichev (approved):
/approve

mdbooth:
This change landed in #50. We should close it.

PR #51, opened 2022-05-03
Bug 2086198: [OCPCLOUD-1506] Disable defaulting webhook for cluster objects

Fedosin:
We don't need CAPI to set any default to the cluster object because we have a custom controller for reconciling it.

This PR drops all MutatingWebhookConfigurations via the import assets logic to prevent the defaulting.


hack/assets/providers.go
```diff
@@ -293,8 +293,20 @@ func importProviders() error {
 			return err
 		}
 
+		objs := []unstructured.Unstructured{}
+
+		// Explicitly remove all defaulting webhooks for the core provider.
+		// We don't need CAPI to set any default to the cluster object because
+		// we have a custom controller for reconciling it.
+		// For more information: https://issues.redhat.com/browse/OCPCLOUD-1506
+		for _, obj := range p.components.Objs() {
```
alexander-demichev, line 302: These better be handled here https://github.com/openshift/cluster-capi-operator/blob/main/hack/assets/customizations.go#L51. The code also removes webhooks for machine, machinesets, etc., but it should only remove webhook from the generic clusters and all infra clusters.

hack/assets/providercustomizations.go
```diff
@@ -100,3 +110,35 @@ func gcpCustomizations(obj *unstructured.Unstructured) {
...
+func removeClusterDefaultingWebhooks(obj *unstructured.Unstructured) {
+	mutatingWebhookConfiguration := &admissionregistrationv1.MutatingWebhookConfiguration{}
+	if err := scheme.Convert(obj, mutatingWebhookConfiguration, nil); err != nil {
+		panic(err)
+	}
+
+	webhooks := []admissionregistrationv1.MutatingWebhook{}
+	for _, webhook := range mutatingWebhookConfiguration.Webhooks {
+		switch webhook.Name {
+		case "default.cluster.cluster.x-k8s.io",
```
alexander-demichev, line 123: can we make this list more dynamic as more providers will be added in the future?

alexander-demichev (changes requested):

elmiko (commented):
this generally looks good to me, i'm not 100% sure on that regex for the webhook kind, but it looks correct to me.
/lgtm

PR #52, opened 2022-05-05
Add openshift annotations to operator deployment

alexander-demichev:
Add openshift annotations to operator deployment


Fedosin (approved):
/lgtm

PR #54, opened 2022-05-12
Create service account secret manually

alexander-demichev:
Create service account secret manually as described here, https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/ Current implementation will not work on 1.24 because the secret for SA is not created automatically anymore.


JoelSpeed:
/approve

Lets get this in to unblock the rebase and then iterate from there to move to token requests and handling the rotation ourselves

ravisantoshgudimetla (commented):
/lgtm

JoelSpeed:
Things seem to be working here, two out of three tech preview jobs passed and the failing one has positive logging from the affected code

JoelSpeed:
/override ci/prow/e2e-aws-serial

This worked earlier and nothing has changed. Also, this operator only runs when in tech preview so there's nothing being tested from this PR in this particular test run.

PR #57, opened 2022-05-14
Bug 2087149: Add watch predicates to kubeconfig controller

Fedosin:
Now it doesn't react on changes in the managed secrets, which allows users to modify them.

### DIFF
--- a/pkg/controllers/kubeconfig/watch_predicates.go
+++ b/pkg/controllers/kubeconfig/watch_predicates.go
@@ -1,0 +1,56 @@
+package kubeconfig
+
+import (
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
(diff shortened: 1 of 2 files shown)


pkg/controllers/kubeconfig/watch_predicates.go
```diff
@@ -0,0 +1,56 @@
...
+		UpdateFunc:  func(e event.UpdateEvent) bool { return isOwnedTokenSecret(e.ObjectNew) },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return isOwnedTokenSecret(e.Object) },
+		GenericFunc: func(e event.GenericEvent) bool { return isOwnedTokenSecret(e.Object) },
+	}
+}
+
+func kubeconfigSecretPredicate() predicate.Funcs {
+	isKubeconfigSecret := func(obj runtime.Object) bool {
+		secret, ok := obj.(*corev1.Secret)
+		if !ok {
```
alexander-demichev, line 43: are you sure that we want the code to panic and not use the pattern we have here https://github.com/openshift/cluster-capi-operator/blob/main/pkg/controllers/secretsync/watch_predicates.go#L21?

Fedosin: We decided to panic in these cases. For instance: https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/controllers/controlplanemachineset/watch_filters.go#L56-L59

Here's @JoelSpeed comment: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/28#discussion_r844859820

alexander-demichev (approved):
/approve
/lgtm

PR #59, opened 2022-05-18
Bug 2089254: Rotate token secret if its older than 30 minutes

alexander-demichev:
The token secret might expire, this PR makes sure that the secret is rotated every 30 minutes.

### DIFF
--- a/pkg/controllers/kubeconfig/suite_test.go
+++ b/pkg/controllers/kubeconfig/suite_test.go
@@ -3,6 +3,7 @@ package kubeconfig
 import (
 	"context"
 	"testing"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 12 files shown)


pkg/controllers/kubeconfig/suite_test.go
```diff
@@ -23,6 +24,7 @@ var (
 	cfg     *rest.Config
 	cl      client.Client
 	ctx     = context.Background()
+	timeout = 10 * time.Minute
```
JoelSpeed, line 27: 10 minute timeout?

alexander-demichev: probably it's too much, I replaced it with 1 minute

JoelSpeed: Normally I would expect to use something like 10 seconds, 1 minute for the controller to react would indicate an issue to me and surely means the watches aren't working right

alexander-demichev: 👍done

PR #60, opened 2022-05-25
Bug 2090236: Only reconcile annotations and status for clusters

alexander-demichev:
Currently, our controllers enforce some values in cluster object specs, this might not work for UPI. The PR drops most of the logic and also removes validating webhooks so we can use clusters in a more flexible way.


pkg/controllers/cluster/infra.go
```diff
@@ -0,0 +1,97 @@
...
+func (r *GenericInfraClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithName("InfraClusterController")
+
+	if err := r.Client.Get(ctx, req.NamespacedName, r.InfraCluster); err != nil && !errors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	log.Info("Reconciling infrastructure cluster")
+
+	switch r.InfraCluster.(type) {
```
JoelSpeed, line 40: Do we actually need to know the type? Can we not just use the generic controller-runtime client.Object to handle all this without caring about the type?

pkg/controllers/cluster/infra.go
```diff
@@ -0,0 +1,97 @@
...
+func (r *GenericInfraClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(r.InfraCluster).
+		Complete(r)
+}
+
+func (r *GenericInfraClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithName("InfraClusterController")
+
+	if err := r.Client.Get(ctx, req.NamespacedName, r.InfraCluster); err != nil && !errors.IsNotFound(err) {
```
JoelSpeed, line 34: Make a copy of `r.InfraCluster`, don't use the one embedded in the object, it's not going to be threadsafe if you do that

alexander-demichev:
> When we are running this in IPI, does this mean the values must now be populated by a user? Or do you intend in the future for the mapi to capi sync controller to synchronise this information when MAPI machines are present? 

I would expect the installer to handle the creation of clusters in the same way it does with machinesets, before GA users should do it manually and this PR also includes scripts to help with it. We have no way to know that the values we enforce are the proper ones in the case of UPI. We can discuss if some synchronization is possible with mapi<->capi controller in close future.

pkg/controllers/cluster/infra.go
```diff
@@ -0,0 +1,78 @@
...
+		return ctrl.Result{}, fmt.Errorf("unable to convert to unstructured: %v", err)
+	}
+
+	if err := unstructured.SetNestedField(unstructuredInfraCluster, true, "status", "ready"); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to set status: %w", err)
+	}
+
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredInfraCluster, infraClusterCopy); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to convert from unstructured: %v", err)
+	}
```
JoelSpeed, line 62: I don't think you need to convert back do you? The client can handle unstructured objects

alexander-demichev: `unstructuredInfraCluster` has the type `map[string]interface{}` in this case, we have to convert it back.

JoelSpeed: Ahh right, you have a runtime.Unstructured not a meta.Unstructured, annoyingly they are different yes, ok, I'll allow this

Fedosin (approved):
/lgtm
/approve

PR #63, opened 2022-06-21
Bug 2099695: Update CAPG after rebase

alexander-demichev:
Update CAPG after the latest rebase, all manifests are generated automatically using `make assets` command.


alexander-demichev (commented):
/approve

PR #69, opened 2022-08-01
Update docs

alexander-demichev:
Update developer documentation with relevant information.


docs/controllers/kubeconfig.md
```diff
@@ -11,13 +11,21 @@ stateDiagram-v2
...
+    IsServiceAccountSecretTooOld --> DeleterviceAccountSecret: True
+    DeleterviceAccountSecret --> Requeue
     NoOp --> [*]
 ```
 
 If the current platform is not supported, the controller will not create any secret and allow "bring your own" scenarios. 
 In cases where the platform is supported, the controller will create the secret containing kubeconfig.
+
+The controller will the service account secret that was initially created by the CVO. The token in the secret can exprire and has to
+be rotated. The controller will periodically check if the secret is too old and if so, it will delete the secret and wait for
```
lobziik, line 30: I would appreciate actual link to controller code

docs/controllers/infra-cluster.md
```diff
@@ -0,0 +1,24 @@
+# Infra cluster controller
+
+## Overview
+
+Infra cluster controller is responsible for managing InfrastructureCluster(AWSCluster, etc.) CRs. 
+The cluster object will represent the cloud infrastructure where current cluster where operator 
+is running because we treat this cluster as both management and workload.
```
lobziik, line 7: I suggest to rephrase this last sentence. I'm not sure that i've got it right...

lobziik:
I think, it would be nice to have a links to CAPI documentation, where Cluster, ClusterInfrastructure and another objects are described.  It might go to smth like "see also" sections within controller's related pages.

docs/controllers/core-cluster.md
```diff
@@ -0,0 +1,20 @@
+# Core Cluster controller
+
+## Overview
+
+Core cluster controller is responsible for managing Cluster CRs. The cluster object will
+represent the current cluster where operator is running because we treat this cluster as both management and workload.
+It's only purpose it to set `ControlPlaneInitialized` condition to true, in order to make Cluster API mover the cluster
```
JoelSpeed, line 7:
```suggestion
It's only purpose it to set `ControlPlaneInitialized` condition to true, in order to make Cluster API move the cluster
```

docs/controllers/kubeconfig.md
```diff
@@ -11,13 +11,21 @@ stateDiagram-v2
...
+    Requeue --> GetOperatorServiceAccountSecret
+    IsServiceAccountSecretTooOld --> DeleterviceAccountSecret: True
+    DeleterviceAccountSecret --> Requeue
     NoOp --> [*]
 ```
 
 If the current platform is not supported, the controller will not create any secret and allow "bring your own" scenarios. 
 In cases where the platform is supported, the controller will create the secret containing kubeconfig.
+
+The controller will the service account secret that was initially created by the CVO. The token in the secret can exprire and has to
```
JoelSpeed, line 29: > The controller will the service account secret

It will what?

JoelSpeed (commented):
Denis has an outstanding comment but otherwise LGTM

docs/controllers/kubeconfig.md
alexander-demichev: I included links for all controllers at the beginning of the doc

alexander-demichev (commented):
/approve

JoelSpeed:
This is just a docs update, we don't need to rerun CI constantly 

PR #72, opened 2022-08-22
Add tests that compare MAPI and CAPI created instances

alexander-demichev:
This PR adds tests that compare MAPI and CAPI-created instances, it will help us with feature parity. 
I also created a separate vendor directory for E2E.
Currently if instances don't match, it will only print the diff and fail entire suite.


JoelSpeed:
Seems like a good starting test, what do we need to get this running in CI?

alexander-demichev:
It already running, I made it print the diff if instances don't match and fail the suite. That behavior can be changed later if you want to fail. AWS looks very well, the only difference is in security groups, I'm not sure if that's a bug or not.

JoelSpeed:
Ack

/lgtm
/approve

JoelSpeed:

This change doesn't affect Azure

PR #74, opened 2022-09-06
Update manifests to use release.openshift.io/feature-set

deads2k:
This was added in
https://github.com/openshift/cluster-version-operator/pull/821 to allow
more featuresets and allow for a future migration to include actual gates

/assign @JoelSpeed 
/cc @wking


elmiko (commented):
/lgtm

PR #75, opened 2022-09-13
Bug 2118550: Momentarily disable Azure platform support

damdo:
Cluster API on OpenShift is in TechPreview and available only for a limited number of providers: AWS, GCP.
This PR disables Azure as a supported platform, until that will be deemed ready for TechPreview.


lobziik:
@JoelSpeed, @damdo, I suggest to override e2e-azure-techpreview for now. Almost all fails there coming from platform-operators operator and its misconfiguration. Only thing which might be interesting for us is `[sig-storage] In-tree Volumes [Driver: azure-file] [Testpattern: Dynamic PV (default fs)(allowExpansion)] volume-expand Verify if offline PVC expansion works [Suite:openshift/conformance/parallel] [Suite:k8s]`, but i think it has nothing to do with CAPI operator at all.

PR #76, opened 2022-09-21
Bug 2116686, Bug OCPBUGS-1493: Initial implementation of validation webhooks

Fedosin:
This PR starts the implementation of the validation webhooks for CAPI objects.
In particular, it validates that Cluster object infrastructure reference has correct value.

### DIFF
--- a/manifests/0000_30_cluster-api_10_webhooks.yaml
+++ b/manifests/0000_30_cluster-api_10_webhooks.yaml
@@ -7,6 +7,7 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
     release.openshift.io/feature-set: TechPreviewNoUpgrade
+    service.beta.openshift.io/inject-cabundle: "true"
   name: validating-webhook-configuration
 webhooks:
   - admissionReviewVersions:
@@ -17,6 +18,7 @@ webhooks:
         name: cluster-capi-operator-webhook-service
         namespace: openshift-cluster-api
         path: /validate-operator-cluster-x-k8s-io-v1alpha1-coreprovider
+        port: 9443
     failurePolicy: Fail
     name: vcoreprovider.operator.cluster.x-k8s.io
     rules:
@@ -37,6 +39,7 @@ webhooks:
         name: cluster-capi-operator-webhook-service
         namespace: openshift-cluster-api
         path: /validate-operator-cluster-x-k8s-io-v1alpha1-infrastructureprovider
+        port: 9443
     failurePolicy: Fail
     name: vinfrastructureprovider.operator.cluster.x-k8s.io
     rules:
@@ -48,4 +51,26 @@ webhooks:
           - DELETE
         resources:
           - infrastructureproviders
+    sideEffects: None
+  - admissionReviewVersions:
+      - v1
+      - v1beta1
+    clientConfig:
+      service:
+        name: cluster-capi-operator-webhook-service
+        namespace: openshift-cluster-api
+        path: /validate-cluster-x-k8s-io-v1beta1-cluster
+        port: 9443
+    failurePolicy: Fail
+    name: validation.cluster.cluster.x-k8s.io
+    rules:
+    - apiGroups:
+      - cluster.x-k8s.io
+      apiVersions:
+      - v1beta1
+      operations:
+      - CREATE
+      - UPDATE
+      resources:
+      - clusters
     sideEffects: None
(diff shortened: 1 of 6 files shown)


elmiko (commented):
i couldn't tell if the test failures are related to this pr, it didn't look like it, but this generally makes sense to me.
/lgtm

pkg/webhook/cluster.go
```diff
@@ -0,0 +1,67 @@
...
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *ClusterWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) error {
+	cluster, ok := obj.(*v1beta1.Cluster)
+	if !ok {
+		panic("expected to get an of object of type v1beta1.Cluster")
+	}
+	switch cluster.Spec.InfrastructureRef.Kind {
+	case "AWSCluster":
+	case "AzureCluster":
+	case "GCPCluster":
```
JoelSpeed, line 34: Nit, can inline these
```suggestion
	case "AWSCluster", "AzureCluster", "GCPCluster":
```

pkg/webhook/cluster.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	newCluster, ok := newObj.(*v1beta1.Cluster)
+	if !ok {
+		panic("expected to get an of object of type v1beta1.Cluster")
+	}
+
+	switch newCluster.Spec.InfrastructureRef.Kind {
+	case "AWSCluster":
+	case "AzureCluster":
+	case "GCPCluster":
```
JoelSpeed, line 56: Nit, inline please

pkg/webhook/coreprovider.go
```diff
@@ -24,15 +24,37 @@ var _ webhook.CustomValidator = &CoreProviderWebhook{}
...
+		return fmt.Errorf("incorrect core provider name: %s", newCoreProvider.Name)
+	}
+
 	return nil
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
 func (r *CoreProviderWebhook) ValidateDelete(_ context.Context, obj runtime.Object) error {
-	return errors.New("deletion of infrastructure provider is not allowed")
+	return fmt.Errorf("deletion of infrastructure provider is not allowed")
```
JoelSpeed, line 59: You aren't using a format string so this should be `errors.New`

pkg/webhook/infraprovider.go
```diff
@@ -24,15 +24,37 @@ var _ webhook.CustomValidator = &InfrastructureProviderWebhook{}
...
+		return fmt.Errorf("incorrect infra provider name: %s", newInfraProvider.Name)
+	}
+
 	return nil
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
 func (r *InfrastructureProviderWebhook) ValidateDelete(_ context.Context, obj runtime.Object) error {
-	return errors.New("deletion of infrastructure provider is not allowed")
+	return fmt.Errorf("deletion of infrastructure provider is not allowed")
```
JoelSpeed, line 59: You aren't using a format string so this should be `errors.New`

JoelSpeed (commented):
I'd like to see some tests added for this before we merge. Please take a look at the CPMS webhook test for some inspiration on how we might do this.

It would also be good to set up some linting here, again, similar to CPMS

JoelSpeed (commented):
/approve
/lgtm

Fedosin:
/hold cancel

seems working

manifests/0000_30_cluster-api_10_webhooks.yaml
```diff
@@ -49,3 +52,25 @@ webhooks:
         resources:
           - infrastructureproviders
     sideEffects: None
+  - admissionReviewVersions:
+      - v1
+      - v1beta1
```
JoelSpeed, line 57: Do we need both versions or should we just use `v1`?

Fedosin: In this cluster-api [example](https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/webhooks/cluster.go#L49) they recommend to use both versions.

JoelSpeed:
/hold this is legitimately broken

The test suite downloads the latest version of the CLI instead of building from the source vendored into the project, this causes an issue because when new flags are added, the CLI sends new options to the library which then errors out. 

We need to make sure the CLI and library version are consistent.

Fedosin:
@JoelSpeed I proposed a fix for CI: https://github.com/openshift/cluster-capi-operator/pull/80

PR #83, opened 2022-11-04
Bug 2087043: Update manifests to release-4.13

Fedosin:
Bump manifests for CAPI components.

Generated by:
make assets

### DIFF
--- a/manifests/0000_30_cluster-api_02_crd.infrastructure-gcp.yaml
+++ b/manifests/0000_30_cluster-api_02_crd.infrastructure-gcp.yaml
@@ -1270,6 +1270,12 @@ spec:
                           description: Description is an optional description associated
                             with the resource.
                           type: string
+                        enableFlowLogs:
+                          description: 'EnableFlowLogs: Whether to enable flow logging
+                            for this subnetwork. If this field is not explicitly set,
+                            it will not appear in get listings. If not set the default
+                            behavior is to disable flow logging.'
+                          type: boolean
                         name:
                           description: Name defines a unique identifier to reference
                             this resource.
@@ -1283,12 +1289,6 @@ spec:
                           description: Region is the name of the region where the
                             Subnetwork resides.
                           type: string
-                        routeTableId:
-                          description: 'EnableFlowLogs: Whether to enable flow logging
-                            for this subnetwork. If this field is not explicitly set,
-                            it will not appear in get listings. If not set the default
-                            behavior is to disable flow logging.'
-                          type: boolean
                         secondaryCidrBlocks:
                           additionalProperties:
                             type: string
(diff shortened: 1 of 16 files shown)


JoelSpeed:
If we fetch from upstream, we have no control over what it is that we are generating, we also have no ability to update the assets independent of the upstream, are you absolutely sure we want to be picking this from upstream and not our own downstream main branch?

How will this work for older releases, do they need to be set to an old release?

/hold I'd like to understand a bit more about why this is the value it is today before we merge, @Fedosin can you see if you can work out why this has been pinned to a release branch previously

Fedosin:
> If we fetch from upstream, we have no control over what it is that we are generating

Well, it's quite opposite - we define what changes we want to add here. For instance, we change image names, add OpenShift specific label and annotations, and so on. We even have functions exactly for customizations of upstream manifests:
https://github.com/openshift/cluster-capi-operator/blob/main/hack/assets/customizations.go
https://github.com/openshift/cluster-capi-operator/blob/main/hack/assets/providercustomizations.go

The origional goal was to never touch upstream manifests, keep them as-is in operands repositories, and add modifications during asset generation.

JoelSpeed:
/approve cancel

Some discussion about this during the arch call

manifests/0000_30_cluster-api_02_crd.infrastructure-gcp.yaml
```diff
@@ -1283,12 +1289,6 @@ spec:
                           description: Region is the name of the region where the
                             Subnetwork resides.
                           type: string
-                        routeTableId:
-                          description: 'EnableFlowLogs: Whether to enable flow logging
-                            for this subnetwork. If this field is not explicitly set,
-                            it will not appear in get listings. If not set the default
-                            behavior is to disable flow logging.'
-                          type: boolean
```
JoelSpeed, line 1291: They dropped a field?

Fedosin: Renamed it - https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/665

But yeah, it breaks backward compatibility

JoelSpeed: Right, and we are making this change between major releases right? So it should be ok. In the future we would need to  be consistent about upgrade paths for this

Fedosin: Yes, correct.

JoelSpeed:
Seems there's something wrong with the assets check, can you please look into that and let me know what's wrong with it

Fedosin:
Yes, `asset` failure is expected because I manually updated cluster api operator manifests to make it work. Since they are different from what we have in the forked repo, the job fails. It will be fixed when we merge https://github.com/openshift/cluster-api-operator/pull/9

Fedosin:
We can't merge [it](https://github.com/openshift/cluster-api-operator/pull/9) because we need to have correct manifests here first. I think we agreed on updating manifests in semi-manual mode, otherwise those tests won't pass.

JoelSpeed:
/lgtm
/approve

@Fedosin We need to force this through right? The tests won't pass until other repos are updated? Or should the E2E be passing?

Fedosin:
/hold cancel

no, it's all good - all test passed

PR #84, opened 2022-11-08
[OCPCLOUD-1771] Add header to generated manifests

Fedosin:
To indicate that some manifests are automatically generated, we need to add a related header to them.


JoelSpeed:
Seems reasonable
/lgtm
/approve

PR #87, opened 2022-11-23
[OCPCLOUD-1726]: GCP Credential Req. manifest for CAPI to use new API field

odvarkadaniel:
This PR removes all the `predefinedRoles` which give elevated permissions. We get those permissions and put them into the new API field `Permissions []string`. These permissions are known to thanks to the command `gcloud iam roles describe <role>` - this lists all the included permissions we want.


JoelSpeed:
> this lists all the included permissions we want.

Have you confirmed that this list is minimal? It's a very long list so i'd like to make sure that we do actually require each and every permission in the list. If you have, how did you do so?

elmiko:
i'm curious if these permissions should look similar to the [machine api related PR](https://github.com/openshift/machine-api-operator/pull/1087) ?

abutcher:
We're leaking these custom roles created by CCO in CI so holding until we're cleaning them up.
/hold

PR #88, opened 2022-11-28
Don't patch objects if there are no changes

Fedosin:
Currently we send a request to update some objects on each reconciliation, even if these objects were not modified.

To prevent this we introduce checks to ensure that changes were made before sending patch requests.


elmiko:
i don't think the test failures are related

/lgtm

Fedosin:
/hold

ci is broken, I'm fixing it.

PR #94, opened 2022-12-22
fix indentation for webhook manifests

Fedosin:
Despite the fact the indentation is correct, these manifests are not correctly parsed by CVO, and some fields are missing.

To fix it we update the indentation to bring it to a more standard variant.


JoelSpeed:
/approve
/lgtm

PR has been pre-merge verified by QE

PR #106, opened 2023-03-16
OCPBUGS-4147: Enforce single Cluster in openshift-cluster-api namespace

RadekManak:
We want only one Cluster in the openshift-cluster-api namespace that creates machines for the cluster we are running on. To prevent creation of multiple clusters, this PR adds webhook that allows only clusters named <InfrastructureName> that is obtained from the InfrastructureObject.

### DIFF
--- a/pkg/webhook/cluster.go
+++ b/pkg/webhook/cluster.go
@@ -6,15 +6,26 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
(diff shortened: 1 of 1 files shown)


pkg/webhook/cluster.go
```diff
@@ -23,16 +34,60 @@ func (r *ClusterWebhook) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
 		panic("expected to get an of object of type v1beta1.Cluster")
 	}
-	switch cluster.Spec.InfrastructureRef.Kind {
-	case "AWSCluster", "AzureCluster", "GCPCluster", "IBMPowerVSCluster":
-	default:
-		return fmt.Errorf("unsupported cluster infra provider kind: %s", cluster.Spec.InfrastructureRef.Kind)
+
+	errs := []error{}
+	infrastructureRefPath := field.NewPath("spec", "infrastructureRef")
+	if cluster.Spec.InfrastructureRef != nil {
```
JoelSpeed, line 78: This change seems unrelated to the other changes, why do we allow a nil infrastructure ref? When would it ever be created without one? /should there be an else here in case it is empty?

RadekManak: I was fixing a panic and did not want to make it a required field. All the field in the spec are optional. I am not sure what kind of validation we want in openshift-cluster-api namespace vs other namespaces, or if we plan to support running other workload clusters.

JoelSpeed: Makes sense! Thanks

PR #114, opened 2023-05-03
OCPCLOUD-1609: Support for VSphere

odvarkadaniel:
This PR enables CAPI on OCP for VSphere. This will also enable us to investigate what needs to be done in order to run `machineSets` on this provider for CAPI.


odvarkadaniel:
This PR introduces a support for multiple vSphere CAPI related things:
1. E2E pre-submit tests
2. Ability to generate infrastructure cluster with `create-vsphere.sh`
3. Ease of deploying `capv` with developer image
4. Webhook supports vSphere platform

There is also a PR in `release` repository that creates a new job for this e2e suite to run ([PR](https://github.com/openshift/release/pull/40432)).

/hold cancel
/cc @JoelSpeed

odvarkadaniel:
/hold

Have to review the e2e tests failure as we already have the job.

odvarkadaniel:
My quay repository with capv was private. Hopefully that is the issue. Retesting.

/retest

odvarkadaniel:
Assets are going to get fixed in #124 .

odvarkadaniel:
In order for the machineSets to correctly function, we need this PR https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2071 to get merged as well. It assign a DNSName to a vCenter VM, which is necessary for the ignition to boot up the VM correctly. After this gets merged, we will need to update the version of CAPV within cluster-capi to contain the change. E2E tests should pass afterwards.

odvarkadaniel:
Check vsphere-creds secret in kube-system to generate the cluster named secret for capv.

odvarkadaniel:
https://github.com/openshift/machine-config-operator/pull/4037
https://github.com/openshift/cluster-api-provider-vsphere/pull/19

Once these two PRs are merged and set up in this repo accordingly (I don't know what would need to be done to point at the downstream version of CAPV in this case, @damdo), I have verified, that the e2e test case to test machinesets on vSphere work out of the box.

assets/infrastructure-providers/infrastructure-vsphere-provider.yaml
```diff
@@ -0,0 +1,19 @@
...
+    containers:
+    - command:
+      - ./bin/cluster-api-provider-vsphere-controller-manager
+      name: manager
+  fetchConfig:
+    selector:
+      matchLabels:
+        provider.cluster.x-k8s.io/name: vsphere
+        provider.cluster.x-k8s.io/type: infrastructure
+  version: v1.7.0
```
damdo, line 18: We now have `v1.8.3` in payload. I'd be better to switch to that for compatibility.

dev-images.json
```diff
@@ -5,5 +5,6 @@
   "azure-cluster-api-controllers": "quay.io/ademicev/cluster-api-provider-azure:latest",
   "gcp-cluster-api-controllers": "quay.io/ademicev/cluster-api-provider-gcp:latest",
   "ibmcloud-cluster-api-controllers": "quay.io/kabhat/ibmcloud-cluster-api-controllers:dev",
+  "vsphere-cluster-api-controllers": "registry.build01.ci.openshift.org/ci-ln-kq6nx7b/release:latest",
```
damdo, line 8: This won't work in the future as it is a temporary CI image that would've already been deleted by now.
It'd be best if you could pull this down and push it to your quay registry as a public img similarly to the others.

e2e/framework/cluster.go
```diff
@@ -30,6 +30,18 @@ func CreateCoreCluster(cl client.Client, clusterName, infraClusterKind string) *
...
+		if err != nil {
+			Expect(err).ToNot(HaveOccurred())
+		}
+
+		cluster.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{
+			Host: host,
+			Port: port,
+		}
+	}
+
```
damdo, line 44: Do we need this only for VSphere?
Ideally we should avoid special cases in the framework.

e2e/framework/util.go
```diff
@@ -0,0 +1,34 @@
...
+	"strings"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetControlPlaneHostAndPort(cl client.Client) (string, int32, error) {
+	var infraCluster configv1.Infrastructure
+	namespacedName := client.ObjectKey{
+		Namespace: "openshift-cluster-api",
```
damdo, line 14: Do we already have this value in the constants?
Can we reference it from there?

e2e/framework/util.go
```diff
@@ -0,0 +1,34 @@
...
+		Name:      "cluster",
+	}
+
+	if err := cl.Get(ctx, namespacedName, &infraCluster); err != nil {
+		return "", 0, err
+	}
+
+	hostAndPort := strings.TrimPrefix(infraCluster.Status.APIServerURL, "https://")
```
damdo, line 22: Is it always guaranteed that we get the APIServerURL with an `https` prefix? Or could it also be `http` or else?

go.mod
```diff
@@ -29,6 +29,8 @@ require (
 	sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20221007015352-8ad090e0663e
 )
 
+require sigs.k8s.io/cluster-api-provider-vsphere v1.8.3
+
```
damdo, line 33: Any specific reason why this is a separate require?

manifests/0000_30_cluster-api_02_providers.configmap.yaml
```diff
@@ -17,6 +17,10 @@ data:
       type: InfrastructureProvider
       branch: release-4.15
       version: v0.6.0
+    - name: vsphere
+      type: InfrastructureProvider
+      branch: release-4.14
+      version: v1.7.0
```
damdo, line 23: `v1.8.3`

manifests/0000_30_cluster-api_02_providers.configmap.yaml
```diff
@@ -17,6 +17,10 @@ data:
       type: InfrastructureProvider
       branch: release-4.15
       version: v0.6.0
+    - name: vsphere
+      type: InfrastructureProvider
+      branch: release-4.14
```
damdo, line 22: `release-4.15`

providers-list.yaml
```diff
@@ -14,3 +14,7 @@
   type: InfrastructureProvider
   branch: release-4.15
   version: v0.6.0
+- name: vsphere
+  type: InfrastructureProvider
+  branch: release-4.14
+  version: v1.7.0
```
damdo, line 20: Same here

damdo:
@odvarkadaniel something might be wrong with the changes related to the userdata.
See the [CAPI](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-capi-operator/114/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1729048811763929088/artifacts/e2e-aws-capi-techpreview/gather-extra/artifacts/pods/openshift-cluster-api_capi-controller-manager-b9d4c7bd4-rbsnc_manager.log) and [CAPA](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-capi-operator/114/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1729048811763929088/artifacts/e2e-aws-capi-techpreview/gather-extra/artifacts/pods/openshift-cluster-api_capa-controller-manager-78597975d9-82dbg_manager.log) logs in the last `e2e-aws-capi-techpreview` job run.

e2e/framework/cluster.go
odvarkadaniel: Yes, AFAIK, this is necessary for vSphere only. I haven't tried what it would do with other providers if we make this setting global.
VSphere needs it so that the management cluster correctly communicates with vCenter to create resources, edit, etc.

e2e/framework/util.go
odvarkadaniel: Yes, we do. Missed that.

dev-images.json
odvarkadaniel: My bad, forgot to change this back.

e2e/framework/cluster.go
damdo: It looks like we are setting this up also for other platforms (i.e. [AWS](https://github.com/openshift/cluster-capi-operator/pull/126/files#diff-872c7f24229a19d1d00c03edf29459f9fd8c1edcb92fcaa61644ad8a1370e22cR116-R119)), but we are doing this in the `<provider>_test.go` rather than in the framework.

I think it'd be nice to try and centralise this for all the providers in the framework (by removing the `if infraClusterKind == x`, and removing that set up in the other `<provider>_test.go`). If that doesn't work, I'd move the VSPhere one in the `vsphere_test.go` similarly to how it is already done in the AWS one.

odvarkadaniel: But I don't think we are setting it up for the core cluster object for any other provider.

damdo: Ah it's on the core cluster also. That's quite curious.
And vsphere E2Es don't work if this is not set on the core `Cluster` object?

pkg/controllers/secretsync/secret_sync_controller.go
```diff
@@ -103,13 +102,11 @@ func (r *UserDataSecretController) syncSecretData(ctx context.Context, source *c
 		return fmt.Errorf("source secret does not have user data")
 	}
 
-	formatEncoded := base64.StdEncoding.EncodeToString([]byte("ignition"))
```
JoelSpeed, line 106: Should it not be base64 encoded? Or was this a double base64 encoding?

e2e/aws_test.go
```diff
@@ -167,12 +167,12 @@ func newAWSMachineTemplate(cl client.Client, mapiProviderSpec *mapiv1.AWSMachine
...
 		FailureDomain:        &mapiProviderSpec.Placement.AvailabilityZone,
-		CloudInit: awsv1.CloudInit{
-			InsecureSkipSecretsManager: true,
-		},
 		AMI: awsv1.AMIReference{
 			ID: mapiProviderSpec.AMI.ID,
 		},
+		Ignition: &awsv1.Ignition{
+			Version: "3.4",
+		},
```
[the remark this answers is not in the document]
odvarkadaniel: I think we can include it already, the field should be in their API for some time already. Wdyt?

pkg/controllers/secretsync/secret_sync_controller.go
odvarkadaniel: This was a case of double base64 encoding. What is there currently is correct.

e2e/aws_test.go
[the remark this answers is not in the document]
JoelSpeed: It's an enum so it will fail validation and you won't be able to create the object with this value

damdo:
Taking this over.
Now superseded by https://github.com/openshift/cluster-capi-operator/pull/149

/close

e2e/aws_test.go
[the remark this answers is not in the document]
damdo: @JoelSpeed Here is the AWS rebase PR to v2.3.0, which contains @vincepri 's  https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4553 Ignition v3 support for CAPA

PR #119, opened 2023-06-30
[DNM][WIP] Rework: direct CAPI manifest apply

damdo:
A POC to test direct CAPI manifest apply:
- do not deploy upstream `cluster-api-operator` (remove the relevant manifests from this repo's `/manifests`)
- at runtime in `clusteroperator_controller` only apply the configmap with the components loaded from `assets` and don't apply the Provider CRs for cluster-api-operator (this controller can be entirely dropped once the asset generation happens in every provider's fork and the generated configmaps are stored in the respective `/manifests` folders, the applied by the CVO
- at runtime in a new `capiinstaller_controller` fetch the configmaps for the providers, substitute the values (envsubst) and the container images/ commands, parse the various components and apply them to the cluster


damdo:
Closing in favour of #121

PR #124, opened 2023-07-24
OCPCLOUD-2012: Add granular permissions for Azure

odvarkadaniel:
Instead of using predefined roles on Azure, we want to use granular permissions to gain more control of what can be done on the cluster.

PR #125, opened 2023-07-26
Use the workload identity

RadekManak:
This PR splits the introduction of workload identity part of #124 and allows us to do the granular permissions in later PR.


damdo:
Is this reviewable? @RadekManak

PR #126, opened 2023-08-31
OCPBUGS-18504: e2e: set ControlPlaneEndpoint in AWSCluster

damdo:
This sets the ControlPlaneEndpoint before Cluster object creation. To make the Cluster go into `Provisioned` state.
This is a temporary workaround, as we expect the Cluster & InfrastructureCluster objects creation and the population of the ControlPlaneEndpoint is going to happen in a dedicated controller within the operator.

PR #128, opened 2023-09-11
MULTIARCH-3761: Power VS CAPI Remove deprecated flag

Karthik-K-N:
`powervs-provider-id-fmt` flag is deprecated and we are supposed to use `provider-id-fmt` flag. [Reference](https://github.com/openshift/cluster-api-provider-ibmcloud/blob/main/main.go#L168C2-L175)

### DIFF
--- a/assets/infrastructure-providers/infrastructure-ibmcloud.yaml
+++ b/assets/infrastructure-providers/infrastructure-ibmcloud.yaml
@@ -57,8 +57,8 @@ data:
           - args:
             - --leader-elect
             - --metrics-bind-addr=127.0.0.1:8080
-            - --powervs-provider-id-fmt=v2
-            - --provider-id-fmt=${PROVIDER_ID_FORMAT:=v1}
+            - --powervs-provider-id-fmt=${POWERVS_PROVIDER_ID_FORMAT:=v1}
+            - --provider-id-fmt=v2
             - --service-endpoint=${SERVICE_ENDPOINT:=none}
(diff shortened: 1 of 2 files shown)


damdo:
Thanks for the PR @Karthik-K-N 
Assets are failing for a different reason, which is going to be fixed in #126.
Then we can retest the assets and merge this too.

damdo (commented):
/approve

Karthik-K-N:
> Thanks for the PR @Karthik-K-N Assets are failing for a different reason, which is going to be fixed in #126. Then we can retest the assets and merge this too.

sure, Thanks will watch and update accordingly

assets/infrastructure-providers/infrastructure-ibmcloud.yaml
```diff
@@ -57,8 +57,8 @@ data:
           - args:
             - --leader-elect
             - --metrics-bind-addr=127.0.0.1:8080
-            - --powervs-provider-id-fmt=v2
-            - --provider-id-fmt=${PROVIDER_ID_FORMAT:=v1}
+            - --powervs-provider-id-fmt=${POWERVS_PROVIDER_ID_FORMAT:=v1}
```
Karthik-K-N, line 60: This flag will be automatically removed while generating asset during next capi version update.
As it is already [removed](https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/config/manager/manager.yaml#L29-L33) in upstream

PR #135, opened 2023-10-24
OCPBUGS-19842: Limit deletion webhook to openshift namespace

RadekManak:
This PR changes the Cluster delete webhook, so that it will only prevent deletion when the cluster is in openshift-cluster-api namespace.

### DIFF
--- a/pkg/webhook/cluster.go
+++ b/pkg/webhook/cluster.go
@@ -2,7 +2,6 @@ package webhook
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


damdo (commented):
/approve

pkg/webhook/cluster.go
```diff
@@ -21,6 +21,8 @@ const (
 	openshiftCAPINamespace = "openshift-cluster-api"
 )
 
+var errClusterDeletionNotAllowedInOpenShiftCAPINamespace = errors.New("deletion of cluster is not allowed in openshift-cluster-api namespace")
```
damdo, line 24: I think you could compose this error string with a fmt.Sprintf internally so you can reference `openshiftCAPINamespace` instead of hardcoding it

JoelSpeed:
/lgtm

This only affects a tech preview feature and is not something we currently exercise in any release blocking or informing jobs

PR #140, opened 2023-10-26
OCPBUGS-19841: Add format key into the worker-user-data secret

odvarkadaniel:
There is a consensus in the upstream community, that every provider, that supports ignition as its bootstrapping method should adopt field `format` with value `ignition` (encoded as base64 string) to indicate that the user intends to use this feature. If this value would be empty, the upstream providers operator would assume that the user wants to use the other option called `cloud-init` that we don't intend to support in the future.

### DIFF
--- a/pkg/controllers/secretsync/secret_sync_controller.go
+++ b/pkg/controllers/secretsync/secret_sync_controller.go
@@ -2,6 +2,7 @@ package secretsync
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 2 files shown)


damdo:
@odvarkadaniel what kind of testing did you perform to check this works as expected? In other words, does it play correctly with the component that was requesting this change?

Also could you add a description along the lines of the bug to explain the change here?
Thanks!

odvarkadaniel:
> @odvarkadaniel what kind of testing did you perform to check this works as expected? In other words, does it play correctly with the component that was requesting this change?
> 
> Also could you add a description along the lines of the bug to explain the change here? Thanks!

I only tested this with vSphere at the moment and works as expected. I don't see any potential issues with this change for other providers. Maybe @miyadav could pre-merge test the change to verify this.

A usage of this field can be seen in CAPA:

- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/scope/machine.go#L203-L205
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmachine_controller.go#L843

damdo:
Got it thanks.
I'm happy with the testing you did with VSphere and checking that it works.
Also considering that for the other providers we have green E2E tests.

/lgtm
/retest

damdo (commented):
/approve

Thanks!

odvarkadaniel:
/retest 

VSphere not passing should not be a concern here. From the logs, we can tell that the issue is not brought in with these changes. Should be valid for an override of the test.

odvarkadaniel:
> /retest
> 
> VSphere not passing should not be a concern here. From the logs, we can tell that the issue is not brought in with these changes. Should be valid for an override of the test.

VSphere CI is currently broken.

PR #132, opened 2023-10-11
OCPCLOUD-2214,OCPBUGS-17287,OCPBUGS-21642: Bump capi/capi providers/k8s 1.27

damdo:
Done:
- bump k8s dependencies 1.27 (aside from the laggard GCP, handled later)
- bump CAPI provider deps to versions compatible to 1.27
- bump CAPI core deps to version compatible to 1.27

> [!WARNING]  
> This PR needs to be "GREEN BUTTON MERGED" at the same time as the cluster-api-operator [PR](https://github.com/openshift/cluster-api-operator/pull/25), as the new code of that operator references a new CRD (deployed by this repo), so to avoid breaking the payload we need to merge them together. (We are planning to fix this, [see enhancement](https://github.com/openshift/enhancements/pull/1461)) 

Todo:
- [ ] bump providers-list.yaml and rerun make assets (once provider forks rebase PRs have merged)

As part of this process we are also merging Rebase PRs on the CAPI provider forks.
I will track here E2E tests ran from this branch/PR against the rebase PRs in the various CAPI provider repos forks.

I will track here E2E tests ran from this branch/PR against the rebase PRs in the various CAPI provider repos forks ([details for E2E runs below]()).
- [x] [AWS](https://github.com/openshift/cluster-api-provider-aws/pull/478) - [test-run](https://github.com/openshift/cluster-capi-operator/pull/132#issuecomment-1759468096) 
- [x] GCP (no PR yet, only `master` testing) - [test-run](https://github.com/openshift/cluster-capi-operator/pull/132#issuecomment-1761620439) 
- [x] [IBMCloud](https://github.com/openshift/cluster-api-provider-ibmcloud/pull/59) - [test-run](https://github.com/openshift/cluster-api-provider-ibmcloud/pull/59#issuecomment-1758936273) 
- [x] [Vsphere](https://github.com/openshift/cluster-api-provider-vsphere/pull/18) - not testable due to no E2Es yet
- [x] [Azure](https://github.com/openshift/cluster-api-provider-azure/pull/284) - not testable due to no E2Es yet

PR #145, opened 2023-11-20
OCPBUGS-23467: manifests/0000_30_cluster-api_10_webhooks: Deletion manifest for validating-webhook-configuration

wking:
The resource was renamed to `cluster-capi-operator` in 1096d5c8ee (#132), and that's a much more specific name for the cluster-scoped resource.  However, clusters upating from 4.15.0-ec.1 with the old name should not have two similar-but-not-quite-matching versions of the webhook configuration, or updates can stick with `cluster-api` ClusterOperator going `Degraded=True` on `SyncingFailed` [with][1]:

    Failed to resync for operator: 4.15.0-ec.2 because &{%!e(string=unable to reconcile CoreProvider: unable to create or update CoreProvider: Internal error occurred: failed calling webhook "vcoreprovider.operator.cluster.x-k8s.io": failed to call webhook: the server could not find the requested resource)}

This commit adds [a deletion manifest][2] to ask the cluster-version operator to remove the resource, to avoid leaking the old resource into new clusters.

[1]: https://issues.redhat.com/browse/OCPBUGS-23467
[2]: https://github.com/openshift/enhancements/blob/8a11b7085c69b078690d50958f541d2a817d85df/dev-guide/cluster-version-operator/dev/object-deletion.md

### DIFF
--- a/manifests/0000_30_cluster-api_10_webhooks.yaml
+++ b/manifests/0000_30_cluster-api_10_webhooks.yaml
@@ -100,3 +100,14 @@ webhooks:
         resources:
           - providers
     sideEffects: None
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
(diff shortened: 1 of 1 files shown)


damdo (commented):
/approve

PR #148, opened 2023-11-30
OCPBUGS-24301: openstack: Don't substitute manager container command

mdbooth:
Also remove openstack from image replacement as openstack is not in cluster-capi-operator's image-references and does its own release image substitution anyway.


damdo (commented):
Thanks matt
/approve
/lgtm

PR #157, opened 2024-01-24
OCPBUGS-28539: Start operator controllers only on supported platforms, add unsupported controller

damdo:
Relying only on the CAPI provider (transport) ConfigMap to be or not to be there, to decide whether to enable CAPI on a platform or not, is not always sufficient as while we develop we may want to add a ConfigMap without actually enabling the controller.

Moreover without this in place we would always setup the Core CAPI provider and the operator webhooks even on platforms which are not supported, which is wasteful and not ideal.

As such we want to start the operator controllers and set up webhooks only on supported platforms.

When the operator is running on a platform where CAPI is not supported, run the unsupported controller instead.

### DIFF
--- a/pkg/controllers/unsupported/unsupported_controller.go
+++ b/pkg/controllers/unsupported/unsupported_controller.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package unsupported
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-capi-operator/pkg/controllers"
+	"github.com/openshift/cluster-capi-operator/pkg/operatorstatus"
+)
+
+const (
+	capiUnsupportedPlatformMsg = "Cluster API is not yet implemented on this platform"
+)
+
+// UnsupportedController on unsupported platforms watches and keeps the cluster-api ClusterObject up to date.
+type UnsupportedController struct {
+	operatorstatus.ClusterOperatorStatusClient
+	Scheme *runtime.Scheme
+}
+
+// Reconcile reconciles the cluster-api ClusterOperator object.
+func (r *UnsupportedController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithName("UnsupportedController")
+	log.Info(fmt.Sprintf("Reconciling %q ClusterObject", controllers.ClusterOperatorName))
+
+	if err := r.ClusterOperatorStatusClient.SetStatusAvailable(ctx, capiUnsupportedPlatformMsg); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to set conditions for %q ClusterObject: %w", controllers.ClusterOperatorName, err)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *UnsupportedController) SetupWithManager(mgr ctrl.Manager) error {
+	build := ctrl.NewControllerManagedBy(mgr).
+		For(&configv1.ClusterOperator{}, builder.WithPredicates(clusterOperatorPredicates()))
+
+	return build.Complete(r)
+}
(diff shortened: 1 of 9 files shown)


JoelSpeed:
/approve
/lgtm

SGTM

damdo:
/hold
We need to make sure the ClusterObject conditions are updated even though the CAPI controllers do not run.

damdo:
@JoelSpeed I've added the changes we discussed in our chat. PTAL

damdo:
/hold
For pre-merge testing of OCPBUGS-28539

pkg/controllers/unsupported/unsupported_controller_test.go
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+package unsupported
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("CAPI unsupported controller", func() {
```
JoelSpeed, line 23: No tests?

pkg/controllers/unsupported/unsupported_controller.go
```diff
@@ -0,0 +1,57 @@
...
+	operatorstatus.ClusterOperatorStatusClient
+	Scheme *runtime.Scheme
+}
+
+// Reconcile reconciles the cluster-api ClusterOperator object.
+func (r *CapiUnsupportedController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithName("CapiUnsupportedController")
+	log.Info(fmt.Sprintf("Reconciling %q ClusterObject", controllers.ClusterOperatorName))
+
+	if err := r.ClusterOperatorStatusClient.SetStatusAvailable(ctx); err != nil {
```
JoelSpeed, line 44: I assume there's no way to set the status available with a message saying that we are running on an unsupported platform?

pkg/controllers/unsupported/unsupported_controller.go
```diff
@@ -0,0 +1,57 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-capi-operator/pkg/controllers"
+	"github.com/openshift/cluster-capi-operator/pkg/operatorstatus"
+)
+
+// CapiUnsupportedController on unsupported platforms watches and keeps the cluster-api ClusterObject up to date.
+type CapiUnsupportedController struct {
```
JoelSpeed, line 34: Unsupported, or NoOp?

damdo: I had it NoOp and changed it to Unsupported, so it reads as CAPI Unsupported.
I think the latter is more meaningful and explicit in telling we are running in an unsupported mode.
The NoOp is quite vague and won't be true if in the future we decide to do "some op" in this controller.

pkg/controllers/unsupported/unsupported_controller_test.go
damdo: I wanted to make sure the approach was correct before starting to write tests. If we are happy with the controller, I can write some 👍

pkg/controllers/unsupported/unsupported_controller.go
JoelSpeed: I guess my concern is whether a customer would read this and say, why is there something unsupported on my cluster. Since we use that term frequently to tell customers what they can and cannot do.

pkg/controllers/unsupported/unsupported_controller.go
damdo: We could potentially use the `message` field in the `type: Available` condition? 🤔 
Not sure if that would break other assuptions?
```
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2024-01-26T13:27:00Z",
                        "message": "Cluster CAPI Operator is available at 4.16.0-0.ci.test-2024-01-26-124620-ci-op-8kd6z0s4-latest",
                        "reason": "AsExpected",
                        "status": "True",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2024-01-26T13:27:00Z",
                        "reason": "AsExpected",
                        "status": "False",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2024-01-26T13:27:00Z",
                        "reason": "AsExpected",
                        "status": "False",
                        "type": "Degraded"
                    },
                    {
                        "lastTransitionTime": "2024-01-26T13:27:00Z",
                        "reason": "AsExpected",
                        "status": "True",
                        "type": "Upgradeable"
                    }
```

pkg/controllers/unsupported/unsupported_controller.go
JoelSpeed: Stick with it as it is for now, but I think having the cluster operator saying something is probably a good idea. Maybe more down the lines of "Cluster API is not implemented on this platform"

damdo: Yes I agree with this last point, maybe something like? https://github.com/openshift/cluster-capi-operator/pull/157#discussion_r1469352816

pkg/controllers/unsupported/unsupported_controller.go
JoelSpeed: I think message in available is the correct place, if you have concerns though, check with the OTA team

damdo: I checked with them and they said it is the intended usage. I've update the PR to use it in the unsupported case to notify the user about it 👍

pkg/controllers/unsupported/unsupported_controller_test.go
damdo: Tests added

JoelSpeed (commented):
/lgtm
/approve

PR #159, opened 2024-02-14
OCPBUGS-29476: manifests-gen: store core capi crds also in their own manifest

damdo:
Store Core CAPI CRDs also in their own manifest to get them applied by CVO directly. We want these to be installed independently from whether the cluster-capi-operator is enabled, as other Openshift components rely on them.

### DIFF
--- a/manifests-gen/customizations.go
+++ b/manifests-gen/customizations.go
@@ -18,11 +18,8 @@ import (
 type resourceKey string
 
 const (
-	crdKey        resourceKey = "crds"
-	otherKey      resourceKey = "other"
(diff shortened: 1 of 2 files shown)


JoelSpeed (commented):
Discussed on call, would like to make sure we don't duplicate the CRDs between the manifests generically and the transport configmaps, we don't want CVO and CCO fighting over the two manifests

damdo:
@JoelSpeed now updated to store CRDs only in transport ConfigMap or in their own plain manifest.

JoelSpeed (commented):
/lgtm

PR #169, opened 2024-04-12
WIP: Add InfraCluster controller for object generation

damdo:


pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,189 @@
...
+	if r.Infra.Status.PlatformStatus == nil {
+		return fmt.Errorf("infrastructure PlatformStatus should not be nil: %w", err)
+	}
+
+	awsCluster := &awsv1.AWSCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      defaultInfraClusterName,
+			Namespace: framework.CAPINamespace,
+		},
+		Spec: awsv1.AWSClusterSpec{
```
JoelSpeed, line 93: What else goes in this spec that we could set from here? Are there other sources of information we can leverage to populate this?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to parse apiUrl port: %w", err)
+	}
+
+	if r.Infra.Status.PlatformStatus == nil {
+		return fmt.Errorf("infrastructure PlatformStatus should not be nil: %w", err)
+	}
+
+	awsCluster := &awsv1.AWSCluster{
+		ObjectMeta: metav1.ObjectMeta{
```
JoelSpeed, line 89: Do we need to annotate this so that it is not managed by the infracluster controller?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,189 @@
...
+		Spec: awsv1.AWSClusterSpec{
+			Region: r.Infra.Status.PlatformStatus.AWS.Region,
+			ControlPlaneEndpoint: clusterv1.APIEndpoint{
+				Host: apiUrl.Hostname(),
+				Port: int32(port),
+			},
+		},
+	}
+
+	if err := r.Create(ctx, awsCluster); err != nil {
```
JoelSpeed, line 102: What happens if the resource already exists?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,298 @@
...
+
+// Reconcile reconciles the cluster-api ClusterOperator object.
+func (r *InfraClusterController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithName("InfraClusterController")
+
+	res, err := r.reconcile(ctx, log)
+	if err != nil {
+		if err := r.setDegradedCondition(ctx, log); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set conditions for InfraCluster controller: %w", err)
+		}
```
damdo, line 67: Not sure we want to degrade at every error? 🤔

pkg/controllers/infracluster/infracluster_controller.go
damdo: This PR was still WIP back then. This comment was addressed in the implementation since then.

pkg/controllers/infracluster/infracluster_controller.go
damdo: This PR was still WIP back then. This comment was addressed in the implementation since then.

pkg/controllers/infracluster/infracluster_controller.go
damdo: This PR was still WIP back then. This comment was addressed in the implementation since then.

pkg/controllers/infracluster/watch_predicates.go
```diff
@@ -0,0 +1,24 @@
+package infracluster
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
```
damdo, line 10: TODO: add a watch for infraClusters ?

damdo:
/hold
Until all TODOs are handled

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	case configv1.VSpherePlatformType:
+		log.Info(fmt.Sprintf("detected platform %q is supported, but not implemented yet", r.Platform))
+	case configv1.OpenStackPlatformType:
+		log.Info(fmt.Sprintf("detected platform %q is supported, but not implemented yet", r.Platform))
+	default:
+		log.Info(fmt.Sprintf("detected platform %q is not supported, skipping capi controllers setup", r.Platform))
+	}
+
+	// At this point, the InfraCluster exists.
+	managedByAnnotationVal, found := r.InfraCluster.GetAnnotations()[clusterv1.ManagedByAnnotation]
```
JoelSpeed, line 95: I'm surprised that the InfraCluster is a member of the reconciler 🤔

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+		if isReady {
+			// The Infrastructure for this CAPI Cluster is already ready - nothing to do.
+			return ctrl.Result{}, nil
+		}
+
+		infraClusterCopy, ok := r.InfraCluster.DeepCopyObject().(client.Object)
+		if !ok {
+			return ctrl.Result{}, fmt.Errorf("unable to convert InfraCluster runtime.Object to client.Object")
+		}
```
JoelSpeed, line 118: If the infraCluster was returned by the `ensure` function, you could have it as a `client.Object` without the need for any additional deepcopying, since CR always returns copies anyway

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,288 @@
...
+// ensureAWSCluster ensures the AWSCluster cluster object exists.
+func (r *InfraClusterController) ensureAWSCluster(ctx context.Context) error {
+	target := awsv1.AWSCluster{ObjectMeta: metav1.ObjectMeta{
+		Name:      defaultInfraClusterName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	// Checking whether AWSCluster object exists. If it doesn't, create it.
+	//nolint:nestif
+	if err := r.Get(ctx, client.ObjectKeyFromObject(&target), &target); err != nil && errors.IsNotFound(err) {
```
JoelSpeed, line 153: Would suggest inverting some of the conditions here to remove indentation from the "Happy path"

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,288 @@
...
+		log.Info(fmt.Sprintf("InfraCluster is annotated with an unrecognized externally managed annotation value %q"+
+			"- skipping as not managed by this controller.", managedByAnnotationVal))
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// ensureAWSCluster ensures the AWSCluster cluster object exists.
+func (r *InfraClusterController) ensureAWSCluster(ctx context.Context) error {
+	target := awsv1.AWSCluster{ObjectMeta: metav1.ObjectMeta{
```
JoelSpeed, line 146: Typically you declare as a pointer here rather than taking the address several times through this function

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	build := ctrl.NewControllerManagedBy(mgr).
+		For(&configv1.ClusterOperator{}, builder.WithPredicates(clusterOperatorPredicates()))
+		// TODO: write a function that based the Infra returns a watch to the cloud specific InfraClusters objects.
+
+	return build.Complete(r)
+}
+
+// setManagedByAnnotation is setting the ManagedBy Annotation.
+// When this annotation is set, Cluster API will ignore the InfraCluster object and assume it is managed by an external entity.
+func setManagedByAnnotation(annotations map[string]string) map[string]string {
```
JoelSpeed, line 246: Maps are pointers so I don't think you need to return and reset this like this

pkg/controllers/infracluster/infracluster_controller_test.go
```diff
@@ -0,0 +1,266 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	awsv1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var _ = Describe("InfraCluster", func() {
+	ocpInfraAWS := configv1.Infrastructure{
```
JoelSpeed, line 20: Could use the builder pattern from actuator package test utils

pkg/controllers/infracluster/infracluster_controller_test.go
```diff
@@ -0,0 +1,266 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      defaultInfraClusterName,
+			Namespace: defaultCAPINamespace,
+			Annotations: map[string]string{
+				clusterv1.ManagedByAnnotation: managedByAnnotationValueClusterCAPIOperator,
+			},
+		},
+	}
+
+	// When the annotation is present and the value is empty, it is CAPI owned. When the annotation is
```
JoelSpeed, line 43: Why empty for us owning it, shouldn't we have a unique value?

pkg/controllers/infracluster/infracluster_controller_test.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	Context("When there is no InfraCluster", func() {
+		var ic *awsv1.AWSCluster
+		BeforeEach(func() {
+			ic = infraClusterWithExternallyManagedByAnnotation.DeepCopy()
+			mustReconcile(ocpInfraAWS, ic, true)
+		})
+		AfterEach(func() {
+			Expect(test.CleanupAndWait(ctx, cl, ic)).To(Succeed())
```
JoelSpeed, line 70: The argument for the `ic` here should just be an empty version of the object, you don't need to pass specific objects. It is designed to clean up all instances of the type

pkg/controllers/infracluster/infracluster_controller_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	r := &InfraClusterController{
+		ClusterOperatorStatusClient: operatorstatus.ClusterOperatorStatusClient{
+			Client: cl,
+		},
+		Infra:    ocpInfra,
+		Platform: ocpInfra.Status.PlatformStatus.Type,
+	}
+
+	_, err := r.Reconcile(ctx, reconcile.Request{
```
JoelSpeed, line 258: Why are we calling reconcile manually rather than setting it up with the manager and using asynchronous assertions?
We've seen in the past that single reconciles can be fragile as there's no retry logic. Prefer to run the manager for each test and trigger the reconcile via the watches

pkg/controllers/infracluster/infracluster_controller_test.go
```diff
@@ -0,0 +1,266 @@
...
+					},
+				}
+				Expect(cl.Get(ctx, types.NamespacedName{Namespace: actual.Namespace, Name: actual.Name}, actual)).To(Succeed())
+				Expect(actual.Status.Ready).To(BeFalse(),
+					"should have not changed the Status.Ready field")
+			})
+		})
+	})
+
+	Context("When there is an InfraCluster with an externally ManagedBy Annotation with non cluster-capi-operator value", func() { //nolint:dupl
```
JoelSpeed, line 138: I normally disable dupl for test suites btw

pkg/controllers/infracluster/suite_test.go
```diff
@@ -0,0 +1,66 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Controller Suite")
+}
+
+var _ = BeforeSuite(func() {
+	logf.SetLogger(klog.Background())
+
+	By("bootstrapping test environment")
+	var err error
+	testEnv = &envtest.Environment{}
```
JoelSpeed, line 52: No CRDs to install?

pkg/util/platform.go
```diff
@@ -13,19 +13,20 @@ const (
...
+func GetPlatform(ctx context.Context, infra configv1.Infrastructure) (configv1.PlatformType, error) {
 	if infra.Status.PlatformStatus == nil {
 		return "", nil
 	}
 
 	return infra.Status.PlatformStatus.Type, nil
 }
+
+func GetInfra(ctx context.Context, cl client.Reader) (configv1.Infrastructure, error) {
+	infra := configv1.Infrastructure{}
```
JoelSpeed, line 25: Normally it's a pointer. It's idiomatic in K8s to pass around pointer to your object. It means that it satisfies the interfaces of `client.Object` 
```suggestion
	infra := &configv1.Infrastructure{}
```

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	Platform     configv1.PlatformType
+	Infra        configv1.Infrastructure
+	InfraCluster client.Object
+}
+
+// Reconcile reconciles the cluster-api ClusterOperator object.
+func (r *InfraClusterController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithName("InfraClusterController")
+
+	res, err := r.reconcile(ctx, log)
```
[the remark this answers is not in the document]
damdo: Yeah, this is not following that convention because we don't actually have a concrete type to get, but rather before knowing what object to get we need to switch over the platform and then kick off the specific reconciliation function for the concrete InfraCluster type (e.g. `reconcileAWSCluster()`).
So we came up with this which is the cleanest way of reconciling this that we could think at the time.

pkg/controllers/infracluster/infracluster_controller.go
damdo: Yes, that was to simplify the code but it is not good practice as this could be reused and be stale. We are changing this.

pkg/controllers/infracluster/infracluster_controller.go
damdo: Yeah +1

pkg/controllers/infracluster/infracluster_controller.go
damdo: 👍

pkg/controllers/infracluster/infracluster_controller.go
damdo: Yeah, we can apply annotations inline when we create the infraCluster directly.

pkg/controllers/infracluster/infracluster_controller_test.go
damdo: Up until now, we only set the key of the `clusterv1.ManagedByAnnotation`. So this is consistent with the existing approach.
We could change this of course, but then there might be upgrade consequences. WDYT?

pkg/controllers/infracluster/infracluster_controller_test.go
damdo: It is not how `test.CleanupAndWait` works as this is a local package.
What you are describing is `cluster-api-actuator-pkg`'s `testutils.CleanupResources` behaviour, which also deletes the namespace after deleting the resource belonging to it. We refactored to use that instead.

pkg/controllers/infracluster/suite_test.go
damdo: `test.StartEnvTest` installs CRDs centrally

pkg/controllers/infracluster/infracluster_controller_test.go
damdo: Switched to using the manager.

damdo:
@JoelSpeed feedback addressed.

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,265 @@
...
+
+	return infraCluster, nil
+}
+
+// reconcile performs the main business logic for installing Cluster API components in the cluster.
+// Notably it fetches CAPI providers "transport" ConfigMap(s) matching the required labels,
+// it extracts from those ConfigMaps the embedded CAPI providers manifests for the components
+// and it applies them to the cluster.
+//
+//nolint:unparam
```
JoelSpeed, line 96: Don't think you need unparam now?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	// Controller conditions for the Cluster Operator resource
+	InfraClusterControllerAvailableCondition = "InfraClusterControllerAvailable"
+	InfraClusterControllerDegradedCondition  = "InfraClusterControllerDegraded"
+
+	defaultCAPINamespace    = "openshift-cluster-api"
+	defaultInfraClusterName = "cluster"
+	clusterOperatorName     = "cluster-api"
+	// This is the managedByAnnotation value that this controller sets by default when it creates an InfraCluster object.
+	// If the managedByAnnotation key is set, and it has this as the value, it means this controller is managing the InfraCluster.
+	managedByAnnotationValueClusterCAPIOperator = ""
```
JoelSpeed, line 39: Can we change this to something before we ship CAPI as GA?

pkg/controllers/infracluster/infracluster_controller_test.go
JoelSpeed: Upgrade for? We don't support anything in CAPI right now right? So changing it, yeah we would have to change our docs etc, but, wouldn't/shouldn't break anyone in production yet?

pkg/controllers/infracluster/infracluster_controller_test.go
```diff
@@ -0,0 +1,234 @@
...
+		})
+	})
+})
+
+func mustPatchAWSInfraClusterReadiness(awsInfraCluster *awsv1.AWSCluster, readiness bool) {
+	patchBase := client.MergeFrom(awsInfraCluster.DeepCopy())
+	awsInfraCluster.Status = awsv1.AWSClusterStatus{
+		Ready: readiness,
+	}
+	Expect(cl.Status().Patch(ctx, awsInfraCluster, patchBase)).To(Succeed(), "should have been able to patch AWSCluster")
```
JoelSpeed, line 192: What about using eventually and the komega UpdateStatus concept, would that be more stable here or is patch stable enough?

pkg/controllers/infracluster/suite_test.go
```diff
@@ -0,0 +1,111 @@
...
+	Expect(cl).NotTo(BeNil())
+
+	// Set Ginkgo default timeouts
+	Default.SetDefaultEventuallyTimeout(60 * time.Second)
+	Default.SetDefaultEventuallyPollingInterval(1 * time.Second)
+
+	komega.SetClient(cl)
+	komega.SetContext(ctx)
+
+	fmt.Println(CreateKubeconfigFileForRestConfig(*cfg))
```
JoelSpeed, line 75: Why are we printing this out?

pkg/controllers/infracluster/infracluster_controller.go
damdo: ok, I think I can change it to `cluster-capi-operator-infracluster-controller`

pkg/controllers/infracluster/suite_test.go
damdo: Left it in for debugging reasons, will remove.

pkg/controllers/infracluster/infracluster_controller_test.go
damdo: Yeah, that's better 👍 will change

JoelSpeed:
Changes look good, do we want to squash the commits at all? Anything else you'd like to do before merging?

damdo:
Commits squashed. So this is now ready for labelling @JoelSpeed

JoelSpeed:
Have seen outstanding feedback noted in tracker cards in Jira
* Preventing the reboot a couple of minutes after startup
* Looking into degraded conditions

/approve
/lgtm

PR #170, opened 2024-04-19
OCPCLOUD-2584: Implement compression of components in configMap

RadekManak:
This PR updates the manifest generator to compress the components in the infrastructure provider manifest configmap. The components are only compressed when their size is over the 1MiB configMap size limit.

When compressed, the components are stored in:
```
binaryData:
  components-zstd: <binary>
```
and when uncompressed, they stay in plaintext form of:
```
Data:
  components: <text>
```

The capi installer controller is updated to handle the compressed variant of the configMap.

To help with code review on repos that compress the configMap I have made the generator write the modified manifests into infrastructure-components-openshift.yaml in plainText form. This file is not consumed by anything.

The "registry.ci.openshift.org/openshift:kube-rbac-proxy" imagestream  was not being replaced by the correct image from the capi-images configmap, so I fixed that. Maybe this was previously only on manifests deployed by CVO?


nrb:
So far this is looking good, pending fixing the lints and adding tests.

damdo:
> To help with code review on repos that compress the configMap I have made the generator write the modified manifests into infrastructure-components-openshift.yaml in plainText form. This file is not consumed by anything.

Great idea!

manifests-gen/providers.go
```diff
@@ -102,17 +104,52 @@ func (p *provider) writeProviderComponentsToManifest(fileName string, objs []uns
...
 }
 
-// writeProviderComponentsConfigmap allows to write provider components to the provider (transport) ConfigMap.
-func (p *provider) writeProviderComponentsConfigmap(fileName string, objs []unstructured.Unstructured) error {
-	combined, err := utilyaml.FromUnstructured(objs)
+// compressToZstd compressses bytes using zstd
+func (p *provider) compressToZstd(data []byte) (bytes.Buffer, error) {
+	var compressed bytes.Buffer
+
+	writer, err := zstd.NewWriter(&compressed, zstd.WithEncoderLevel(zstd.EncoderLevelFromZstd(19)))
```
damdo, line 111: Could we use the [exposed constants](https://github.com/klauspost/compress/blob/7ae2138b16cc43afcea3ce7d3d2f2625fb389d51/zstd/encoder_options.go#L149-L170) instead of the number here?

Also I'd use the fastest option, to improve decompression time, as we will do it quite often during reconciles

manifests-gen/providers.go
```diff
@@ -141,6 +182,53 @@ func (p *provider) writeProviderComponentsConfigmap(fileName string, objs []unst
...
+		}
+	} else {
+		cm.Data = map[string]string{"components": string(combined)}
+	}
+
+	return cm, nil
+}
+
+// compressToZstd compressses bytes using zstd
+func compressToZstd(data []byte) (bytes.Buffer, error) {
```
damdo, line 210: Is this an exact copy of the (p *provider) compressToZstd above?

manifests-gen/providers.go
```diff
@@ -102,17 +104,52 @@ func (p *provider) writeProviderComponentsToManifest(fileName string, objs []uns
...
 	}
 
+	other, err := utilyaml.FromUnstructured(resourceMap[otherKey])
+	if err != nil {
+		return fmt.Errorf("error converting unstructured object to YAML: %w", err)
+	}
+
+	combined := utilyaml.JoinYaml(crds, other)
+
+	return os.WriteFile(path.Join(*basePath, "openshift", "infrastructure-components-openshift.yaml"), ensureNewLine(combined), 0600)
```
damdo, line 144: Can we make this filename a constant and also add a comment similar to the one you left in the description re: it being used only for reviewing purposes? Thanks!

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -329,19 +331,18 @@ func (r *CapiInstallerController) extractProviderComponents(cm corev1.ConfigMap)
 	// Certain provider components have drone/envsubst environment variables interpolated within the manifest.
 	// Substitute them with the value defined in the environment variable (see setFeatureGatesEnvVars()).
 	// If that's not set, fallback to the default value defined in the template.
```
damdo, line 333: Can you move this comment down with the envsubst.EvalEnv ?

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -351,6 +352,40 @@ func (r *CapiInstallerController) extractProviderComponents(cm corev1.ConfigMap)
 	return replacedYamlManifests, nil
 }
 
+// extractManifests extracts and processes component manifest from given confimap.
```
damdo, line 355:
```suggestion
// extractManifests extracts and processes component manifests from given ConfigMap.
```

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -329,19 +331,18 @@ func (r *CapiInstallerController) extractProviderComponents(cm corev1.ConfigMap)
 	// Certain provider components have drone/envsubst environment variables interpolated within the manifest.
 	// Substitute them with the value defined in the environment variable (see setFeatureGatesEnvVars()).
 	// If that's not set, fallback to the default value defined in the template.
-	components, err := envsubst.EvalEnv(cm.Data["components"])
+
+	yamlManifests, err := extractManifests(cm)
 	if err != nil {
-		return nil, fmt.Errorf("failed to substitute env vars in CAPI manifests: %w", err)
+		return nil, fmt.Errorf("failed to extract manifests from configMap: %w", err)
```
damdo, line 337: Could you also throw in this line the `namespace/name` for this configmap for easier debugging?

damdo (commented):
Thanks for the patience @RadekManak 
Mostly LGTM, left a couple of comments

manifests-gen/providers.go
RadekManak: Changed to SpeedBestCompression. In zstd the compression level does not impact decompression speed.

manifests-gen/providers.go
RadekManak: Removed the other one. I think it accidentally came back during rebase.

pkg/controllers/capiinstaller/capi_installer_controller.go
RadekManak: This already had the ConfigMap name up the error chain. I've added namespace to it.

manifests-gen/providers.go
damdo: Gotcha, thanks for clarifying!

manifests-gen/providers.go
damdo: TY

pkg/controllers/capiinstaller/capi_installer_controller.go
damdo: Ok makes sense, thanks

damdo (commented):
/approve

I'll add LGTM later if you want to fix that string thing

PR #172, opened 2024-05-01
OCPBUGS-33170: All containers must fallback to logs on error

JoelSpeed:
In 4.17, we are enforcing this value for all OpenShift containers via tests, link in bug comment. This helps with debugging to ensure we have logs available.

### DIFF
--- a/manifests-gen/customizations.go
+++ b/manifests-gen/customizations.go
@@ -234,6 +234,9 @@ func customizeDeployments(obj *unstructured.Unstructured) {
 		if container.Name == "kube-rbac-proxy" {
 			container.Image = "registry.ci.openshift.org/openshift:kube-rbac-proxy"
 		}
+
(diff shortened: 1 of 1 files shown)


nrb:
/lgtm

I'd like for @damdo to do final approval, since I don't understand the full impact of this change - what do we need to run in order to get all the manifests updated? A PR per provider to update [these manifest ConfigMaps](https://github.com/openshift/cluster-api-provider-aws/blob/master/openshift/manifests/0000_30_cluster-api_04_cm.infrastructure-aws.yaml)?

damdo (commented):
/approve

damdo:
Let’s first merge the cluster-capi-operator change to the manifests-gen and then open a PR to vendor/exec generation for each provider.

damdo:
This manifests-gen change does not affect the payload as it is a manually ran tool

PR #173, opened 2024-05-13
OCPCLOUD-2568: Add MAPI->CAPI and CAPI->MAPI conversion packages, add AWS implementation

damdo:
Adds mapi <-> capi conversion pkgs to aid conversion between MAPI and CAPI resources and the other way around.


damdo:
/cc @JoelSpeed 
Ready for a first pass review

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/utils/ptr"
+	capav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+var (
+	errCAPIMachineAWSMachineTemplateAWSClusterCannotBeNil = errors.New("provided Machine and AWSMachineTemplate and AWSCluster can not be nil")
```
JoelSpeed, line 37:
```suggestion
	errCAPIMachineAWSMachineTemplateAWSClusterCannotBeNil = errors.New("provided Machine, AWSMachineTemplate and AWSCluster can not be nil")
```

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	errUnsupportedCAPATenancy                             = errors.New("unable to convert unsupported CAPA Tenancy")
```
JoelSpeed, line 38: CAPA tenancy? Probably best to make it more about how the user would view this. We are unable to convert because this is not supported in Machine API, so maybe a message about field not supported in legacy Machine API implementation is better?

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	errUnsupportedHTTPTokensState                         = errors.New("unable to convert unsupported HTTPTokens State")
+)
+
+// MachineAndAWSMachineTemplateAndAWSCluster stores the details of a Cluster API Machine and AWSMachineTemplate and AWSCluster.
+type MachineAndAWSMachineTemplateAndAWSCluster struct {
+	Machine    *capiv1.Machine
+	Template   *capav1.AWSMachineTemplate
+	AWSCluster *capav1.AWSCluster
+}
```
JoelSpeed, line 47: Do we want this to be public, or rather, should it be private behind an interface and use a constructor as you have done?

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	// The use of ARN and Filters to reference AMIs was present
+	// in CAPA but has been deprecated and then removed
+	// ref: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3257
+	mapaProviderConfig.AMI.ID = m.Template.Spec.Template.Spec.AMI.ID
+
+	mapaProviderConfig.TypeMeta = metav1.TypeMeta{
+		Kind: "AWSMachineProviderConfig",
+		// In the machineSets both "awsproviderconfig.openshift.io/v1beta1" and "machine.openshift.io/v1beta1" can be found.
+		// Here we always settle on one of the two.
+		APIVersion: "awsproviderconfig.openshift.io/v1beta1",
```
JoelSpeed, line 95: This should be `machine.openshift.io/v1beta1` which is the newer of the two

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	mapaProviderConfig.PublicIP = m.Template.Spec.Template.Spec.PublicIP
+	mapaProviderConfig.Placement = mapiv1.Placement{
+		AvailabilityZone: ptr.Deref(m.Machine.Spec.FailureDomain, ""),
+		Tenancy:          mapaTenancy,
+		Region:           m.AWSCluster.Spec.Region,
+	}
+
+	mapiAWSSecurityGroups := convertAWSSecurityGroupstoMAPI(m.Template.Spec.Template.Spec.AdditionalSecurityGroups)
+
+	mapaProviderConfig.SecurityGroups = mapiAWSSecurityGroups
```
JoelSpeed, line 115: I would probably cuddle these two together since they are related. Do we even need to declare them on separate lines?

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	warnings = append(warnings, warn...)
+
+	mapaProviderConfig.MetadataServiceOptions = mapiAWSMetadataOptions
+
+	mapaProviderConfig.UserDataSecret = &corev1.LocalObjectReference{
+		Name: ptr.Deref(m.Machine.Spec.Bootstrap.DataSecretName, "worker-user-data"),
+	}
+
+	// TODO: CredentialsSecret conversion.
+	// mapaProviderConfig.CredentialsSecret
```
JoelSpeed, line 148: Where does this come from within CAPA?

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	)
+
+	mapaSpec, warn, err := FromMachineAndAWSMachineTemplateAndAWSCluster(m.Machine, m.Template, m.AWSCluster).ToProviderSpec()
+	if err != nil {
+		errors = append(errors, err)
+	}
+
+	warnings = append(warnings, warn...)
+
+	mapiMachine := fromMachineToMachine(m.Machine)
```
JoelSpeed, line 175: Probably want to rename this func to be `fromCAPIMachineToMAPIMachine`

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+				Annotations: m.MachineSet.ObjectMeta.Annotations,
+			},
+		}, m.Template, m.AWSCluster).ToProviderSpec()
+	if err != nil {
+		errors = append(errors, err)
+	}
+
+	warnings = append(warnings, warn...)
+
+	mapiMachineSet, warn, err := FromMachineSetToMachineSet(m.MachineSet)
```
JoelSpeed, line 215: `FromCAPIMachineSetToMAPIMachineSet`?

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	var (
+		rawBytes []byte
+		err      error
+	)
+
+	if rawBytes, err = json.Marshal(spec); err != nil {
+		return nil, fmt.Errorf("error marshalling providerSpec: %w", err)
+	}
```
JoelSpeed, line 251: Cleaner as below?
```suggestion
	rawBytes, err := json.Marshal(spec)
	if err != nil {
		return nil, fmt.Errorf("error marshalling providerSpec: %w", err)
	}
```

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+
+	return &runtime.RawExtension{
+		Raw: rawBytes,
+	}, nil
+}
+
+// AWSProviderSpecFromRawExtension unmarshals a raw extension into an AWSMachineProviderSpec type.
+func AWSProviderSpecFromRawExtension(rawExtension *runtime.RawExtension) (mapiv1.AWSMachineProviderConfig, error) {
```
JoelSpeed, line 259: Where is this used?

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,432 @@
...
+	case capav1.HTTPTokensStateRequired:
+		auth = mapiv1.MetadataServiceAuthenticationRequired
+	default:
+		errors = append(errors, fmt.Errorf("%w: %q", errUnsupportedHTTPTokensState, capiMetadataOpts.HTTPTokens))
+	}
+
+	// TODO: lossy: These fields are not present in MAPI, so they will be lost in CAPI -> MAPI conversion.
+	// opts.HTTPEndpoint
+	// opts.HTTPPutResponseHopLimit
+	// opts.InstanceMetadataTags
```
JoelSpeed, line 298: If any of these are present, they need to error out as the feature is not support in MAPA

pkg/capi2mapi/aws_test.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/utils/ptr"
+	capav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+var _ = Describe("capi2mapi AWS", Ordered, func() {
```
JoelSpeed, line 27: Why ordered?

pkg/capi2mapi/aws_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	It("should be able to convert a CAPI Machine and AWSMachineTemplate to a MAPI AWSProviderSpec", func() {
+		awsProviderSpecConfig, warns, err :=
+			FromMachineAndAWSMachineTemplateAndAWSCluster(awsMachine, awsMachineTemplate, capiAWSCluster).ToProviderSpec()
+		Expect(awsProviderSpecConfig).To(Not(BeNil()), "should not have a nil MAPI ProviderSpecConfig")
+		Expect(err).ToNot(HaveOccurred(), "should have been able to convert AWSMachineTemplateSpec to AWSProviderSpec")
+		Expect(warns).To(BeEmpty(), "should have not warned while converting AWSMachineTemplateSpec to AWSProviderSpec")
+	})
+
+	It("should be able to convert a MAPI Machine to a CAPI Machine", func() {
```
JoelSpeed, line 77: I think the MAPI and CAPI are inverted here

pkg/capi2mapi/aws_test.go
```diff
@@ -0,0 +1,92 @@
...
+	. "github.com/onsi/gomega"
+
+	"k8s.io/utils/ptr"
+	capav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+var _ = Describe("capi2mapi AWS", Ordered, func() {
+
+	awsMachineTemplate := &capav1.AWSMachineTemplate{
```
JoelSpeed, line 29: I think we need to exhaust this template and populate as many fields as we can convert, and then also add some negative cases around the lossy versions.

Would also be good to have a fully populated expected spec, rather than just checking against nil

pkg/capi2mapi/machine.go
```diff
@@ -0,0 +1,74 @@
...
+			Annotations: capiMachine.Annotations,
+		},
+		Spec: mapiv1.MachineSpec{
+			ObjectMeta: mapiv1.ObjectMeta{
+				Name:         capiMachine.ObjectMeta.Name,
+				GenerateName: capiMachine.ObjectMeta.GenerateName,
+				Namespace:    capiMachine.ObjectMeta.Namespace,
+				//Labels:          capiMachine.ObjectMeta.Labels,
+				Annotations:     capiMachine.ObjectMeta.Annotations,
+				OwnerReferences: capiMachine.OwnerReferences,
```
JoelSpeed, line 46: These fields should not be populated from here, there's an objectmeta with the spec of the CAPI machine that this information should be coming from

pkg/capi2mapi/machine.go
```diff
@@ -0,0 +1,74 @@
...
+			},
+			ProviderID: capiMachine.Spec.ProviderID,
+
+			// LifecycleHooks: // TODO: lossy: find alternative in CAPI for this.
```
JoelSpeed, line 50: These are implemented as annotations in CAPI

pkg/capi2mapi/machine.go
```diff
@@ -0,0 +1,74 @@
...
+			// Taints: // TODO: lossy: Not Present on CAPI Machines, only done via BootstrapProvider?
```
JoelSpeed, line 51: Yeah this could be an issue 🤔 We may need to create a very minimal bootstrap provider that handles taints and perhaps other synchronisation to maintain compatibility

pkg/mapi2capi/machineset.go
```diff
@@ -0,0 +1,58 @@
...
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+const (
+	capiMachineSetAPIVersion = "cluster.x-k8s.io/v1beta1"
+	capiMachineSetKind       = "MachineSet"
+)
+
+// FromMachineSetToMachineSet takes a MAPI MachineSet and returns a converted CAPI MachineSet.
+func FromMachineSetToMachineSet(mapiMachineSet *mapiv1.MachineSet) (capiv1.MachineSet, []string, error) {
```
JoelSpeed, line 32: Should this be public if it has restrictions of higher level functions, I think it and it's CAPI counterpart need to be private

pkg/mapi2capi/machineset.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	capiMachineSet.Spec.Selector = mapiMachineSet.Spec.Selector
+	capiMachineSet.Spec.Template.Labels = mapiMachineSet.Spec.Template.Labels
+	// capiMachineSet.Spec.ClusterName // populated by higher level functions
+	capiMachineSet.Spec.Replicas = mapiMachineSet.Spec.Replicas
+	// capiMachineSet.Spec.Template.Spec.ClusterName // populated by higher level functions
+	capiMachineSet.Spec.Template.Spec.InfrastructureRef = corev1.ObjectReference{
+		APIVersion: awsTemplateAPIVersion,
+		Kind:       awsTemplateKind,
+		Name:       mapiMachineSet.Name,
```
JoelSpeed, line 52: This is not always going to be true, the name of the template has to be based on the content of the template itself, so this will be dynamic, guessing it will need to be set by something higher level

pkg/mapi2capi/machine.go
```diff
@@ -0,0 +1,79 @@
...
+			ProviderID: mapiMachine.Spec.ProviderID,
+
+			// Version: not necessary (optionally used by bootstrap providers).
+			// FailureDomain: populated by higher level functions.
+			// ClusterName: populated by higher level functions.
+
+			// TODO: figure out if/how to convert these.
+			// NodeDrainTimeout: ,
+			// NodeVolumeDetachTimeout: ,
+			// NodeDeletionTimeout: ,
```
JoelSpeed, line 57: How do we deal with fields that exist only within MAPI here, eg, I would expect an error presently for a MAPI machine that has taints, but I don't see where that is

pkg/mapi2capi/aws.go
```diff
@@ -0,0 +1,431 @@
...
+	capav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	errInfrastructureInfrastructureNameCannotBeNil  = errors.New("infrastructure cannot be nil and infrastructure.Status.InfrastructureName cannot be empty")
+	errUnableToFindAMIReference                     = errors.New("unable to find a valid AMI resource reference")
+	errUnableToConvertUnsupportedAMIFilterReference = errors.New("unable to convert AMI Filters reference. Not supported in CAPI")
+	errUnableToConvertUnsupportedAMIARNReference    = errors.New("unable to convert AMI ARN reference. Not supported in CAPI")
```
JoelSpeed, line 38: Eventually we will need to do something to resolve these and convert them to IDs to allow this to be converted, but, not yet :)

pkg/mapi2capi/aws.go
```diff
@@ -0,0 +1,431 @@
...
+type AWSMachineAndInfra struct {
+	Machine        *mapiv1.Machine
+	Infrastructure *configv1.Infrastructure
+}
+
+// AWSMachineSetAndInfra stores the details of a Machine API AWSMachine and Infra.
+type AWSMachineSetAndInfra struct {
+	MachineSet     *mapiv1.MachineSet
+	Infrastructure *configv1.Infrastructure
+}
```
JoelSpeed, line 57: Should be private, expose interfaces not the types directly

pkg/mapi2capi/aws.go
```diff
@@ -0,0 +1,431 @@
...
+		errs = append(errs, err)
+	}
+
+	spec := capav1.AWSMachineTemplateSpec{
+		Template: capav1.AWSMachineTemplateResource{
+			Spec: capav1.AWSMachineSpec{
+				AMI:                      capiAWSAMIReference,
+				AdditionalSecurityGroups: capiAdditionalSecurityGroups,
+				AdditionalTags:           capiAdditionalAWSTags,
+				// CloudInit. Not defined as we use ignition in OpenShift.
```
JoelSpeed, line 210: For fields not supported, we should prevent these fields from being present in these resources

pkg/mapi2capi/aws.go
```diff
@@ -0,0 +1,431 @@
...
+					StorageType: capav1.IgnitionStorageTypeOptionUnencryptedUserData, // Hardcoded for OpenShift.
+				},
+				// ImageLookupBaseOS. Not used in OpenShift.
+				// ImageLookupFormat. Not used in OpenShift.
+				// ImageLookupOrg. Not used in OpenShift.
+
+				// TODO: what to do with instanceID? in MAPI that's not in AWSMachineProviderConfig but outside of it.
+				// Is this propagated down by the CAPA controller automatically from the CAPI Machine
+
+				// InstanceID. This is dynamically populated by the controller.
```
JoelSpeed, line 223: Is it required/will the controller attempt to create a new instance if this isn't populated?

pkg/mapi2capi/aws.go
```diff
@@ -0,0 +1,431 @@
...
+
+	if len(errs) > 0 {
+		return capav1.AWSMachineTemplateSpec{}, warnings, utilerrors.NewAggregate(errs)
+	}
+
+	return spec, warnings, nil
+}
+
+// AWSProviderSpecFromRawExtension unmarshals a raw extension into an AWSMachineProviderSpec type.
+func AWSProviderSpecFromRawExtension(rawExtension *runtime.RawExtension) (mapiv1.AWSMachineProviderConfig, error) {
```
JoelSpeed, line 253: Does this need to be public?

pkg/mapi2capi/aws_test.go
```diff
@@ -0,0 +1,142 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	mapiv1 "github.com/openshift/api/machine/v1beta1"
+	machinebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1beta1"
+
+	"k8s.io/utils/ptr"
+)
+
+var _ = Describe("mapi2capi AWS", Ordered, func() {
```
JoelSpeed, line 28: I don't think this needs to be ordered

JoelSpeed (commented):
This is a good start, you've covered the majority of the conversion, but I think the error cases/points where it's not obvious what to do will need continued discussion.

While reviewing, there are a few things I think we need to focus on here:
* For all cases where we cannot convert, return an error and create a Jira so that eventually the error is removed
* Ensure the conversions are injective, and report errors when a field cannot be converted because it does not exist on the other side
* Convert status
* Improve testing with negative cases and variations of fields

pkg/capi2mapi/aws_test.go
damdo: Taken from CPMS, no special reason.

pkg/capi2mapi/aws_test.go
damdo: Yes that's for sure!
Locally I've been running roundtrip conversion tests against real world MachineSets which include fully populated specs. I think we need to get that into a repeatable pipeline for presubmits. So we can continually test against those.

pkg/mapi2capi/machineset.go
damdo: +1, Yes, let's only use higher level functions to access these.

pkg/capi2mapi/aws_test.go
JoelSpeed: Lets talk about this separately, but, I think we should try to get some automated testing set up around this, likely will have to be done in a private repository somehow

pkg/mapi2capi/aws.go
damdo: So adding validation rules for CAPI Machine / CAPA AWSMachine & AWSMachineTemplate or really any CAP* <Provider>Machine and <Provider>MachineTemplate disallowing fields that we don't support/use on the OCP side.

pkg/mapi2capi/aws.go
damdo: I think this is useful also externally if one wants to extract the providerSpec from a MAPI MachineSet to then only feed that in the converter.

pkg/mapi2capi/aws.go
JoelSpeed: Exactly, in the first instance that may be quite restrictive, but we can potentially relax them down the path

pkg/mapi2capi/aws.go
JoelSpeed: Ack, I think maybe lets move all of these out to a utils package somewhere, need to make sure they are actually used in more than one place though

pkg/capi2mapi/aws.go
JoelSpeed: So these fields are required and always present, so we need to work out a suitable handling of this. Eg if the values are the defaults, then we should not error, if they are non default, we should error.

Should investigate how MAPI handles this, perhaps we want to add these on the MAPI side to save conversion

pkg/capi2mapi/aws_test.go
JoelSpeed, line 1: Some un-committed round-trip conversion is being used to do further testing. Idea here would be to use unit testing for cases where we know there are issues, and have some happy path of known good fields.

pkg/capi2mapi/aws.go
```diff
@@ -0,0 +1,428 @@
...
+	var (
+		errors   []error
+		warnings []string
+	)
+
+	mapaSpec, warn, err := FromMachineAndAWSMachineTemplateAndAWSCluster(
+		&capiv1.Machine{
+			Spec: m.MachineSet.Spec.Template.Spec,
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: m.MachineSet.ObjectMeta.Annotations,
```
JoelSpeed, line 206: I don't think this is the right objectmeta

pkg/mapi2capi/machine.go
damdo: Yes errors, see the Taints handling I added now, we can use that as an example.

pkg/conversion/mapi2capi/machine.go
```diff
@@ -0,0 +1,158 @@
...
+	return annotations
+}
+
+// handleUnsupportedMachineFields checks for fields that are not supported by CAPI and returns a list of errors.
+func handleUnsupportedMachineFields(spec mapiv1.MachineSpec) field.ErrorList {
+	var errs field.ErrorList
+
+	fldPath := field.NewPath("spec")
+
+	// ObjectMeta related fields should never get converted (aside from labels and annotations).
```
nrb, line 134: Are callers expected to clear these fields before handing the struct to the `fromMAPIMachineToCAPIMachine` function?

pkg/conversion/mapi2capi/machineset.go
```diff
@@ -0,0 +1,65 @@
...
+	mapiv1 "github.com/openshift/api/machine/v1beta1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+// fromMAPIMachineSetToCAPIMachineSet takes a MAPI MachineSet and returns a converted CAPI MachineSet.
+func fromMAPIMachineSetToCAPIMachineSet(mapiMachineSet *mapiv1.MachineSet) (*capiv1.MachineSet, utilerrors.Aggregate) {
```
nrb, line 28: I notice `MachineSet`s don't have a helper to check the `ObjectMeta` information. Should they?

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -0,0 +1,122 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	machinebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1beta1"
+
+	"k8s.io/utils/ptr"
+)
+
+var _ = Describe("mapi2capi AWS", func() {
+	// mapiProviderConfig := &mapiv1.AWSMachineProviderConfig{
```
nrb, line 28: Could we put this in a `testdata` JSON/YAML file instead of a comment and a builder? I worry that this comment will get out of date fast.

pkg/conversion/mapi2capi/machine.go
JoelSpeed: No, I'm not expecting users in their clusters to have these fields set, since they don't actually do anything if you set them (the controller ignores them).

So we return an error. They'll have to clear them if they want to proceed.

pkg/conversion/mapi2capi/machineset.go
JoelSpeed: Which ObjectMeta are you talking about? There are three levels of ObjectMeta in the MachineSet and it's only the one in spec.template.spec.metadata that needs checks, but that comes from the machine conversion level if you look at where this is called.

pkg/conversion/mapi2capi/aws_test.go
JoelSpeed: This was used in a test that I deleted, I just commented it out so that the compiler wouldn't complain while I was hacking.

I'll get rid of it I think since we aren't actually going to use it

theobarberbany (commented):
First pass looking over the `capi2mapi` stuff. Looks good to me, only a few questions (mostly for my benefit 😛 )

pkg/conversion/capi2mapi/aws.go
```diff
@@ -0,0 +1,492 @@
...
+			awsCluster: ac,
+		},
+	}
+}
+
+// ToProviderSpec converts a capi2mapi MachineAndAWSMachineTemplateAndAWSCluster into a MAPI AWSMachineProviderConfig.
+//
+//nolint:funlen
+func (m machineAndAWSMachineAndAWSCluster) toProviderSpec() (*mapiv1.AWSMachineProviderConfig, []string, error) {
+	if m.machine == nil || m.awsMachine == nil || m.awsCluster == nil {
```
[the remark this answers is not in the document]
JoelSpeed: They were here originally, and I'm happy to leave them here, it makes it obvious that future calls to the objects within don't need to check.

If they were further up the stack, you may not notice that this has already been checked right?

pkg/conversion/capi2mapi/aws.go
```diff
@@ -0,0 +1,492 @@
...
+		return nil, nil, errCAPIMachineAWSMachineAWSClusterCannotBeNil
+	}
+
+	var (
+		warnings []string
+		errors   field.ErrorList
+	)
+
+	fldPath := field.NewPath("spec")
```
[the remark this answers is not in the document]
JoelSpeed: You can see when we call the tenancy we pass `fldPath.Child` to child the path off as we pass it in

But yes, this is all about pass nicely pathed errors around, this is library used to construct API errors so the errors should look familiar

pkg/conversion/capi2mapi/aws.go
```diff
@@ -0,0 +1,492 @@
...
+		errors = append(errors, errs...)
+	}
+
+	warnings = append(warnings, warn...)
+
+	mapaProviderConfig := mapiv1.AWSMachineProviderConfig{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "AWSMachineProviderConfig",
+			// In the machineSets both "awsproviderconfig.openshift.io/v1beta1" and "machine.openshift.io/v1beta1" can be found.
+			// Here we always settle on one of the two.
```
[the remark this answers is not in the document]
JoelSpeed: This is just a string value, and yes, it could be an error in a round trip, because the object will have changed, but... I don't care 😅 

Nothing inspects the value of this field so it doesn't change anything functionally

pkg/conversion/capi2mapi/aws.go
```diff
@@ -0,0 +1,492 @@
...
+			APIVersion: "machine.openshift.io/v1beta1",
+		},
+		// ObjectMeta - TODO(OCPCLOUD-XXXX)
```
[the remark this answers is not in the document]
JoelSpeed: This one I didn't complete looking into. But now I'm thinking about it... I've never seen this objectmeta used. To my knowledge, it doesn't do anything 🤔 

But yes, will need to add a card to handle this, or just forbid its usage

pkg/conversion/capi2mapi/aws.go
```diff
@@ -0,0 +1,492 @@
...
+
+	if capiMetadataOpts == nil {
+		return mapiv1.MetadataServiceOptions{}, nil, nil
+	}
+
+	var auth mapiv1.MetadataServiceAuthentication
+
+	switch capiMetadataOpts.HTTPTokens {
+	case "":
+		//
```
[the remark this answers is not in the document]
JoelSpeed: Hmm, no, I think this one is complete as it were. I think this is just to show there should be no action taken when the value is the empty string.

That said, will need to double check that the behaviour is the same for empty string vs empty string

pkg/conversion/capi2mapi/aws.go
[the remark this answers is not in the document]
theobarberbany: Ah gotcha, makes sense!

pkg/conversion/capi2mapi/aws.go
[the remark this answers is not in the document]
theobarberbany: Ah - I meant for `APIVersrion`? Is that the same too?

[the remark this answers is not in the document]
JoelSpeed: Yeah neither of these values actually matter, they just have to be populated so that RawExtension doesn't complain

pkg/capi2mapi/machine.go
```diff
@@ -27,34 +30,74 @@ const (
...
+func fromCAPIMachineToMAPIMachine(capiMachine *capiv1.Machine) (*mapiv1.Machine, error) {
+	errs := []*field.Error{}
+
 	mapiMachine := &mapiv1.Machine{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        capiMachine.Name,
 			Namespace:   mapiNamespace,
 			Labels:      capiMachine.Labels,
 			Annotations: capiMachine.Annotations,
+			// OwnerReferences: These need to be converted so that any MachineSet owning a Machine is represented with the correct owner reference between the two APIs.
```
nrb: Does this need a jira card reference? The comment confuses me because it says a conversion is needed, but there's no logic or card reference.

pkg/mapi2capi/aws.go
```diff
@@ -27,87 +28,90 @@ import (
...
-	return awsProviderSpecAndInfra{Spec: s, Infrastructure: i}
+	machineSet     *mapiv1.MachineSet
+	infrastructure *configv1.Infrastructure
+	*awsMachineAndInfra
 }
 
 // FromAWSMachineAndInfra wraps a Machine API Machine for AWS and the OCP Infrastructure object into a mapi2capi AWSProviderSpec.
-func FromAWSMachineAndInfra(m *mapiv1.Machine, i *configv1.Infrastructure) awsMachineAndInfra {
-	return awsMachineAndInfra{Machine: m, Infrastructure: i}
+func FromAWSMachineAndInfra(m *mapiv1.Machine, i *configv1.Infrastructure) Machine {
```
nrb: Is the `Machine` return type correct?

nrb: Ah, I see - it's a `Machine` interface.

nrb:
/lgtm

Double checked card references too, they all made sense to me.

pkg/conversion/capi2mapi/machine.go
```diff
@@ -0,0 +1,147 @@
...
+func fromCAPIMachineToMAPIMachine(capiMachine *capiv1.Machine) (*mapiv1.Machine, error) {
+	errs := []*field.Error{}
+
+	mapiMachine := &mapiv1.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        capiMachine.Name,
+			Namespace:   mapiNamespace,
+			Labels:      capiMachine.Labels,
+			Annotations: capiMachine.Annotations,
+			// OwnerReferences: These need to be converted so that any MachineSet owning a Machine is represented with the correct owner reference between the two APIs.
```
nrb, line 46: OCPCLOUD-2716?

pkg/conversion/mapi2capi/machine.go
```diff
@@ -0,0 +1,163 @@
...
+func fromMAPIMachineToCAPIMachine(mapiMachine *mapiv1.Machine) (*capiv1.Machine, field.ErrorList) {
+	var errs field.ErrorList
+
+	capiMachine := &capiv1.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        mapiMachine.Name,
+			Namespace:   capiNamespace,
+			Labels:      mapiMachine.Labels,
+			Annotations: mapiMachine.Annotations,
+		},
```
nrb, line 47: Do we need the OwnerReferences here? OCPCLOUD-2716?

pkg/conversion/capi2mapi/machine.go
JoelSpeed: I have it on line 63 already, but can add here as well if you prefer

pkg/conversion/mapi2capi/machine.go
JoelSpeed: On line 78 but can add here

theobarberbany:
/lgtm 

Looks good & Makes sense!

PR #183, opened 2024-06-28
OCPBUGS-36296: fix: sort CredentialsRequest manifests after namespace

damdo:
CredentialsRequests go into the openshift-cloud-credential-operator namespace, so they can come before or after the openshift-cluster-api namespace. But because [they ask for Secrets in the openshift-cluster-api namespace](https://github.com/openshift/cluster-capi-operator/blob/3c6037f0a3e1b3b446d648029e2dea0f90761150/manifests/0000_30_cluster-api_00_credentials-request.yaml#L16), there would be less race and drama if the CredentialsRequest manifests.


wking (approved):
/lgtm
/Jira refresh

PR #185, opened 2024-07-02
OCPCLOUD-2207: Generate VSphereCluster

damdo:
Generate VSphereCluster

### DIFF
--- a/cmd/cluster-capi-operator/main.go
+++ b/cmd/cluster-capi-operator/main.go
@@ -49,6 +49,7 @@ import (
 	crwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	configv1 "github.com/openshift/api/config/v1"
+	mapiv1 "github.com/openshift/api/machine/v1beta1"
 	"github.com/openshift/cluster-capi-operator/pkg/controllers"
 	"github.com/openshift/cluster-capi-operator/pkg/controllers/capiinstaller"
 	"github.com/openshift/cluster-capi-operator/pkg/controllers/cluster"
@@ -79,6 +80,7 @@ func initScheme(scheme *runtime.Scheme) {
 	utilruntime.Must(clusterctlv1.AddToScheme(scheme))
 	utilruntime.Must(ibmpowervsv1.AddToScheme(scheme))
 	utilruntime.Must(vspherev1.AddToScheme(scheme))
+	utilruntime.Must(mapiv1.AddToScheme(scheme))
 }
 
 //nolint:funlen
@@ -155,6 +157,7 @@ func main() {
 		DefaultNamespaces: map[string]cache.Config{
 			*managedNamespace:                {},
 			secretsync.SecretSourceNamespace: {},
+			"kube-system":                    {}, // For fetching cloud credentials.
 		},
 		SyncPeriod: &syncPeriod,
 	}
--- a/pkg/controllers/infracluster/aws.go
+++ b/pkg/controllers/infracluster/aws.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package infracluster
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"strconv"
+
+	"github.com/go-logr/logr"
+	cerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	awsv1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ensureAWSCluster ensures the AWSCluster cluster object exists.
+func (r *InfraClusterController) ensureAWSCluster(ctx context.Context, log logr.Logger) (client.Object, error) {
+	target := &awsv1.AWSCluster{ObjectMeta: metav1.ObjectMeta{
+		Name:      r.Infra.Status.InfrastructureName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	// Checking whether InfraCluster object exists. If it doesn't, create it.
+
+	if err := r.Get(ctx, client.ObjectKeyFromObject(target), target); err != nil && !cerrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get InfraCluster: %w", err)
+	} else if err == nil {
+		return target, nil
+	}
+
+	log.Info(fmt.Sprintf("AWSCluster %s/%s does not exist, creating it", target.Namespace, target.Name))
+
+	apiURL, err := url.Parse(r.Infra.Status.APIServerInternalURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse apiURL: %w", err)
+	}
+
+	port, err := strconv.ParseInt(apiURL.Port(), 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse apiURL port: %w", err)
+	}
+
+	if r.Infra.Status.PlatformStatus == nil {
+		return nil, fmt.Errorf("infrastructure PlatformStatus should not be nil: %w", err)
+	}
+
+	target = &awsv1.AWSCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      r.Infra.Status.InfrastructureName,
+			Namespace: defaultCAPINamespace,
+			// The ManagedBy Annotation is set so CAPI infra providers ignore the InfraCluster object,
+			// as that's managed externally, in this case by this controller.
+			Annotations: map[string]string{
+				clusterv1.ManagedByAnnotation: managedByAnnotationValueClusterCAPIOperatorInfraClusterController,
+			},
+		},
+		Spec: awsv1.AWSClusterSpec{
+			Region: r.Infra.Status.PlatformStatus.AWS.Region,
+			ControlPlaneEndpoint: clusterv1.APIEndpoint{
+				Host: apiURL.Hostname(),
+				Port: int32(port),
+			},
+		},
+	}
+
+	if err := r.Create(ctx, target); err != nil {
+		return nil, fmt.Errorf("failed to create InfraCluster: %w", err)
+	}
+
+	log.Info(fmt.Sprintf("InfraCluster '%s/%s' successfully created", defaultCAPINamespace, r.Infra.Status.InfrastructureName))
+
+	return target, nil
+}
--- a/pkg/controllers/infracluster/infracluster_controller.go
+++ b/pkg/controllers/infracluster/infracluster_controller.go
@@ -19,18 +19,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/url"
-	"strconv"
 
 	"github.com/go-logr/logr"
 
-	cerrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"k8s.io/client-go/rest"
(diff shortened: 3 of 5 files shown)


cmd/cluster-capi-operator/main.go
```diff
@@ -155,6 +155,7 @@ func main() {
 		DefaultNamespaces: map[string]cache.Config{
 			*managedNamespace:                {},
 			secretsync.SecretSourceNamespace: {},
+			"kube-system":                    {}, // For fetching cloud credentials.
```
nrb, line 158: This will never change?

damdo: Correct, it is the location where vsphere credentials are stored in OCP

damdo:
e2e-gcp-ovn-techpreview is permafailing for unrelated reasons

pkg/controllers/infracluster/aws.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// ensureAWSCluster ensures the AWSCluster cluster object exists.
+func (r *InfraClusterController) ensureAWSCluster(ctx context.Context, log logr.Logger) (client.Object, error) {
+	target := &awsv1.AWSCluster{ObjectMeta: metav1.ObjectMeta{
+		Name:      r.Infra.Status.InfrastructureName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	// Checking whether InfraCluster object exists. If it doesn't, create it.
```
nrb, line 39: In what cases would it exist (or not)?

pkg/controllers/infracluster/aws.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if err := r.Get(ctx, client.ObjectKeyFromObject(target), target); err != nil && !cerrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get InfraCluster: %w", err)
+	} else if err == nil {
+		return target, nil
```
nrb, line 44: Should we confirm that the existing Cluster `AWSCluster` is managed by us before returning here? Is there a webhook somewhere ensuring that a user can't make one that isn't managed by this controller?

cmd/cluster-capi-operator/main.go
nrb: Is there a reason not to use a shared constant for it here and in infracluster_controller.go?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -222,63 +171,51 @@ func (r *InfraClusterController) reconcileInfraCluster(ctx context.Context, log
...
-		return nil, fmt.Errorf("failed to get InfraCluster: %w", err)
-	} else if err == nil {
-		return target, nil
-	}
-
-	log.Info(fmt.Sprintf("AWSCluster %s/%s does not exist, creating it", target.Namespace, target.Name))
+// ensureInfraCluster ensures an InfraCluster object exists in the cluster.
+func (r *InfraClusterController) ensureInfraCluster(ctx context.Context, log logr.Logger) (client.Object, error) {
+	var infraCluster client.Object
+	// TODO: implement InfraCluster generation for missing platforms.
```
nrb, line 177: Can we remove this TODO now?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -356,3 +293,37 @@ func getReadiness(infraCluster client.Object) (bool, error) {
 
 	return val, nil
 }
+
+// getRawMAPIProviderSpec returns a raw Machine ProviderSpec from the the cluster.
+func getRawMAPIProviderSpec(ctx context.Context, cl client.Client) ([]byte, error) {
+	var rawProviderSpec []byte
+	if cpms, _ := getActiveCPMS(ctx, cl); cpms != nil {
```
nrb, line 300: This appears to be the only place that we call `getActiveCPMS` and we drop the error. Should we log errors from `Get` and not return them in the function? Or is there a later use that we might have to act on the error?

pkg/controllers/infracluster/vsphere.go
```diff
@@ -0,0 +1,204 @@
...
+	return string(username), string(password), nil
+}
+
+// getVSphereServerAddr obtains the VSphere Server address.
+func (r *InfraClusterController) getVSphereServerAddr(ctx context.Context) (string, error) {
+	if r.Infra.Spec.PlatformSpec.VSphere == nil || len(r.Infra.Spec.PlatformSpec.VSphere.VCenters) == 0 {
+		// Devise VSphere server addr via MAPI providerSpec.
+		machineSpec, err := getVSphereMAPIProviderSpec(ctx, r.Client)
+		if err != nil || machineSpec == nil {
+			return "", errUnableToGetVSphereMAPIProviderSpec
```
nrb, line 193: Will this error ever be reported? It looks like this is the only possible error that could happen at L49 in this file, and that function returns a different error without wrapping.

pkg/controllers/infracluster/vsphere.go
```diff
@@ -0,0 +1,204 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	vspherev1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	errObtainingVSphereServerAddr             = errors.New("error obtaining VSphere server address")
+	errUnableToGetVSphereMAPIProviderSpec     = errors.New("unable to get VSphere MAPI ProviderSpec")
+	errUnableToFindCredsSecret                = errors.New("unable to get the VSphere credentials secret")
```
nrb, line 40: I like the idea of using known errors for these credential cases, since callers could switch on the error returned. However, for that to be useful in code, these would need to be exported.

pkg/controllers/infracluster/vsphere.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// getVSphereCredentials obtains the VSphere credentials from the well-known credentials secret.
+func (r *InfraClusterController) getVSphereCredentials(ctx context.Context, vsphereServerAddr string) (string, string, error) {
+	vSphereCredentialsSecret := &corev1.Secret{}
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Namespace: kubeSystemNamespace,
+		Name:      vSphereCredentialsName,
+	}, vSphereCredentialsSecret); err != nil {
+		return "", "", fmt.Errorf("%w %s/%s: %w", errUnableToFindCredsSecret, kubeSystemNamespace, vSphereCredentialsName, err)
```
nrb, line 171: Like I said above, I like the idea of named errors for these cases. However, afaik by wrapping them with `fmt.Errorf`, we can't do any sort of type matching against them later.

Maybe we define a small type for these since main thing that seems to differ is the error message? Something like in https://earthly.dev/blog/golang-errors/.

nrb (commented):
Overall this looks good, but had a few questions around how some of the objects might be used and how errors might be improved.

pkg/controllers/infracluster/aws.go
damdo: Yes I guess it can be created by the (CAPI based) Installer and pivoted, or it can be created via Install provided Manifests.
This does account for those cases and does not override it

pkg/controllers/infracluster/aws.go
damdo: See https://github.com/openshift/cluster-capi-operator/pull/185#discussion_r1705415762

pkg/controllers/infracluster/infracluster_controller.go
damdo: Ideally we need to also support the other platforms with an ensure (aside from Openstack, which handles the creation separately in their own controller).

pkg/controllers/infracluster/infracluster_controller.go
damdo: Changed the logic around to check both nil and err

pkg/controllers/infracluster/vsphere.go
damdo: Changed the logic around to make `machineSpec == nil` unnecessary and always wrap the err

pkg/controllers/infracluster/vsphere.go
damdo: I created these and the ones in infracluster_controller.go because they are required by the new linter we added. 
I don't necessarily thing this package will be used anywhere else other than through the manager, hence I don't think it is worth or useful exporting the errs.

pkg/controllers/infracluster/vsphere.go
damdo: Not sure it is worth doing now, as we are not using those errors anywhere other than where they are thrown from.
I think we can refactor them if we want to export and match on them in the future.

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -19,18 +19,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/url"
-	"strconv"
 
 	"github.com/go-logr/logr"
 
 	cerrors "k8s.io/apimachinery/pkg/api/errors"
```
JoelSpeed, line 25: This is traditionally imported as `kerrors`, may want to fix at some point separately

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -356,3 +295,47 @@ func getReadiness(infraCluster client.Object) (bool, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to get control plane machine set: %w", err)
+	}
+
+	if cpms == nil {
+		// The CPMS is not present or inactive.
+		// Devise VSphere providerSpec via one of the machines in the cluster.
+		machineSetList := &mapiv1beta1.MachineSetList{}
+		if err := cl.List(ctx, machineSetList, client.InNamespace(defaultMAPINamespace)); err != nil {
+			return nil, fmt.Errorf("%w: %w", errUnableToListMachineSets, err)
```
JoelSpeed, line 311: I've never seen someone wrap two errors like this, I don't think this is what we want, I'd suggest replacing the first `%w` with just a string message

pkg/controllers/infracluster/vsphere.go
```diff
@@ -0,0 +1,204 @@
...
+	errUnableToFindUsernameVSphereCredsSecret = errors.New("unable to find username in the VSphere credentials secret")
+)
+
+// ensureVSphereCluster ensures the VSphereCluster cluster object exists.
+//
+//nolint:funlen
+func (r *InfraClusterController) ensureVSphereCluster(ctx context.Context, log logr.Logger) (client.Object, error) {
+	vsphereServerAddr, err := r.getVSphereServerAddr(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", errObtainingVSphereServerAddr, err)
```
JoelSpeed, line 51: Pretty sure the linter shouldn't be requiring double wrapping like this, I'm not sure it's even supported

pkg/controllers/infracluster/vsphere.go
JoelSpeed: When wrapping them with the `%w` verb, you can type match later with existing `errors` package tooling, `errors.As` or `errors.Is`. But, I don't know what it would do with two wraps, lets skip the double wrapping

pkg/controllers/infracluster/vsphere.go
```diff
@@ -0,0 +1,204 @@
...
+	return string(username), string(password), nil
+}
+
+// getVSphereServerAddr obtains the VSphere Server address.
+func (r *InfraClusterController) getVSphereServerAddr(ctx context.Context) (string, error) {
+	if r.Infra.Spec.PlatformSpec.VSphere == nil || len(r.Infra.Spec.PlatformSpec.VSphere.VCenters) == 0 {
+		// Devise VSphere server addr via MAPI providerSpec.
+		machineSpec, err := getVSphereMAPIProviderSpec(ctx, r.Client)
+		if err != nil {
+			return "", fmt.Errorf("%w: %w", errUnableToGetVSphereMAPIProviderSpec, err)
```
JoelSpeed, line 193: Lets skip this double wrap

JoelSpeed:
I just tried this locally and removing the double wrapping is fine, it's only static errors that should be declared at the top level. When you are already wrapping another error, do the wrapping and add the additional text locally please

pkg/controllers/infracluster/infracluster_controller.go
damdo: Updated

pkg/controllers/infracluster/aws.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	log.Info(fmt.Sprintf("AWSCluster %s/%s does not exist, creating it", target.Namespace, target.Name))
```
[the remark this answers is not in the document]
damdo: I'm not a fan of the way log.Info adds the extra keys and values, that's why I use the fmt.Sprintf, but happy to change if others prefer that instead.

damdo:
PR updated and feedback addressed. cc. @JoelSpeed

pkg/controllers/infracluster/aws.go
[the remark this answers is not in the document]
JoelSpeed: The frameworks are moving towards the keys and values, and we use the same in CPMS as well, I'd like us to follow the frameworks here as it enables better machine parsing of error messages which helps our consumers

JoelSpeed (commented):
/lgtm
/approve

PR #189, opened 2024-07-19
OCPCLOUD-2206: Generate AzureCluster resource for CAPI cluster

racheljpg:
Hello! 
This is a PR for [OCPCLOUD-2205](https://issues.redhat.com/browse/OCPCLOUD-2206) to generate a GCPCluster resource for CAPI clusters. Thanks!

### DIFF
--- a/pkg/controllers/infracluster/azure.go
+++ b/pkg/controllers/infracluster/azure.go
@@ -1,0 +1,293 @@
+/*
+Copyright 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package infracluster
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+	"strconv"
+
+	"github.com/go-logr/logr"
+	mapiv1beta1 "github.com/openshift/api/machine/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	cerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ptr "k8s.io/utils/ptr"
+	azurev1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	errUnableToGetAzureMAPIProviderSpec = errors.New("unable to get Azure MAPI ProviderSpec")
+	errUnableToGetAzureClientID         = errors.New("unable to get Azure Client ID")
+	errUnableToGetAzureTenantID         = errors.New("unable to get Azure Tenant ID")
+	errPlatformStatusNil                = errors.New("platform status should not be nil")
+)
+
+const (
+	clusterSecretName               = "capz-manager-cluster-credential"    // #nosec G101
+	capzManagerBootstrapCredentials = "capz-manager-bootstrap-credentials" // #nosec G101
+)
+
+// ensureAzureCluster ensures the AzureCluster cluster object exists.
+func (r *InfraClusterController) ensureAzureCluster(ctx context.Context, log logr.Logger) (client.Object, error) {
+	// Get the Azure Bootstrap Credentials Secret.
+	// This is created by the Cluster Credential Operator and should always exist. We expect to always find it, if not, we error.
+	capzManagerBootstrapSecret := &corev1.Secret{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: capzManagerBootstrapCredentials}, capzManagerBootstrapSecret); err != nil && !cerrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get Azure Boostrap Credentials Secret: %w", err)
+	}
+
+	if err := r.ensureClusterSecret(ctx, *capzManagerBootstrapSecret); err != nil {
+		return nil, fmt.Errorf("error obtaining Azure Cluster Secret: %w", err)
+	}
+
+	if err := r.ensureClusterIdentity(ctx, *capzManagerBootstrapSecret); err != nil {
+		return nil, fmt.Errorf("error ensuring Azure Cluster Identity: %w", err)
+	}
+
+	target := &azurev1.AzureCluster{ObjectMeta: metav1.ObjectMeta{
+		Name:      r.Infra.Status.InfrastructureName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	if err := r.ensureAzureInfraCluster(ctx, target, log); err != nil {
+		return nil, fmt.Errorf("error ensuring Azure Infra Cluster: %w", err)
+	}
+
+	return target, nil
+}
+
+// getAzureMAPIProviderSpec returns a Azure Machine ProviderSpec from the the cluster.
+func getAzureMAPIProviderSpec(ctx context.Context, cl client.Client) (*mapiv1beta1.AzureMachineProviderSpec, error) {
+	rawProviderSpec, err := getRawMAPIProviderSpec(ctx, cl)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain MAPI ProviderSpec: %w", err)
+	}
+
+	providerSpec := &mapiv1beta1.AzureMachineProviderSpec{}
+	if err := yaml.Unmarshal(rawProviderSpec, providerSpec); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal MAPI ProviderSpec: %w", err)
+	}
+
+	return providerSpec, nil
+}
+
+// getAzureLocation obtains the Azure Location.
+func (r *InfraClusterController) getAzureLocation(ctx context.Context) (string, error) {
+	// Devise Azure location via MAPI providerSpec.
+	machineSpec, err := getAzureMAPIProviderSpec(ctx, r.Client)
+	if err != nil {
+		return "", fmt.Errorf("error getting azure providerSpec: %w", err)
+	}
+
+	if machineSpec == nil {
+		return "", errUnableToGetAzureMAPIProviderSpec
+	}
+
+	return machineSpec.Location, nil
+}
+
+// ensureClusterSecret ensures the AzureClusterSecret exists, and if it doesn't, creates it.
+func (r *InfraClusterController) ensureClusterSecret(ctx context.Context, capzManagerBootstrapSecret corev1.Secret) error {
+	// Get the Azure Cluster Secret.
+	// CAPZ controllers expect a Secret Ref with a different data structure to what the secret we get from the Cluster Credential Operator above provides.
+	// That is why we are creating a new secret below to copy the values to.
+	clusterSecret := &corev1.Secret{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: clusterSecretName}, clusterSecret); err != nil && !cerrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get Azure Cluster Secret: %w", err)
+	} else if err == nil {
+		// When the object already exists, there's nothing to do.
+		return nil
+	}
+
+	if err := createNewAzureSecret(ctx, r.Client, capzManagerBootstrapSecret.Data["azure_client_secret"]); err != nil {
+		return fmt.Errorf("failed to create Azure Cluster secret: %w", err)
+	}
+
+	return nil
+}
+
+// createNewAzureSecret creates a new Azure Cluster Secret.
+func createNewAzureSecret(ctx context.Context, cl client.Client, azureClientSecret []byte) error {
+	azureSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      clusterSecretName,
+			Namespace: defaultCAPINamespace,
+		},
+		Immutable: ptr.To(true),
+		Data: map[string][]byte{
+			"clientSecret": azureClientSecret,
+		},
+	}
+
+	if err := cl.Create(ctx, azureSecret); err != nil && !cerrors.IsAlreadyExists(err) {
+		return fmt.Errorf("failed to create new Azure Cluster Secret: %w", err)
+	}
+
+	return nil
+}
+
+// ensureClusterIdentity ensures the AzureClusterSecret exists, and if it doesn't, creates it.
+func (r *InfraClusterController) ensureClusterIdentity(ctx context.Context, capzManagerBootstrapSecret corev1.Secret) error {
+	azureClusterIdentity := &azurev1.AzureClusterIdentity{}
+	// Get the Azure Cluster Identity.
+	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: r.Infra.Status.InfrastructureName}, azureClusterIdentity); err != nil && !cerrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get Azure Cluster Identity: %w", err)
+	} else if err == nil {
+		// When the object already exists, there's nothing to do.
+		return nil
+	}
+
+	if err := r.createAzureClusterIdentity(ctx, capzManagerBootstrapSecret); err != nil {
+		return fmt.Errorf("failed to create new Azure Cluster Identity: %w", err)
+	}
+
+	return nil
+}
+
+// createNewAzureClusterIdenity creates a new AzureClusterIdentity.
+func (r *InfraClusterController) createAzureClusterIdentity(ctx context.Context, capzManagerBootstrapSecret corev1.Secret) error {
+	azureClientID, ok := capzManagerBootstrapSecret.Data["azure_client_id"]
+	if !ok {
+		return errUnableToGetAzureClientID
+	}
+
+	azureTenantID, ok := capzManagerBootstrapSecret.Data["azure_tenant_id"]
+	if !ok {
+		return errUnableToGetAzureTenantID
+	}
+
+	azureClusterIdentity := &azurev1.AzureClusterIdentity{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      r.Infra.Status.InfrastructureName,
+			Namespace: defaultCAPINamespace,
+			Annotations: map[string]string{
+				// The ManagedBy Annotation is set so CAPI infra providers ignore the InfraCluster object,
+				// as that's managed externally, in this case by the cluster-capi-operator's infracluster controller.
+				clusterv1.ManagedByAnnotation: managedByAnnotationValueClusterCAPIOperatorInfraClusterController,
+			},
+		},
+		Spec: azurev1.AzureClusterIdentitySpec{
+			Type:              azurev1.ServicePrincipal,
+			AllowedNamespaces: &azurev1.AllowedNamespaces{NamespaceList: []string{defaultCAPINamespace}},
+			ClientID:          string(azureClientID),
+			TenantID:          string(azureTenantID),
+			ClientSecret:      corev1.SecretReference{Name: clusterSecretName, Namespace: defaultCAPINamespace},
+		},
+	}
+
+	// The Azure Cluster Identtiy does not exist, so it needs to be created.
+	if err := r.Create(ctx, azureClusterIdentity); err != nil && !cerrors.IsAlreadyExists(err) {
+		return fmt.Errorf("failed to create Azure Cluster Identity: %w", err)
+	}
+
+	return nil
+}
+
+// ensureAzureInfraCluster ensures the InfraCluster exists, and if it doesn't, creates it.
+func (r *InfraClusterController) ensureAzureInfraCluster(ctx context.Context, target *azurev1.AzureCluster, log logr.Logger) error {
+	if r.Infra.Status.PlatformStatus == nil {
+		return errPlatformStatusNil
+	}
+
+	// Checking whether InfraCluster object exists. If it doesn't, create it.
+	if err := r.Get(ctx, client.ObjectKeyFromObject(target), target); err != nil && !cerrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get InfraCluster: %w", err)
+	} else if err == nil {
+		// When the object already exists, there's nothing to do.
+		return nil
+	}
+
+	log.Info(fmt.Sprintf("AzureCluster %s/%s does not exist, creating it", target.Namespace, target.Name))
+
+	apiURL, err := url.Parse(r.Infra.Status.APIServerInternalURL)
+	if err != nil {
+		return fmt.Errorf("failed to parse apiUrl: %w", err)
+	}
+
+	port, err := strconv.ParseInt(apiURL.Port(), 10, 32)
+	if err != nil {
+		return fmt.Errorf("failed to parse apiUrl port: %w", err)
+	}
+
+	providerSpec, err := getAzureMAPIProviderSpec(ctx, r.Client)
+	if err != nil {
+		return fmt.Errorf("error obtaining Azure Provider Spec: %w", err)
+	}
+
+	location, err := r.getAzureLocation(ctx)
+	if err != nil {
+		return fmt.Errorf("error obtaining Azure Cluster location: %w", err)
+	}
+
+	azureCluster := r.newAzureCluster(providerSpec, apiURL, port, location)
+	if err := r.Create(ctx, azureCluster); err != nil {
+		return fmt.Errorf("error creating New Azure Cluster: %w", err)
+	}
+
+	log.Info(fmt.Sprintf("InfraCluster '%s/%s' successfully created", defaultCAPINamespace, r.Infra.Status.InfrastructureName))
+
+	return nil
+}
+
+// createNewAzureCluster creates a new Azure Infra Cluster.
+func (r *InfraClusterController) newAzureCluster(providerSpec *mapiv1beta1.AzureMachineProviderSpec, apiURL *url.URL, port int64, location string) *azurev1.AzureCluster {
+	return &azurev1.AzureCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      r.Infra.Status.InfrastructureName,
+			Namespace: defaultCAPINamespace,
+			// The ManagedBy Annotation is set so CAPI infra providers ignore the InfraCluster object,
+			// as that's managed externally, in this case by this controller.
+			Annotations: map[string]string{
+				clusterv1.ManagedByAnnotation: managedByAnnotationValueClusterCAPIOperatorInfraClusterController,
+			},
+		},
+
+		Spec: azurev1.AzureClusterSpec{
+			AzureClusterClassSpec: azurev1.AzureClusterClassSpec{
+				Location:         location,
+				AzureEnvironment: "AzurePublicCloud",
+				IdentityRef: &corev1.ObjectReference{
+					Name:      r.Infra.Status.InfrastructureName,
+					Namespace: defaultCAPINamespace,
+					Kind:      "AzureClusterIdentity",
+				},
+			},
+			NetworkSpec: azurev1.NetworkSpec{
+				NodeOutboundLB: &azurev1.LoadBalancerSpec{
+					Name: r.Infra.Status.InfrastructureName,
+					BackendPool: azurev1.BackendPool{
+						Name: r.Infra.Status.InfrastructureName,
+					},
+				},
+				Vnet: azurev1.VnetSpec{
+					Name:          providerSpec.Vnet,
+					ResourceGroup: providerSpec.NetworkResourceGroup,
+				},
+			},
+			ResourceGroup: providerSpec.ResourceGroup,
+			ControlPlaneEndpoint: clusterv1.APIEndpoint{
+				Host: apiURL.Hostname(),
+				Port: int32(port),
+			},
+		},
+	}
+}
--- a/pkg/controllers/infracluster/infracluster_controller.go
+++ b/pkg/controllers/infracluster/infracluster_controller.go
@@ -26,8 +26,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 
-	azurev1 "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
-
 	"k8s.io/client-go/rest"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -102,15 +100,11 @@ func (r *InfraClusterController) Reconcile(ctx context.Context, req ctrl.Request
 
 func (r *InfraClusterController) reconcile(ctx context.Context, log logr.Logger) (ctrl.Result, error) {
 	infraCluster, err := r.ensureInfraCluster(ctx, log)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to ensure InfraCluster: %w", err)
-	}
-
-	if infraCluster == nil {
-		log.Info("Could not find or create an InfraCluster on this supported platform. Waiting for one to be manually created.")
-		// This normally means the InfraCluster generation hasn't been implemented for this supported platform yet.
-		// See the TODO above.
+	if err != nil && errors.Is(err, errPlatformNotSupported) {
+		log.Info("Could not find or create an InfraCluster on this platform as it is not yet supported.")
 		return ctrl.Result{}, nil
+	} else if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to ensure InfraCluster: %w", err)
 	}
 
 	// At this point, the InfraCluster exists.
@@ -194,12 +188,17 @@ func (r *InfraClusterController) ensureInfraCluster(ctx context.Context, log log
 			return nil, fmt.Errorf("error ensuring GCPCluster: %w", err)
 		}
 	case configv1.AzurePlatformType:
-		azureCluster := &azurev1.AzureCluster{}
-		if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: r.Infra.Status.InfrastructureName}, azureCluster); err != nil && !kerrors.IsNotFound(err) {
-			return nil, fmt.Errorf("error getting InfraCluster object: %w", err)
+		if r.Infra.Status.PlatformStatus.Azure.CloudName == configv1.AzureStackCloud {
+			log.Info("%s cloud environment for platform %s is not supported", "environment", configv1.AzureStackCloud, "platform", configv1.AzurePlatformType)
+			return nil, errPlatformNotSupported
 		}
 
-		infraCluster = azureCluster
+		var err error
+
+		infraCluster, err = r.ensureAzureCluster(ctx, log)
+		if err != nil {
+			return nil, fmt.Errorf("error getting InfraCluster object: %w", err)
+		}
 	case configv1.PowerVSPlatformType:
 		powervsCluster := &ibmpowervsv1.IBMPowerVSCluster{}
 		if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: r.Infra.Status.InfrastructureName}, powervsCluster); err != nil && !kerrors.IsNotFound(err) {
@@ -222,7 +221,7 @@ func (r *InfraClusterController) ensureInfraCluster(ctx context.Context, log log
 
 		infraCluster = openstackCluster
 	default:
-		return nil, fmt.Errorf("skipping capi controllers setup on platform %s: %w", r.Platform, errPlatformNotSupported)
+		return nil, errPlatformNotSupported
 	}
 
 	return infraCluster, nil
--- a/e2e/azure_test.go
+++ b/e2e/azure_test.go
@@ -37,7 +37,6 @@ var _ = Describe("Cluster API Azure MachineSet", Ordered, func() {
 		}
 		framework.CreateCoreCluster(cl, clusterName, "AzureCluster")
 		mapiMachineSpec = getAzureMAPIProviderSpec(cl)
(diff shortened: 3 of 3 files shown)


pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -356,3 +367,37 @@ func getReadiness(infraCluster client.Object) (bool, error) {
 
 	return val, nil
 }
+
+// getRawMAPIProviderSpec returns a raw Machine ProviderSpec from the the cluster.
+func getRawMAPIProviderSpec(ctx context.Context, cl client.Client) ([]byte, error) {
```
nrb, line 372: This came from https://github.com/openshift/machine-api-provider-gcp/pull/86?

damdo:
@racheljpg now that https://github.com/openshift/cluster-capi-operator/pull/185 merged this can now be updated to use the changes that PR introduced

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	azureTenantID, ok := capzManagerBootstrapCredentials.Data["azure_tenant_id"]
+	if !ok {
+		return nil, fmt.Errorf("failed to get azureTenantID: %w", err)
+	}
+
+	providerSpec, err := getAzureMAPIProviderSpec(ctx, r.Client)
+	if err != nil {
+		return nil, fmt.Errorf("error obtaining GCP Provider Spec: %w", err)
```
nrb, line 96:
```suggestion
		return nil, fmt.Errorf("error obtaining Azure Provider Spec: %w", err)
```

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,207 @@
...
+	if !ok {
+		return nil, fmt.Errorf("failed to get azureClientID: %w", err)
+	}
+
+	azureTenantID, ok := capzManagerBootstrapCredentials.Data["azure_tenant_id"]
+	if !ok {
+		return nil, fmt.Errorf("failed to get azureTenantID: %w", err)
+	}
+
+	providerSpec, err := getAzureMAPIProviderSpec(ctx, r.Client)
```
nrb, line 94: For debugging, we might want to output this to the logs. The providerSpec.ResourceGroup is the value we're using for the CAPI AzureCluster's ResourceGroup, and based on https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/189/pull-ci-openshift-cluster-capi-operator-main-e2e-azure-capi-techpreview/1833880963109621760/artifacts/e2e-azure-capi-techpreview/gather-extra/artifacts/pods/openshift-cluster-api_cluster-capi-operator-6ff69b6f54-d66cc_cluster-capi-operator.log that value is empty.

racheljpg: Thanks Nolan, I've done that now! I think you're right that it's empty at this point

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error obtaining Azure Provider Spec: %w", err)
+	}
+
+	log.Info("ProviderSpec.resourceGroup: %q", providerSpec.ResourceGroup)
```
nrb, line 99: So this shows up in the logs as `"ProviderSpec.resourceGroup: %q"`. That was my mistake.

The following suggestion will also work, based on the definition of `log.Info`.

```suggestion
	log.Info("Resource Group from provider spec", "ProviderSpec.resourceGroup", providerSpec.ResourceGroup)
```

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,210 @@
...
+	log.Info("ProviderSpec.networkResourceGroup: %q", providerSpec.NetworkResourceGroup)
```
nrb, line 100:
```suggestion
	log.Info("Network Resource Group from provider spec", "ProviderSpec.networkResourceGroup", providerSpec.NetworkResourceGroup)
```

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,218 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+	"strconv"
+
+	"github.com/go-logr/logr"
+	mapiv1beta1 "github.com/openshift/api/machine/v1beta1"
+	"gopkg.in/yaml.v2"
```
nrb, line 27: This may not be the cause of the problems, but use `sig.k8s.io/yaml` instead here. The marshalling behavior is different between the two libraries.

https://github.com/openshift/cluster-capi-operator/blob/main/pkg/controllers/infracluster/vsphere.go#L34

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,218 @@
...
+	rawProviderSpec, err := getRawMAPIProviderSpec(ctx, cl)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain MAPI ProviderSpec: %w", err)
+	}
+
+	providerSpec := &mapiv1beta1.AzureMachineProviderSpec{}
+	if err := yaml.Unmarshal(rawProviderSpec, providerSpec); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal MAPI ProviderSpec: %w", err)
+	}
+
```
nrb, line 201: Let's try logging `providerSpec` here to see what came out of it.

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,205 @@
...
+		Spec: azurev1.AzureClusterIdentitySpec{
+			Type:              azurev1.ServicePrincipal,
+			AllowedNamespaces: &azurev1.AllowedNamespaces{NamespaceList: []string{defaultCAPINamespace}},
+			ClientID:          string(azureClientID),
+			TenantID:          string(azureTenantID),
+			ClientSecret:      corev1.SecretReference{Name: clusterSecretName, Namespace: defaultCAPINamespace},
+		},
+	}
+
+	if err := r.Create(ctx, azureClusterIdentity); err != nil && !apierrors.IsAlreadyExists(err) {
```
JoelSpeed, line 123: Do we check if this already exists? 

Do we need to prevent this from being removed? (Like we do for InfraCluster)

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,205 @@
...
+			// as that's managed externally, in this case by this controller.
+			Annotations: map[string]string{
+				clusterv1.ManagedByAnnotation: managedByAnnotationValueClusterCAPIOperatorInfraClusterController,
+			},
+		},
+
+		Spec: azurev1.AzureClusterSpec{
+			AzureClusterClassSpec: azurev1.AzureClusterClassSpec{
+				Location:         location,
+				AzureEnvironment: "AzurePublicCloud",
```
JoelSpeed, line 141: What do we do for AzureStack? Do we have a check before we create these resources that we are on public cloud and not stack?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,205 @@
...
+					Namespace: defaultCAPINamespace,
+					Kind:      "AzureClusterIdentity",
+				},
+			},
+			NetworkSpec: azurev1.NetworkSpec{
+				NodeOutboundLB: &azurev1.LoadBalancerSpec{
+					Name: r.Infra.Status.InfrastructureName,
+					BackendPool: azurev1.BackendPool{
+						Name: r.Infra.Status.InfrastructureName,
+					},
```
JoelSpeed, line 153: Is this always true? How did we work this out?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,205 @@
...
+	if err := yaml.Unmarshal(rawProviderSpec, providerSpec); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal MAPI ProviderSpec: %w", err)
+	}
+
+	return providerSpec, nil
+}
+
+// getAzureLocation obtains the Azure Server address.
+func (r *InfraClusterController) getAzureLocation(ctx context.Context) (string, error) {
+	if r.Infra.Spec.PlatformSpec.Azure == nil {
```
JoelSpeed, line 194: I would expect the spec to be nil in most cases, do we mean status?

Is the location available from the infra status?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,205 @@
...
+		// Devise Azure location via MAPI providerSpec.
+		machineSpec, err := getAzureMAPIProviderSpec(ctx, r.Client)
+		if err != nil || machineSpec == nil {
+			return "", errUnableToGetAzureMAPIProviderSpec
+		}
+
+		return machineSpec.Location, nil
+	}
+
+	return "", nil
```
JoelSpeed, line 204: Is this the intended action when the azure spec is nil?

pkg/controllers/infracluster/azure.go
racheljpg: Maybe a question for @damdo when he's back, I'm not very knowledgeable on AzureStack. There's no checking done around AzureStack in our e2es either. I think the right behaviour is probably to fix the e2es up first, wdyt?

pkg/controllers/infracluster/azure.go
racheljpg: Looking into this!

pkg/controllers/infracluster/azure.go
racheljpg: Yes, you're right. I've reworked this function (addressing your last comment also).
Location [isn't available from the status unfortunately](https://github.com/openshift/api/blob/3e5de946111c67cb1f2b09ff7da569012d15931d/config/v1/types_infrastructure.go#L538), so this function is necessary.

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,205 @@
...
+		Name:      r.Infra.Status.InfrastructureName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	capzManagerBootstrapCredentialsKey := client.ObjectKey{Namespace: defaultCAPINamespace, Name: capzManagerBootstrapCredentials}
+	capzManagerBootstrapCredentials := &corev1.Secret{}
+
+	if err := r.Get(ctx, capzManagerBootstrapCredentialsKey, capzManagerBootstrapCredentials); err != nil {
+		return nil, fmt.Errorf("failed to create Azure Cluster Secret: %w", err)
+	}
```
JoelSpeed, line 59: This is just trying to get the credentials secret, so, if it doesn't exist, we error here and I think the error message is wrong right?

We should be able to check if it doesn't exist and handle the absence of one or both of the secret and the AzureCluster no?

pkg/controllers/infracluster/azure.go
JoelSpeed: Do we want to check if this exists when the AzureCluster exists already?

Thinking about a scenario where we have managed to create the AzureCluster, and it refers to the identity, but the identity doesn't exist. At the moment this logic won't resolve that issue

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,238 @@
...
+		return nil, fmt.Errorf("failed to get InfraCluster: %w", err)
+
+	case target.Spec.AzureEnvironment == "AzureStackCloud":
+		log.Info("AzureEnvironment not yet supported")
+		return nil, errAzureStackNotSupported
+
+	case target.GetAnnotations()[clusterv1.ManagedByAnnotation] == managedByAnnotationValueClusterCAPIOperatorInfraClusterController:
+		if err := r.Get(ctx, capzManagerBootstrapCredentialsKey, capzManagerBootstrapCredentials); err != nil {
+			return nil, fmt.Errorf("failed to get Azure Cluster Secret: %w", err)
+		} else if capzManagerBootstrapCredentialsKey.Namespace == "" || capzManagerBootstrapCredentialsKey.Name == "" {
```
JoelSpeed, line 76: What values do we expect these to be?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,238 @@
...
+			return nil, fmt.Errorf("bootstrap credentials are not expected values: %w", err)
```
JoelSpeed, line 77: It's not an error if they aren't the values we would populate, we just want to make sure that we only ensure the secret when the values are the values we would populate

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,239 @@
...
+		Name:      r.Infra.Status.InfrastructureName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	// Checking whether InfraCluster object exists. If it doesn't, create it.
+	if err := r.Get(ctx, client.ObjectKeyFromObject(target), target); err != nil && !cerrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get InfraCluster: %w", err)
+	} else if err == nil {
+		return target, nil
+	}
```
JoelSpeed, line 60: Where did we get to with ensuring the credentials exist when this is all in the default configuration?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// Get the Azure Bootstrap Credentials Secret.
+	capzManagerBootstrapSecret := &corev1.Secret{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: capzManagerBootstrapCredentials}, capzManagerBootstrapSecret); err != nil {
+		if !cerrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get Azure Boostrap Credentials Secret: %w", err)
+		}
+	}
+
+	// Get the Azure Cluster Secret.
```
JoelSpeed, line 72: What's the purpose of this secret?

pkg/controllers/infracluster/azure.go
JoelSpeed: Did we get anywhere with this?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -194,12 +192,17 @@ func (r *InfraClusterController) ensureInfraCluster(ctx context.Context, log log
 			return nil, fmt.Errorf("error ensuring GCPCluster: %w", err)
 		}
 	case configv1.AzurePlatformType:
-		azureCluster := &azurev1.AzureCluster{}
-		if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: r.Infra.Status.InfrastructureName}, azureCluster); err != nil && !kerrors.IsNotFound(err) {
-			return nil, fmt.Errorf("error getting InfraCluster object: %w", err)
+		if r.Infra.Status.PlatformStatus.Azure.CloudName == configv1.AzureStackCloud {
+			log.Info("%s cloud environment for platform %s is not supported", "environment", configv1.AzureStackCloud, "platform", configv1.AzurePlatformType)
+			return nil, nil //nolint:nilnil
```
JoelSpeed, line 197: Should probably make this a sentinel error at some point, not blocking

nrb:
```
level=error msg=RESPONSE 400: 400 Bad Request
level=error msg=ERROR CODE: PublicIPCountLimitReached
level=error msg=--------------------------------------------------------------------------------
level=error msg={
level=error msg=  "error": {
level=error msg=    "code": "PublicIPCountLimitReached",
level=error msg=    "message": "Cannot create more than 100 public IP addresses for this subscription in this region.",
level=error msg=    "details": []
level=error msg=  }
level=error msg=} 
```

Unrelated failures.

/retest

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,239 @@
...
+
+		Spec: azurev1.AzureClusterSpec{
+			AzureClusterClassSpec: azurev1.AzureClusterClassSpec{
+				Location:         location,
+				AzureEnvironment: "AzurePublicCloud",
+				IdentityRef: &corev1.ObjectReference{
+					Name:      r.Infra.Status.InfrastructureName,
+					Namespace: defaultCAPINamespace,
+					Kind:      "AzureClusterIdentity",
+				},
```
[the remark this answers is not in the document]
racheljpg: Do you mean something like [this line here](https://github.com/openshift/cluster-capi-operator/blob/f25f923f4bdda39f7ba93ef4510583ebe1f06cdf/pkg/controllers/infracluster/azure.go#L138), or something else?

pkg/controllers/infracluster/azure.go
racheljpg: We're attempting to get the `capzManagerBootstrapSecret` [from here](https://github.com/openshift/cluster-capi-operator/blob/47e02d0f326bbf65714f5b6f400995c332ea0e2b/manifests/0000_30_cluster-api_01_credentials-request.yaml#L78), and if it isn't found with the contents we expect, we then attempt to create this secret instead. I think this should answer your question above too if I'm not mistaken :)

pkg/controllers/infracluster/infracluster_controller.go
JoelSpeed: In which case, we should talk to the installer team to see if they know of any ill effect of implementing this change on a cluster from before that date. Perhaps backend pools were not in use before?

pkg/controllers/infracluster/azure.go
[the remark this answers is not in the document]
JoelSpeed: Yes, but we [exit](https://github.com/openshift/cluster-capi-operator/blob/f25f923f4bdda39f7ba93ef4510583ebe1f06cdf/pkg/controllers/infracluster/azure.go#L59) before that point at the moment, so we need some conditionals here and to fork off for "does it not exist" vs "does it exist and look like what we are expecting"

pkg/controllers/infracluster/azure.go
JoelSpeed: Hmm, that's not how I'm reading this.

If the CAPZ secret exists, we get to this point and create the cluster secret using data from it?

Where is this secret read? It seems to be referenced in the AzureClusterIdentity?

Are we copying this secret across because it has a different format requirement to what CCO produces?

damdo: If I'm not mistaken, I think Joel here is referring to why we are creating a new secret to reference within the AzureIdentity rather that referencing directly the CAPZ Manager Bootstrap Secret, which gets created by the CCO.

The reason lies in the structure of the Secret's Data. CAPZ controllers expect a [specific Data structure](https://capz.sigs.k8s.io/topics/identities#service-principal) that is different to what CCO provides.

racheljpg: Ahhh yes sorry Joel I misunderstood you. Thank you for the explanation Dam!

pkg/controllers/infracluster/azure.go
racheljpg: We're getting the values from the secret the CCO creates, so they _should_ be correct.
Do you expect these to ever have been changed/edited? Would the secret even work as expected if these values were updated?

pkg/controllers/infracluster/azure.go
JoelSpeed: Can we explain that in a comment with the code please, I fear that may get lost with time

pkg/controllers/infracluster/azure.go
JoelSpeed: My point was more that we are fetching the AzureCluster, but we then do no introspection on it, we just return.

What happens if we created the AzureCluster and later, someone accidentally removes the Azure identity type, or the secret that it refers to? We should put the default versions of those back if the AzureCluster still references them right?

damdo:
@racheljpg make sure we [delete this function call](https://github.com/racheljpg/cluster-capi-operator/blob/970609209820d76c6c8e37aa0c40ccdd317aa865/e2e/azure_test.go#L40) and the corresponding function, so that the logic you are developing for generating the infra cluster is exercised

damdo:
@rachel It looks like it works: 🎉 
```
[~] $ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/189/pull-ci-openshift-cluster-capi-operator-main-e2e-azure-capi-techpreview/1841908828967276544/artifacts/e2e-azure-capi-techpreview/gather-extra/artifacts/pods/openshift-cluster-api_cluster-capi-operator-7b77547b8b-9wjmg_cluster-capi-operator.log | grep infracluster_controller | head -n
3
I1003 19:35:26.762825       1 infracluster_controller.go:88] "Reconciling InfraCluster" logger="InfraClusterController" controller="clusteroperator" controllerGroup="config.openshift.io" controllerKind="ClusterOperator" ClusterOperator="cluster-api" namespace="" name="cluster-api" reconcileID="5e02d3d8-0352-4c4a-8ef9-7705b9f2f7c6"
I1003 19:35:27.005306       1 infracluster_controller.go:171] "InfraCluster 'openshift-cluster-api/ci-op-0ffz31z3-d9d1a-rb9fm' successfully set to Ready" logger="InfraClusterController" controller="clusteroperator" controllerGroup="config.openshift.io" controllerKind="ClusterOperator" ClusterOperator="cluster-api" namespace="" name="cluster-api" reconcileID="5e02d3d8-0352-4c4a-8ef9-7705b9f2f7c6"
I1003 19:35:27.015782       1 infracluster_controller.go:88] "Reconciling InfraCluster" logger="InfraClusterController" controller="clusteroperator" controllerGroup="config.openshift.io" controllerKind="ClusterOperator" ClusterOperator="cluster-api" namespace="" name="cluster-api" reconcileID="a6551af9-dd75-437d-a339-0cb991112530"
```

pkg/controllers/infracluster/infracluster_controller.go
racheljpg: Hey, after speaking with the installer team can confirm [Terraform was using the same value previously](https://github.com/openshift/installer/blob/release-4.16/data/data/azure/vnet/public-lb.tf#L101-L106), so this shouldn't break any clusters installed before it was introduced in the installer code :) 
Do you think we need to include this, though? I think we just want an object with the most minimum info possible to get up and running so it can be reconciled. Is it better to leave it out in that case if it's not necessary?

pkg/controllers/infracluster/azure.go
racheljpg: Done!

pkg/controllers/infracluster/infracluster_controller.go
JoelSpeed: As far as I know, this value is needed for control plane machines, so lets include it, if we know it

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	azureTenantID, ok := capzManagerBootstrapSecret.Data["azure_tenant_id"]
+	if !ok {
+		return nil, fmt.Errorf("failed to get azureTenantID: %w", err)
+	}
+
+	azureClusterIdentity := &azurev1.AzureClusterIdentity{}
+	// Get the Azure Cluster Identity.
+	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: r.Infra.Status.InfrastructureName}, azureClusterIdentity); err != nil && !cerrors.IsAlreadyExists(err) {
```
JoelSpeed, line 100: Why would you have an `AlreadyExists` error on a `Get` call?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,251 @@
...
+)
+
+const (
+	clusterSecretName               = "capz-manager-cluster-credential"    // #nosec G101
+	capzManagerBootstrapCredentials = "capz-manager-bootstrap-credentials" // #nosec G101
+)
+
+// ensureAzureCluster ensures the AzureCluster cluster object exists.
+//
+//nolint:funlen,cyclop
```
JoelSpeed, line 48: Maybe adding these is a sign that you could refactor this function to be a little simpler?

Perhaps some of the stages should be factored out to make the flow easier to understand?

```
ensureClusterSecret()
ensureClusterIdentity()
ensureAzureInfraCluster()
```

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,251 @@
...
+	target := &azurev1.AzureCluster{ObjectMeta: metav1.ObjectMeta{
+		Name:      r.Infra.Status.InfrastructureName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	// Checking whether InfraCluster object exists. If it doesn't, create it.
+	if err := r.Get(ctx, client.ObjectKeyFromObject(target), target); err != nil {
+		if !cerrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get InfraCluster: %w", err)
+		}
```
JoelSpeed, line 139: If this were factored out into a separate function, I wonder if you could make this less indented by returning early in a nice way?

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -194,12 +193,17 @@ func (r *InfraClusterController) ensureInfraCluster(ctx context.Context, log log
 			return nil, fmt.Errorf("error ensuring GCPCluster: %w", err)
 		}
 	case configv1.AzurePlatformType:
-		azureCluster := &azurev1.AzureCluster{}
-		if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: r.Infra.Status.InfrastructureName}, azureCluster); err != nil && !kerrors.IsNotFound(err) {
-			return nil, fmt.Errorf("error getting InfraCluster object: %w", err)
+		if r.Infra.Status.PlatformStatus.Azure.CloudName == configv1.AzureStackCloud {
+			log.Info("%s cloud environment for platform %s is not supported", "environment", configv1.AzureStackCloud, "platform", configv1.AzurePlatformType)
+			return nil, errUnsupportedPlatformEnvironment
```
JoelSpeed, line 198: When this is returned to a higher level function, does it get unwrapped and handled or does this go into a backoff?

Since it's private, I'm expecting at the moment it goes into backoff?

pkg/controllers/infracluster/azure.go
racheljpg: I guess the intention here was to check if it already exists _with_ the contents we expect, and error if it doesn't, since we don't want to use a ClusterIdentity with the incorrect info.
@damdo does that match up with what our intentions should be?

pkg/controllers/infracluster/azure.go
racheljpg: I think this is similar to what I asked above - I'm not sure I see how this could be pulled out on it's own when this is supposed to be the purpose of this function, right? Or am I misunderstanding?

pkg/controllers/infracluster/infracluster_controller.go
racheljpg: Not sure on this one, can take a look after I have some meetings, but if @damdo sees this first, do you know? :)

pkg/controllers/infracluster/azure.go
JoelSpeed: My point here is that you're doing a get call, not a create. You wouldn't get an already exists on a get call, ever.

pkg/controllers/infracluster/azure.go
JoelSpeed: You can always factor things into smaller chunks if they help tell the story.

If your top level function reads as four or five "do this bit" and factors into smaller functions to handle the implementation of those, that's a nice way to make the function readable.

We quite often use this pattern where we have a `Reconcile` and a `reconcile` function, where the top level does the boilerplate, and the inner reconcile does more specific business logic.

At the end of the day, the reason we don't like long functions is that they are hard to follow, so this is only about readability and improving the experience for us, 6 months down the line when we can no longer remember what's happening here

pkg/controllers/infracluster/infracluster_controller.go
JoelSpeed: If you were handling it, it would be somewhere around https://github.com/openshift/cluster-capi-operator/blob/714c1746d8a9a014a07dc0ffac0bf5bf98750692/pkg/controllers/infracluster/infracluster_controller.go#L105-L114, but I don't see anything there right now handling this sentinel error specifically.

You'll want an `errors.Is` somewhere around there and then handle this in the same way you're currently handling a `nil` infra cluster (which we should probably also move to this sentinel pattern)

pkg/controllers/infracluster/azure.go
racheljpg: I've refactored it further, unfortunately it just moved the `//nolint:funlen` down to `ensureAzureInfraCluster()`, but `ensureAzureCluster` should be much tidier now?

pkg/controllers/infracluster/azure.go
racheljpg: Yeah, I see what you mean now, I've refactored it further so let me know what you think now whenever you get the chance

pkg/controllers/infracluster/infracluster_controller.go
racheljpg: I've added handling for that error, sorry I don't get what you mean by moving the infra cluster to the same pattern, do you mean the infra cluster should be assigned a sentinel error if it's nil that we handle here?

pkg/controllers/infracluster/azure.go
racheljpg: huh, seems I broke something moving that out to it's own function so will take a look tomorrow

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -113,6 +112,11 @@ func (r *InfraClusterController) reconcile(ctx context.Context, log logr.Logger)
 		return ctrl.Result{}, nil
 	}
 
+	if errors.Is(err, errPlatformNotSupported) {
```
JoelSpeed, line 115: We will never get here, because of line 104, you'll need to make some adjustments to the logic.

Looking at the content of ensueInfraCluster, it looks like it already returns an unsupported platform error of some description, but that's not being handled here either, should they use the same error?

Whn would `infraCluster` be nil on line 108?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,278 @@
...
+		if err != nil {
+			return fmt.Errorf("error obtaining Azure Provider Spec: %w", err)
+		}
+
+		location, err := r.getAzureLocation(ctx)
+		if err != nil {
+			return fmt.Errorf("error obtaining Azure Cluster location: %w", err)
+		}
+
+		target = &azurev1.AzureCluster{
```
JoelSpeed, line 229: Separating the creation of this azure cluster object (I mean literally instantiating the object and all the fields within) into a separate function would make this cleaner

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,278 @@
...
+
+	if r.Infra.Status.PlatformStatus == nil {
+		return fmt.Errorf("infrastructure PlatformStatus should not be nil: %w", err)
+	}
+
+	// Checking whether InfraCluster object exists. If it doesn't, create it.
+	if err := r.Get(ctx, client.ObjectKeyFromObject(target), target); err != nil {
+		if !cerrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get InfraCluster: %w", err)
+		}
```
JoelSpeed, line 215: How could you change these lines to mean that the "it doesn't exist and we should create it case" is not wrapped in this large if-block?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,278 @@
...
+		if !ok {
+			return fmt.Errorf("failed to get azureClientID: %w", err)
+		}
+
+		azureTenantID, ok := capzManagerBootstrapSecret.Data["azure_tenant_id"]
+		if !ok {
+			return fmt.Errorf("failed to get azureTenantID: %w", err)
+		}
+
+		azureClusterIdentity = &azurev1.AzureClusterIdentity{
```
JoelSpeed, line 165: Could instantiate this object in a helper to reduce the complexity of this function

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// ensureClusterIdentity ensures the AzureClusterSecret exists, and if it doesn't, creates it.
+func (r *InfraClusterController) ensureClusterIdentity(ctx context.Context, capzManagerBootstrapSecret corev1.Secret) error {
+	azureClusterIdentity := &azurev1.AzureClusterIdentity{}
+	// Get the Azure Cluster Identity.
+	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: r.Infra.Status.InfrastructureName}, azureClusterIdentity); err != nil {
+		if !cerrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get Azure Cluster Identity: %w", err)
+		}
```
JoelSpeed, line 153: Could we refactor this to reduce indentation for the main body of the function?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,278 @@
...
+	capzManagerBootstrapCredentials = "capz-manager-bootstrap-credentials" // #nosec G101
+)
+
+// ensureAzureCluster ensures the AzureCluster cluster object exists.
+func (r *InfraClusterController) ensureAzureCluster(ctx context.Context, log logr.Logger) (client.Object, error) {
+	// Get the Azure Bootstrap Credentials Secret.
+	// This is created by the Cluster Credential Operator and should always exist. We expect to always find it, if not, we error.
+	capzManagerBootstrapSecret := &corev1.Secret{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: capzManagerBootstrapCredentials}, capzManagerBootstrapSecret); err != nil {
+		if !cerrors.IsNotFound(err) {
```
JoelSpeed, line 52: Could you combine the two if statements here?

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,300 @@
...
+					ResourceGroup: providerSpec.NetworkResourceGroup,
+				},
+			},
+			ResourceGroup: providerSpec.ResourceGroup,
+			ControlPlaneEndpoint: clusterv1.APIEndpoint{
+				Host: apiURL.Hostname(),
+				Port: int32(port),
+			},
+		},
+	}
```
JoelSpeed, line 291: This could be its own function just to simplify things here
```
func (r *InfraClusterController) newAzureCluster(providersSpec machinev1.AzureProviderConfig, apiUrl, port, location string) *azurev1.AzureCluster {
	return &azurev1.AzureCluster{
		ObjectMeta: metav1.ObjectMeta{
			Name:      r.Infra.Status.InfrastructureName,
			Namespace: defaultCAPINamespace,
			// The ManagedBy Annotation is set so CAPI infra providers ignore the InfraCluster object,
			// as that's managed externally, in this case by this controller.
			Annotations: map[string]string{
				clusterv1.ManagedByAnnotation: managedByAnnotationValueClusterCAPIOperatorInfraClusterController,
			},
		},

		Spec: azurev1.AzureClusterSpec{
			AzureClusterClassSpec: azurev1.AzureClusterClassSpec{
				Location:         location,
				AzureEnvironment: "AzurePublicCloud",
				IdentityRef: &corev1.ObjectReference{
					Name:      r.Infra.Status.InfrastructureName,
					Namespace: defaultCAPINamespace,
					Kind:      "AzureClusterIdentity",
				},
			},
			NetworkSpec: azurev1.NetworkSpec{
				NodeOutboundLB: &azurev1.LoadBalancerSpec{
					Name: r.Infra.Status.InfrastructureName,
					BackendPool: azurev1.BackendPool{
						Name: r.Infra.Status.InfrastructureName,
					},
				},
				Vnet: azurev1.VnetSpec{
					Name:          providerSpec.Vnet,
					ResourceGroup: providerSpec.NetworkResourceGroup,
				},
			},
			ResourceGroup: providerSpec.ResourceGroup,
			ControlPlaneEndpoint: clusterv1.APIEndpoint{
				Host: apiURL.Hostname(),
				Port: int32(port),
			},
		},
	}
}
```

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,293 @@
...
+		Namespace: defaultCAPINamespace,
+	}}
+
+	log.Info(fmt.Sprintf("AzureCluster %s/%s does not exist, creating it", target.Namespace, target.Name))
+
+	if err := r.ensureAzureInfraCluster(ctx, target); err != nil {
+		return nil, fmt.Errorf("error ensuring Azure Infra Cluster: %w", err)
+	}
+
+	log.Info(fmt.Sprintf("InfraCluster '%s/%s' successfully created", defaultCAPINamespace, r.Infra.Status.InfrastructureName))
```
JoelSpeed, line 77: This log line is in the wrong place, we should only print this if we do actually create the object, currently it will print even when the create is not achieved, perhaps move to line 247?

racheljpg: Done!

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,293 @@
...
+
+	if err := r.createNewAzureCluster(*providerSpec, *apiURL, port, location); err != nil {
+		return errCouldNotCreateCluster
+	}
+
+	return nil
+}
+
+// createNewAzureCluster creates a new Azure Infra Cluster.
+func (r *InfraClusterController) createNewAzureCluster(providerSpec mapiv1beta1.AzureMachineProviderSpec, apiURL url.URL, port int64, location string) *azurev1.AzureCluster {
```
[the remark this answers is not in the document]
racheljpg: Updated :)

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,293 @@
...
+	if err := r.ensureClusterIdentity(ctx, *capzManagerBootstrapSecret); err != nil {
+		return nil, fmt.Errorf("error ensuring Azure Cluster Identity: %w", err)
+	}
+
+	target := &azurev1.AzureCluster{ObjectMeta: metav1.ObjectMeta{
+		Name:      r.Infra.Status.InfrastructureName,
+		Namespace: defaultCAPINamespace,
+	}}
+
+	log.Info(fmt.Sprintf("AzureCluster %s/%s does not exist, creating it", target.Namespace, target.Name))
```
JoelSpeed, line 71: Oops, this one also not in the right please, think this should be line 220 right?

racheljpg: yes, sorry I missed that one too

pkg/controllers/infracluster/azure.go
```diff
@@ -0,0 +1,293 @@
...
+	return providerSpec, nil
+}
+
+// getAzureLocation obtains the Azure Location.
+func (r *InfraClusterController) getAzureLocation(ctx context.Context) (string, error) {
+	// Devise Azure location via MAPI providerSpec.
+	machineSpec, err := getAzureMAPIProviderSpec(ctx, r.Client)
+	if err != nil || machineSpec == nil {
+		return "", errUnableToGetAzureMAPIProviderSpec
+	}
```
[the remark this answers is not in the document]
racheljpg: yeah good point, updated!

PR #196, opened 2024-08-29
OCPCLOUD-2511: manifests-gen: add InfraCluster ValidatingAdmissionPolicy

damdo:
Adds creation of infracluster VAP for CAPI infra providers via manifests-generator

Tested at: https://github.com/openshift/cluster-api-provider-vsphere/pull/48


manifests-gen/customizations.go
```diff
@@ -326,3 +334,59 @@ func mergeMaps[K comparable, V any](maps ...map[K]V) map[K]V {
...
+
+// addInfraClusterProtectionPolicy adds a Validating Admission Policy and Binding for protecting
+// InfraClusters created by the cluster-capi-operator from deletion and editing.
+func addInfraClusterProtectionPolicy(objs []unstructured.Unstructured, providerName string) []unstructured.Unstructured {
+	policy := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "admissionregistration.k8s.io/v1beta1",
+			"kind":       "ValidatingAdmissionPolicy",
+			"metadata": map[string]interface{}{
+				"name": "protect-capi-" + providerName + "cluster",
```
JoelSpeed, line 346: Can we name these similarly to how I'm naming the cloud controller manager ones, so, perhaps
```
openshift-cluster-api-protect-<infra>-cluster
```
Want to make it obvious that this is a core openshift thing and which namespace it belongs to

manifests-gen/customizations.go
```diff
@@ -326,3 +334,59 @@ func mergeMaps[K comparable, V any](maps ...map[K]V) map[K]V {
...
+							"apiGroups":   []interface{}{"infrastructure.cluster.x-k8s.io"},
+							"apiVersions": []interface{}{"*"},
+							"operations":  []interface{}{"UPDATE", "DELETE"},
+							"resources":   []interface{}{providerName + "clusters"},
+						},
+					},
+				},
+				"validations": []interface{}{
+					map[string]interface{}{
+						"expression": "!(oldObject.metadata.annotations['" + clusterv1.ManagedByAnnotation + "'] == '" + managedByAnnotationValueClusterCAPIOperatorInfraClusterController + "')",
```
JoelSpeed, line 362: On delete, do we definitely have `oldObject`?

What happens if the resource does not have `annotations` at all? Please make sure to test these policies

manifests-gen/customizations.go
```diff
@@ -326,3 +334,59 @@ func mergeMaps[K comparable, V any](maps ...map[K]V) map[K]V {
...
+			},
+		},
+	}
+
+	binding := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "admissionregistration.k8s.io/v1beta1",
+			"kind":       "ValidatingAdmissionPolicyBinding",
+			"metadata": map[string]interface{}{
+				"name": "protect-capi-" + providerName + "cluster-binding",
```
JoelSpeed, line 375: Naming as above, but we don't need the `binding` suffix IMO, we already have the resource type by the time we see the name

manifests-gen/customizations.go
damdo: > On delete, do we definitely have oldObject?

We do always have `oldObject` for DELETE and UPDATE, it is only null for CREATE, which we don't have in this rule, ref: https://github.com/kubernetes/api/blob/2ed5b3feeb965633f8ec6d42a765eb06e940cafd/admissionregistration/v1alpha1/types.go#L254

> What happens if the resource does not have annotations at all? Please make sure to test these policies

Good point, I thought it would ignore it and validate to false, but it looks like it fails instead, which makes more sense.
I tested things out on the [CEL playground here](https://playcel.undistro.io/?content=H4sIAAAAAAAAA5VTTW%2BcMBD9K64P%2BZBqFliWLpZyinLorWqiHBJ6MGYI7sIY2WaVqOp%2F7wDZTQ%2BrVXvD4zdv3nsefnENHZf801Wr%2FFUPQdUqqEgh2qCCseiv2cVFiZe6G30AF72K3dZHxq56heoFalG9XTKD7FTr3Hnq4rnkZ%2FhK%2FoPd3LAjRmg1GGEHcCpYJww2Th2vLAZnuw5cya%2F5Zz5N%2BorDGMjTzmAt2eP90IKD26XhQ48skTHtYFb0YHrwQfWDpLFpnGYiLkS8fkhimaVykz%2BVfIL%2FZWFup4pk1fJ1xpBk%2F%2Bdk5lutzjKqCfUCOFGRHsmSqYCqB5pmRIfi9WeTuHYndJKnX0SLoUgPED8oTTgSgr41TRAHESRvwjjwdnQaHsH5mbzkSbxOk3yJYTSUq0632yqOM5Er3Yhss9GiKig10HUOcbGNs6Qo0Q%2Bgl6QXe986hXCH9WANhvcMW%2BsDGaJkqBadEB%2Ftey20iWrYHzI5So%2B07ReawTqiybNsPZ1NDRhMePsOzfuYZR3ugd48LJV%2FSMuD24OTbK%2BJjuaSClP1%2BowUWqMwLutBy1XTUwU3Aq1mb2ugrZz%2Bt99%2FAO7T3jB3AwAA) and I came up with a better expression rule.

manifests-gen/customizations.go
damdo: Done

manifests-gen/customizations.go
damdo: Sounds good. Done

JoelSpeed:
Changes look better, please make sure this is thoroughly tested prior to merging though

damdo:
@JoelSpeed yes, I did test the CEL rule on the [CEL playground here](https://playcel.undistro.io/?content=H4sIAAAAAAAAA5VTTW%2BcMBD9K64P%2BZBqFliWLpZyinLorWqiHBJ6MGYI7sIY2WaVqOp%2F7wDZTQ%2BrVXvD4zdv3nsefnENHZf801Wr%2FFUPQdUqqEgh2qCCseiv2cVFiZe6G30AF72K3dZHxq56heoFalG9XTKD7FTr3Hnq4rnkZ%2FhK%2FoPd3LAjRmg1GGEHcCpYJww2Th2vLAZnuw5cya%2F5Zz5N%2BorDGMjTzmAt2eP90IKD26XhQ48skTHtYFb0YHrwQfWDpLFpnGYiLkS8fkhimaVykz%2BVfIL%2FZWFup4pk1fJ1xpBk%2F%2Bdk5lutzjKqCfUCOFGRHsmSqYCqB5pmRIfi9WeTuHYndJKnX0SLoUgPED8oTTgSgr41TRAHESRvwjjwdnQaHsH5mbzkSbxOk3yJYTSUq0632yqOM5Er3Yhss9GiKig10HUOcbGNs6Qo0Q%2Bgl6QXe986hXCH9WANhvcMW%2BsDGaJkqBadEB%2Ftey20iWrYHzI5So%2B07ReawTqiybNsPZ1NDRhMePsOzfuYZR3ugd48LJV%2FSMuD24OTbK%2BJjuaSClP1%2BowUWqMwLutBy1XTUwU3Aq1mb2ugrZz%2Bt99%2FAO7T3jB3AwAA) and I did manual testing for:
- issuing deletion on an openshift-cluster-api's infracluster with the expected annotation k+v, deined deletion ✅ 
  ```
  [~] $ oc -n openshift-cluster-api get vspherecluster
  NAME                        READY   SERVER                                       AGE
  ci-ln-clr439t-c1627-r8zvc   true    vcenter-1.ci.ibmc.devcluster.openshift.com   132m
  
  [~] $ oc -n openshift-cluster-api get vspherecluster ci-ln-clr439t-c1627-r8zvc -o yaml | grep managed-by
  cluster.x-k8s.io/managed-by: cluster-capi-operator-infracluster-controller
  
  [~] $ oc -n openshift-cluster-api delete vspherecluster ci-ln-clr439t-c1627-r8zvc
  The vsphereclusters "ci-ln-clr439t-c1627-r8zvc" is invalid: : ValidatingAdmissionPolicy 'openshift-cluster-api-protect-vspherecluster' with binding 'openshift-cluster-api-protect-vspherecluster' denied request: InfraCluster resources with the 'cluster.x-k8s.io/managed-by : cluster-capi-operator-infracluster-controller' annotation cannot be edited or deleted.
  ```
- issuing deletion on an openshift-cluster-api's infracluster with annotations but not the one we want, successfully deleted ✅ 
  ```
  [~] $ oc -n openshift-cluster-version scale deploy/cluster-version-operator --replicas 0
  deployment.apps/cluster-version-operator scaled
  [~] $ oc -n openshift-cluster-api scale deploy/cluster-capi-operator --replicas 0
  deployment.apps/cluster-capi-operator scaled
  [~] $ oc -n openshift-cluster-api get validatingadmissionpolicybinding openshift-cluster-api-protect-vspherecluster -o yaml > /tmp/vb.yaml
  [~] $ oc -n openshift-cluster-api delete validatingadmissionpolicybinding openshift-cluster-api-protect-vspherecluster
  Warning: deleting cluster-scoped resources, not scoped to the provided namespace
  validatingadmissionpolicybinding.admissionregistration.k8s.io "openshift-cluster-api-protect-vspherecluster" deleted
  [~] $ oc -n openshift-cluster-api edit vspherecluster ci-ln-clr439t-c1627-r8zvc # change managed-by annotation to value: a
  vspherecluster.infrastructure.cluster.x-k8s.io/ci-ln-clr439t-c1627-r8zvc edited
  [~] $ oc -n openshift-cluster-api apply -f /tmp/vb.yaml # restore the vapb
  [~] $ oc -n openshift-cluster-api delete vspherecluster ci-ln-clr439t-c1627-r8zvc
  vspherecluster.infrastructure.cluster.x-k8s.io "ci-ln-clr439t-c1627-r8zvc" deleted
  ```
- issuing deletion on an  openshift-cluster-api's infracluster without annotations, successfully deleted ✅ 
  ```
  [~] $ oc -n openshift-cluster-version scale deploy/cluster-version-operator --replicas 0
  deployment.apps/cluster-version-operator scaled
  [~] $ oc -n openshift-cluster-api scale deploy/cluster-capi-operator --replicas 0
  deployment.apps/cluster-capi-operator scaled
  [~] $ oc -n openshift-cluster-api get validatingadmissionpolicybinding openshift-cluster-api-protect-vspherecluster -o yaml > /tmp/vb.yaml
  [~] $ oc -n openshift-cluster-api delete validatingadmissionpolicybinding openshift-cluster-api-protect-vspherecluster
  Warning: deleting cluster-scoped resources, not scoped to the provided namespace
  validatingadmissionpolicybinding.admissionregistration.k8s.io "openshift-cluster-api-protect-vspherecluster" deleted
  [~] $ oc -n openshift-cluster-api edit vspherecluster ci-ln-clr439t-c1627-r8zvc # completely remove annotations
  vspherecluster.infrastructure.cluster.x-k8s.io/ci-ln-clr439t-c1627-r8zvc edited
  [~] $ oc -n openshift-cluster-api apply -f /tmp/vb.yaml # restore the vapb
  [~] $ oc -n openshift-cluster-api delete vspherecluster ci-ln-clr439t-c1627-r8zvc
  vspherecluster.infrastructure.cluster.x-k8s.io "ci-ln-clr439t-c1627-r8zvc" deleted
  ```

Let me know if that sounds sensible. Thanks.

manifests-gen/customizations.go
```diff
@@ -326,3 +334,59 @@ func mergeMaps[K comparable, V any](maps ...map[K]V) map[K]V {
...
+							"apiVersions": []interface{}{"*"},
+							"operations":  []interface{}{"UPDATE", "DELETE"},
+							"resources":   []interface{}{providerName + "clusters"},
+						},
+					},
+				},
+				"validations": []interface{}{
+					map[string]interface{}{
+						"expression": "!(has(oldObject.metadata.annotations) && ('cluster.x-k8s.io/managed-by' in oldObject.metadata.annotations) && (oldObject.metadata.annotations['" + clusterv1.ManagedByAnnotation + "'] == '" + managedByAnnotationValueClusterCAPIOperatorInfraClusterController + "'))",
+						"message":    "InfraCluster resources with the '" + clusterv1.ManagedByAnnotation + " : " + managedByAnnotationValueClusterCAPIOperatorInfraClusterController + "' annotation cannot be edited or deleted.",
```
JoelSpeed, line 363: Are we sure we don't want this to be updated? I don't remember that being a requirement.

My understanding of the task here is to prevent the infra cluster from being deleted, when it is the cluster generated one.

So, why don't we check instead for the name of the object, and just apply a deny straight up if the cluster object is the name of the cluster infrastructure ID?

damdo: > My understanding of the task here is to prevent the infra cluster from being deleted, when it is the cluster generated one.

Yes the card talks about preventing deletions. I originally included updates there because the InfraCluster objects are not that important for CAPI when externally managed and most of the changes to those objects are ignored by CAPI itself, hence the editing prevention.

That said if you think we want to leave this open because after all, this is only a generated one that might need correcting by the user, then by all means we can only prevent deletion.

> So, why don't we check instead for the name of the object, and just apply a deny straight up if the cluster object is the name of the cluster infrastructure ID?

Yes we can deny deletion only if we prefer so.
Though I'd prefer not to use the name of the object as a filter for the VAP, but keep filtering on the CAPI standardised externally managed-by annotation k/v:

a) there are situations in the future where we might end up having multiple of these InfaCluster, even for the same provider (e.g. VSphereClusters for multi Vcenter support, & such). The name wouldn't necessarily match the Cluster Infra ID.

b) the user can provide their own InfraClusters via install time manifests. Those, I think, we should allow deleting as they aren't generated by our operator.

c) there might be InfraClusters with the name that match Cluster Infra ID, like for Openstack, which we instead want to run "internally-managed" by its infracontroller. Those we don't want to cover by these rules.

JoelSpeed: >That said if you think we want to leave this open because after all, this is only a generated one that might need correcting by the user, then by all means we can only prevent deletion.

I'm thinking that we might have either manual or auto fixups in the future if we work out a way to improve this.

> a) there are situations in the future where we might end up having multiple of these InfaCluster, even for the same provider (e.g. VSphereClusters for multi Vcenter support, & such). The name wouldn't necessarily match the Cluster Infra ID.

It wouldn't, but we could control the name of these right?

> b) the user can provide their own InfraClusters via install time manifests. Those, I think, we should allow deleting as they aren't generated by our operator.

So thinking about the future were we move the infra object from install time to the cluster, are we expecting that to be considered externally managed? I wouldn't want the user to be able to delete the one pivoted by the installer either

> c) there might be InfraClusters with the name that match Cluster Infra ID, like for Openstack, which we instead want to run "internally-managed" by its infracontroller. Those we don't want to cover by these rules.

I don't follow. If it's the `InfraCluster` in the `openshift-cluster-api` namespace (which only allows one `Cluster` right?), then we don't want it to be deleted, no matter what's created it, or who's managing it. Deleting this particular `InfraCluster` would have negative effects on the cluster we are running in?

damdo: > I'm thinking that we might have either manual or auto fixups in the future if we work out a way to improve this.
Cool, happy to drop update protection then 👍 

> It wouldn't, but we could control the name of these right?

We could, if we were interested in protecting all types of InfraCluster which happen to have that name. Whereas if the desire is to only protect InfraClusters that we generate, I think using the CAPI standardized annotation is more straightforward and I don't see many downsides as that the CAPI provider contract settled on for externally-managed InfraClusters.

> So thinking about the future were we move the infra object from install time to the cluster, are we expecting that to be considered externally managed? I wouldn't want the user to be able to delete the one pivoted by the installer either

It probably won't be considered externally managed if pivoted, but that IMO should boil down to whether from then on (after pivoting) we will keep reconciling the infra together with the machine management, or we stop doing that and only do the latter. We keep reconciling it? internally managed; otherwise: externally-managed.
The pivoting design for OCP is still unclear IIRC, but my impression is that those CR should come with their own rules, and not from the ones coming from the transport configmaps we use for day 2 CAPI installations.

> I don't follow. If it's the InfraCluster in the openshift-cluster-api namespace (which only allows one Cluster right?), then we don't want it to be deleted, no matter what's created it, or who's managing it. Deleting this particular InfraCluster would have negative effects on the cluster we are running in?

So If everything boils down to whether we want to cover only the InfraCluster that we generate, or also other InfraClusters that may have their name after the cluster ID (which may have been generated by other actors).
If the intention is the latter, and we don't see downsides of blocking other actors from then also being able to delete their own InfraCluster, then yeah.

But how would we be able to reference the name of another resource, like the cluster infra ID, in a VAP rule?
I suspect that won't be possible?

At this point we could even consider disallowing deletion of all InfraClusters in the namespace we own!?

JoelSpeed: > We could, if we were interested in protecting all types of InfraCluster which happen to have that name. Whereas if the desire is to only protect InfraClusters that we generate, I think using the CAPI standardized annotation is more straightforward and I don't see many downsides as that the CAPI provider contract settled on for externally-managed InfraClusters.

I want to prevent deletion of any InfraCluster that would then render the cluster itself in a bad way, be that created by us, or the installer. Basically, if it's referenced by the `Cluster`, I think it should be prevented from deletion. How that will work for the multi vsphere case I'm not sure.

> The pivoting design for OCP is still unclear IIRC, but my impression is that those CR should come with their own rules, and not from the ones coming from the transport configmaps we use for day 2 CAPI installations.

We want to pivot the resource into the cluster to keep it around, but we don't want to reconcile it, apart from to capture any drift from the existing state.

> But how would we be able to reference the name of another resource, like the cluster infra ID, in a VAP rule?
I suspect that won't be possible?

You can use params to bring in fields from other resources, so we could get the cluster ID from the `Infrastructure` object

> At this point we could even consider disallowing deletion of all InfraClusters in the namespace we own!?

That would be safe, but overly safe, and no doubt at some point someone will raise a ticket because of this

damdo: > I want to prevent deletion of any InfraCluster that would then render the cluster itself in a bad way, be that created by us, or the installer. Basically, if it's referenced by the Cluster, I think it should be prevented from deletion. How that will work for the multi vsphere case I'm not sure.

Sounds good, we can add a modification to the validating rule for also protecting `infrastructureName + suffix` Objects when/if we cross the multi VSphereCluster bridge.

> You can use params to bring in fields from other resources, so we could get the cluster ID from the Infrastructure object

Trialing with that as we chat

> That would be safe, but overly safe, and no doubt at some point someone will raise a ticket because of this

Yeah most probably

JoelSpeed (commented):
Changes look good aside from the missing field in the paramRef that we discussed earlier

damdo:
@JoelSpeed Thanks. Required `parameterNotFoundAction` added now. LMK

damdo:
This change doesn't affect payload, it is only a tooling generation hence overriding e2e

/override ci/prow/e2e-aws-ovn-serial ci/prow/e2e-azure-capi-techpreview

PR #199, opened 2024-09-03
fix: infracluster: do not attempt vsphere secret recreation

damdo:
Do not attempt vsphere secret recreation when it already exists.
Otherwise this yields `resourceVersion should not be set on objects to be created` errors.


JoelSpeed (commented):
/approve
/lgtm

damdo:
ci/prow/e2e-azure-capi-techpreview is broken due to the known DNS issues, but it looks ok for what concerns this change.

PR #202, opened 2024-09-05
OCPCLOUD-2643: Bootstrap new manager for migration and sync controllers

JoelSpeed:
This is the first step towards [OCPCLOUD-2643](https://issues.redhat.com//browse/OCPCLOUD-2643) and I've picked it up because it allows forking off into two streams working on migration and sync controllers separately.

This adds a new manager that will eventually register the migration and various sync controllers we have yet to build.
It currently waits to see if the migration feature gate is enabled, and, if it is not, waits to be terminated.

When the gate is enabled, it will start the manager, however, the manager currently has no runnables so is effectively a no-op.

I've added the manager as a separate container within the existing deployment as previously discussed.


JoelSpeed:
E2E is showing this is doing what I expect it to do
/approve

cmd/machine-api-migration/main.go
```diff
@@ -0,0 +1,208 @@
...
+	missingVersion := "0.0.1-snapshot"
+
+	configClient, err := configv1client.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return nil, fmt.Errorf("failed to create config client: %w", err)
+	}
+
+	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+
+	// By default, this will exit(0) if the featuregates change
```
damdo, line 189:
```suggestion
	// By default, this will exit(0) if the featuregates change.
```

cmd/machine-api-migration/main.go
```diff
@@ -0,0 +1,208 @@
...
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		desiredVersion, missingVersion,
+		configInformers.Config().V1().ClusterVersions(),
+		configInformers.Config().V1().FeatureGates(),
+		events.NewLoggingEventRecorder("controlplanemachineset"),
```
damdo, line 194: This should be say something about machineapimigration or something right?

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/cluster-capi-operator
 
-go 1.21
+go 1.22.0
+
+toolchain go1.23.0
```
damdo, line 5: Do we actually want this to be 1.23?

damdo (commented):
Some nits but other than that it LGTM

nrb:
One comment on the go.mod needs to be addressed, but otherwise this looks good

go.mod
theobarberbany: IIRC running `make vendor` (I think just `go mod vendor`) will automatically add the toolchain directive now

PR #203, opened 2024-09-06
OCPBUGS-38006: Use CAPO v0.10 and API v1beta1

mdbooth:
This fixes a crashloop when deploying on OSP because the version of CAPO we deploy no longer serves v1alpha7.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-capi-operator
 
-go 1.21
+go 1.22
```
mdbooth, line 3: We have a Go builder for 4.17, so I don't think it's worth jumping through hoops to try to avoid this bump. CAPO v0.10 uses go 1.22.

mandre (commented):
Last CI run shows that [cluster-capi-operator is now stable](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/203/pull-ci-openshift-cluster-capi-operator-main-e2e-openstack-capi-techpreview/1832072079818625024/artifacts/e2e-openstack-capi-techpreview/gather-extra/artifacts/pods/openshift-cluster-api_cluster-capi-operator-59655d9f88-v7v9f_cluster-capi-operator.log).

/lgtm

mandre:
The e2e-aws-ovn-techpreview job failed but appears to be stable otherwise. Let's try again.
/retest

PR #205, opened 2024-09-10
OCPBUGS-41576: Use CAPO v0.10 and API v1beta1

mandre:
Manual backport of https://github.com/openshift/cluster-capi-operator/pull/203.

This fixes a crashloop when deploying on OSP because the version of CAPO we deploy no longer serves v1alpha7.


mandre:
@sdodson, can you help with the needed `staff-eng-approved` label?

sdodson:
@mandre Doesn't look like this is being backported to previous releases and it's not declared as a release blocker so we'll just wait until 4.17 GA is finalized and the label requirement is removed.

mandre:
> @mandre Doesn't look like this is being backported to previous releases and it's not declared as a release blocker so we'll just wait until 4.17 GA is finalized and the label requirement is removed. If this doesn't merge automatically by Monday run `/tide refresh`

Indeed, this should not go further than 4.17. It has already merged in 4.18 and showed that the operator no longer enters a crashloop, but as the feature is currently still TP, it might indeed not be a release blocker. I discussed this with @mdbooth in the morning and we think it's fine to wait for the first z-stream.

damdo:
The `security` presubmit is not even for release-4.17, not sure why it is asking for it here. See https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-capi-operator/openshift-cluster-capi-operator-release-4.17.yaml

damdo:
@mandre either we override the security job or we try close and reopen the PR (this will re-trigger all the presubmits) to see if we break out of this stall.

mandre:
If you have the power to override the security job, that's probably best.

PR #206, opened 2024-09-16
OCPCLOUD-2643: Bootstrap Machine Sync controller

theobarberbany:
-> Bootstraps Machine Sync controller
-> Adds basic test suite & empty test.


pkg/controllers/machinesync/suite_test.go
```diff
@@ -0,0 +1,110 @@
...
+var _ = BeforeSuite(func() {
+	klog.SetOutput(GinkgoWriter)
+
+	logf.SetLogger(textlogger.NewLogger(textlogger.NewConfig()))
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{
+			filepath.Join("..", "..", "..", "vendor", "github.com", "openshift", "api", "machine", "v1", "zz_generated.crd-manifests"),
+			filepath.Join("..", "..", "..", "vendor", "github.com", "openshift", "api", "machine", "v1beta1", "zz_generated.crd-manifests"),
```
theobarberbany, line 70: TODO: specify custom no upgrade

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	stopManager := func() {
+		mgrCancel()
+		// Wait for the mgrDone to be closed, which will happen once the mgr has stopped
+		<-mgrDone
+	}
+
+	BeforeEach(func() {
+		By("Setting up a namespace for the test")
+		namespace = corev1resourcebuilder.Namespace().WithGenerateName("control-plane-machine-set-controller-").Build()
```
JoelSpeed, line 69: control plane machine set?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -0,0 +1,117 @@
...
+		Expect(k8sClient.Create(ctx, namespace)).To(Succeed())
+		namespaceName = namespace.GetName()
```
JoelSpeed, line 71: If we are using custom namespaces (Which we probably do want to do, will need two, a mapi and capi one), we should make the namespaces configurable as part of the reconciler

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	JustAfterEach(func() {
+		By("Stopping the manager")
+		stopManager()
+	})
+
+	It("should reconcile without erroring", func() {
+		machine = machineBuilder.Build()
+
+		_, err := reconciler.Reconcile(ctx, reconcile.Request{
```
JoelSpeed, line 109: Calling reconcile directly is odd, but, fine for now I guess, I would imagine this test will be replaced with more specific ones when we work out what we need?

tools/tools.go
```diff
@@ -0,0 +1,13 @@
...
+// +build tools
+
+// Official workaround to track tool dependencies with go modules:
+// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+
+package tools
+
+import (
+	_ "github.com/openshift/api/config/v1/zz_generated.crd-manifests"
+	_ "github.com/openshift/api/machine/v1/zz_generated.crd-manifests"
```
JoelSpeed, line 11: Do we need v1 as well?

theobarberbany: 🤦 I forgot to remove this, we don't :)

pkg/controllers/machinesync/machine_sync_controller_test.go
theobarberbany: 👀 Copypasta - will update!

pkg/controllers/machinesync/machine_sync_controller_test.go
theobarberbany: I think I'll throw this in https://github.com/openshift/cluster-capi-operator/pull/207 as that's where we start to use them?

PR #207, opened 2024-09-18
OCPCLOUD-2643: Machinesync fetch resources

theobarberbany:
Don't merge before the bootstrapping PR: 

https://github.com/openshift/cluster-capi-operator/pull/206

/hold


cmd/machine-api-migration/main.go
```diff
@@ -153,6 +159,35 @@ func main() {
...
+	if err != nil {
+		klog.Errorf("failed to create config client: %s", err)
+		os.Exit(1)
+	}
+
+	infra, err := configClient.ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("failed to fetch infrastructure: %s", err)
+		os.Exit(1)
+	}
```
JoelSpeed, line 172: You could use `mgr.GetClient()` and use the controller runtime client rather than constructing a typed client if you prefer

cmd/machine-api-migration/main.go
```diff
@@ -206,3 +241,14 @@ func getFeatureGates(mgr ctrl.Manager) (featuregates.FeatureGateAccess, error) {
 
 	return featureGateAccessor, nil
 }
+
+// getProviderFromInfrastructure returns the PlatformType from the Infrastructure object.
+func getProviderFromInfrastructure(infra *configv1.Infrastructure) (configv1.PlatformType, error) {
+	if infra.Status.PlatformStatus != nil {
+		if infra.Status.PlatformStatus.Type != "" {
```
JoelSpeed, line 248: Could inline these?
```suggestion
	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.Type != "" {
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -18,25 +18,68 @@ package machinesync
...
 }
 
 // SetupWithManager sets the CoreClusterReconciler controller up with the given manager.
 func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	infrastructureMachine, err := getInfrastructureMachineFromProvider(r.platform)
+	if err != nil {
+		return fmt.Errorf("failed to get InfrastructureMachine from Provider: %w", err)
+	}
+
+	if len(r.capiNamespace) == 0 {
```
JoelSpeed, line 64:
```suggestion
	// Allow the namespaces to be set externally for test purposes, when not set, default to the production namespaces.
	if len(r.capiNamespace) == 0 {
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -18,25 +18,68 @@ package machinesync
...
+	errPlatformNotSupported = errors.New(`error determining InfrastructureMachine
+	type, platform not supported`)
+)
+
 // MachineSyncReconciler reconciles CAPI and MAPI machines.
 type MachineSyncReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
+
+	platform      configv1.PlatformType
```
JoelSpeed, line 52: How does this get initialised?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -18,25 +18,68 @@ package machinesync
...
+	capiNamespace string
+	mapiNamespace string
 }
 
 // SetupWithManager sets the CoreClusterReconciler controller up with the given manager.
 func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
JoelSpeed, line 58: Is this called anywhere presently?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -18,25 +18,68 @@ package machinesync
...
+
+	if len(r.capiNamespace) == 0 {
+		r.capiNamespace = capiNamespace
+	}
+
+	if len(r.mapiNamespace) == 0 {
+		r.mapiNamespace = mapiNamespace
+	}
+
+	// I'm unsure of how to scope this to just the CAPI namespace
```
JoelSpeed, line 72: Add a predicate to both `Watches` that restricts to events from the `r.capiNamespace`?

And the caches can be restricted to the correct namespaces too in the manager options in the main.go

pkg/util/watch_filters.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// CapiMachineToMapiMachine maps a Capi Machine to a MAPI machine in the
+// namespace provided.
+func CapiMachineToMapiMachine(namespace string) func(context.Context, client.Object) []reconcile.Request {
```
JoelSpeed, line 30: Nit
```suggestion
func CAPIMachineToMAPIMachine(namespace string) func(context.Context, client.Object) []reconcile.Request {
```

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Ah yoinks - it should be an exported field. Could also add a helper func / opts func, but don't know if we need that yet..

```suggestion
	Platform      configv1.PlatformType
```

pkg/util/watch_filters.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// CAPIMachineToMapiMachine maps a Capi Machine to a MAPI machine in the
+// namespace provided.
+func CAPIMachineToMapiMachine(namespace string) func(context.Context, client.Object) []reconcile.Request {
```
damdo, line 35:
```suggestion
// CAPIMachineToMAPIMachine maps a CAPI Machine to a MAPI machine in the
// namespace provided.
func CAPIMachineToMAPIMachine(namespace string) func(context.Context, client.Object) []reconcile.Request {
```

cmd/machine-api-migration/main.go
```diff
@@ -153,6 +166,32 @@ func main() {
...
 	}
 
+	k8sclient := mgr.GetClient()
+	infra := &configv1.Infrastructure{}
+
+	err = k8sclient.Get(stop, client.ObjectKey{Name: controllers.InfrastructureResourceName}, infra)
+	if err != nil {
+		klog.Errorf("failed to fetch infrastructure: %s", err)
+		os.Exit(1)
+	}
```
JoelSpeed, line 176: Nit, inline this and scope the error variable to just the if statement
```suggestion
	if err := k8sclient.Get(stop, client.ObjectKey{Name: controllers.InfrastructureResourceName}, infra); err != nil {
		klog.Errorf("failed to fetch infrastructure: %s", err)
		os.Exit(1)
	}
```

damdo, line 176: We normally use inline if error handling if the func just returns an error, to minimise the err scope.

cmd/machine-api-migration/main.go
```diff
@@ -153,6 +166,32 @@ func main() {
...
+
+	// Currently we only plan to support AWS, so all others are a noop until they're implemented.
+	switch provider {
+	case configv1.AWSPlatformType:
+		klog.Info("MachineAPIMigration: starting AWS controllers")
+
```
JoelSpeed, line 188: I think the linter is complaining about this blank line

cmd/machine-api-migration/main.go
```diff
@@ -165,8 +204,19 @@ func main() {
 		os.Exit(1)
 	}
 
-	klog.Info("starting manager")
+	reconciler := machinesync.MachineSyncReconciler{
+		Platform: provider,
+
+		MAPINamespace: *mapiManagedNamespace,
+		CAPINamespace: *managedNamespace,
```
JoelSpeed, line 211: I wonder if the capi managed namespace should have capi in the name at the flag level to match?

cmd/machine-api-migration/main.go
```diff
@@ -165,8 +204,19 @@ func main() {
...
+	}
+
+	if err := reconciler.SetupWithManager(mgr); err != nil {
+		klog.Error(err, "failed to set up reconciler with manager")
+		os.Exit(1)
+	}
 
+	klog.Info("starting manager")
```
JoelSpeed, line 219:
```suggestion
	klog.Info("Starting manager")
```

cmd/machine-api-migration/main.go
```diff
@@ -80,6 +87,11 @@ func main() {
 		controllers.DefaultManagedNamespace,
 		"The namespace where CAPI components will run.",
 	)
+	mapiManagedNamespace := flag.String(
+		"mapi-namespace",
+		"openshift-machine-api",
```
damdo, line 92: I think it'd be better to use a const for this

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -18,33 +18,101 @@ package machinesync
...
+const (
+	capiNamespace string = "openshift-cluster-api"
+	mapiNamespace string = "openshift-machine-api"
+)
+
+var (
+	// errPlatformNotSupported is returned when the platform is not supported.
+	errPlatformNotSupported = errors.New(`error determining InfrastructureMachine
+	type, platform not supported`)
+)
```
JoelSpeed, line 48: Is there a reason to use backticks vs double quotes?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -18,33 +18,101 @@ package machinesync
...
+
+	// Allow the namespaces to be set externally for test purposes, when not set,
+	// default to the production namespaces.
+	if len(r.CAPINamespace) == 0 {
+		r.CAPINamespace = capiNamespace
+	}
+
+	if len(r.MAPINamespace) == 0 {
+		r.MAPINamespace = mapiNamespace
+	}
```
damdo, line 76: Any particular reason for using `len() == 0` instead of ` == ""` for these strings?

pkg/util/watch_filters.go
```diff
@@ -0,0 +1,93 @@
...
+func FilterAWSMachines(namespace string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		machine, ok := obj.(*awscapiv1beta1.AWSMachine)
+		if !ok {
+			panic(fmt.Sprintf("expected to get an of object of type awscapiv1beta1.Machine: got type %T", obj))
+		}
+
+		// Check namespace
+		return machine.GetNamespace() == namespace
+
```
JoelSpeed, line 91: Nit: Shouldn't have a new line that's blank after the return
```suggestion
```

pkg/util/watch_filters.go
```diff
@@ -0,0 +1,93 @@
...
+func FilterCAPIMachines(namespace string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		machine, ok := obj.(*capiv1beta1.Machine)
+		if !ok {
+			panic(fmt.Sprintf("expected to get an of object of type capiv1beta1.Machine: got type %T", obj))
+		}
+
+		// Check namespace
+		return machine.GetNamespace() == namespace
+
```
JoelSpeed, line 76:
```suggestion
```

pkg/util/watch_filters.go
```diff
@@ -0,0 +1,93 @@
...
+func FilterMAPIMachines(namespace string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		machine, ok := obj.(*machinev1beta1.Machine)
+		if !ok {
+			panic(fmt.Sprintf("expected to get an of object of type machinev1beta1.Machine: got type %T", obj))
+		}
+
+		// Check namespace
+		return machine.GetNamespace() == namespace
+
```
JoelSpeed, line 61:
```suggestion
```

cmd/machine-api-migration/main.go
theobarberbany: 🤦 I forgot to push the linter fixes before re=requesting a review - my bad!

pkg/util/watch_filters.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+// FilterMAPIMachines filters MAPI Machine requests, ensuring they are in the
+// correct namespace.
+func FilterMAPIMachines(namespace string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		machine, ok := obj.(*machinev1beta1.Machine)
+		if !ok {
+			panic(fmt.Sprintf("expected to get an of object of type machinev1beta1.Machine: got type %T", obj))
```
damdo, line 56: I think there is a typo, same also with the other panics below

pkg/util/watch_filters.go
```diff
@@ -0,0 +1,93 @@
...
+
+		// Check namespace
+		return machine.GetNamespace() == namespace
+
+	})
+}
+
+// FilterAWSMachines filters AWS Machine requests, ensuring they are in the
+// correct namespace.
+func FilterAWSMachines(namespace string) predicate.Predicate {
```
damdo, line 82: Would it be useful to call this FilterInfraMachines and pass in the type (e.g. awscapiv1beta1.AWSMachine) at Predicate building invocation?

pkg/controllers/machinesync/machine_sync_controller.go
damdo, line 48: It looks like the CAPI contract and types calls these infrastructure specific machines: `InfraMachine`(s) (https://github.com/search?q=repo%3Akubernetes-sigs%2Fcluster-api%20InfraMachine&type=code)

I'd probably rename the InfrastructureMachine and Infrastructure Machine occurrences to InfraMachine :)

damdo (commented):
@theobarberbany I did an early pass while I was catching up on things, mainly stylistic nits :)

pkg/util/watch_filters.go
theobarberbany: Yeah, I wasn't 100% happy with this - and I like the idea! I wonder if the go generics could work here too? e.g similar to the new change for watches in controller runtime...

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Nope, happy to change to double quotes.

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Oh.. I thought I'd learnt somewhere along the way that comparing the length was better, although googling now it seems that may be wrong.. will change for  `== ""`!

JoelSpeed: I've also heard that length is better but have no strong opinion, I normally go for compare to empty string personally

cmd/machine-api-migration/main.go
theobarberbany: Hmm - because there's no reasonable option other than`openshift-machine-api`?

damdo: Ah sorry, I meant for storing the default value 😬 in case you need to reuse this string somewhere else where you don't have access to the flags. But we can refactor later if needed.

theobarberbany: Ohh - gotcha 🤣 That's sensible, will update :)

JoelSpeed (commented):
/lgtm
/approve

PR #208, opened 2024-09-20
OCPCLOUD-2750: Add fuzz conversion testing for MAPI2CAPI conversions

JoelSpeed:
This PR adds fuzz conversion testing between MAPI machines to CAPI machines and back again, the idea is that, bar some exceptions where we cannot currently convert, we can round trip the machines and produce the same result at the other end.

Where there are things that we cannot convert, they will be tagged with the appropriate stories to fix the feature gaps and then updated once the gaps are added.

The intention here is to create a pattern for fuzz testing and libraries so that this is easy for other platforms to implement similar.

### DIFF
--- a/pkg/conversion/mapi2capi/aws.go
+++ b/pkg/conversion/mapi2capi/aws.go
@@ -209,6 +209,11 @@ func (m *awsMachineAndInfra) toAWSMachine(providerSpec mapiv1.AWSMachineProvider
 		errs = append(errs, err)
 	}
 
+	instanceMetadataOptions, err := convertMetadataServiceOptionstoCAPI(fldPath.Child("metadataServiceOptions"), providerSpec.MetadataServiceOptions)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
 	spec := capav1.AWSMachineSpec{
 		AMI:                      capiAWSAMIReference,
 		AdditionalSecurityGroups: convertAWSSecurityGroupstoCAPI(providerSpec.SecurityGroups),
@@ -225,7 +230,7 @@ func (m *awsMachineAndInfra) toAWSMachine(providerSpec mapiv1.AWSMachineProvider
 		// ImageLookupOrg. Not used in OpenShift.
 		// NetworkInterfaces. Not used in OpenShift.
 
-		InstanceMetadataOptions: convertMetadataServiceOptionstoCAPI(providerSpec.MetadataServiceOptions),
+		InstanceMetadataOptions: instanceMetadataOptions,
 		InstanceType:            providerSpec.InstanceType,
 		NonRootVolumes:          nonRootVolumes,
 		PlacementGroupName:      providerSpec.PlacementGroupName,
@@ -343,7 +348,7 @@ func convertAWSTagsToCAPI(mapiTags []mapiv1.TagSpecification) capav1.Tags {
 	return capiTags
 }
 
-func convertMetadataServiceOptionstoCAPI(metad mapiv1.MetadataServiceOptions) *capav1.InstanceMetadataOptions {
+func convertMetadataServiceOptionstoCAPI(fldPath *field.Path, metad mapiv1.MetadataServiceOptions) (*capav1.InstanceMetadataOptions, *field.Error) {
 	var httpTokens capav1.HTTPTokensState
 
 	switch metad.Authentication {
@@ -351,8 +356,10 @@ func convertMetadataServiceOptionstoCAPI(metad mapiv1.MetadataServiceOptions) *c
 		httpTokens = capav1.HTTPTokensStateOptional
 	case mapiv1.MetadataServiceAuthenticationRequired:
 		httpTokens = capav1.HTTPTokensStateRequired
+	case "":
+		// This means it's optional on both sides, so no need to set anything.
 	default:
-		return &capav1.InstanceMetadataOptions{}
+		return &capav1.InstanceMetadataOptions{}, field.Invalid(fldPath.Child("authentication"), metad.Authentication, "unsupported value")
 	}
 
 	capiMetadataOpts := &capav1.InstanceMetadataOptions{
@@ -362,7 +369,7 @@ func convertMetadataServiceOptionstoCAPI(metad mapiv1.MetadataServiceOptions) *c
 		HTTPTokens: httpTokens,
 	}
 
-	return capiMetadataOpts
+	return capiMetadataOpts, nil
 }
 
 func convertIAMInstanceProfiletoCAPI(mapiIAM *mapiv1.AWSResourceReference) string {
@@ -417,6 +424,16 @@ func convertAWSBlockDeviceMappingSpecToCAPI(fldPath *field.Path, mapiBlockDevice
 			rootVolume = &volume
 
 			continue
+		}
+
+		if mapping.NoDevice != nil {
+			// Field exists in the API but is never used within the codebase.
+			errs = append(errs, field.Invalid(fldPath.Index(i).Child("noDevice"), mapping.NoDevice, "noDevice is not supported"))
+		}
+
+		if mapping.VirtualName != nil {
+			// Field exists in the API but is never used within the codebase.
+			errs = append(errs, field.Invalid(fldPath.Index(i).Child("virtualName"), mapping.VirtualName, "virtualName is not supported"))
 		}
 
 		volume, warn, err := blockDeviceMappingSpecToVolume(fldPath.Index(i), mapping, false)
--- a/pkg/conversion/mapi2capi/aws.go
+++ b/pkg/conversion/mapi2capi/aws.go
@@ -167,9 +167,16 @@ func (m *awsMachineSetAndInfra) ToMachineSetAndMachineTemplate() (*capiv1.Machin
 	capaMachineTemplate := awsMachineToAWSMachineTemplate(capaMachine, m.machineSet.Name, capiNamespace)
 
 	capiMachineSet, machineSetErrs := fromMAPIMachineSetToCAPIMachineSet(m.machineSet)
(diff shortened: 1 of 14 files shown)


pkg/conversion/test/fuzz.go
```diff
@@ -0,0 +1,235 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	runtimeserializer "k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/utils/ptr"
+
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+// CAPI2MAPIConverterConstructor is a function that constructs a CAPI to MAPI converter.
+// Since the CAPI to MAPI conversion relies on different types, it is expected that the constructor is wrapped in a closure
+// that handles type assertions to fit the interface.
```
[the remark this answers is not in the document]
JoelSpeed: Continuation of previous sentence, not sure exactly why it is wrapped where is, can either unwrap or wrap to match a certain length I guess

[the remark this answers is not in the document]
theobarberbany: The email view makes a lot more sense than gh's web ui. I think leave it as is.

pkg/conversion/mapi2capi/aws.go
```diff
@@ -167,10 +167,17 @@ func (m *awsMachineSetAndInfra) ToMachineSetAndMachineTemplate() (*capiv1.Machin
...
-	errs = append(errs, machineSetErrs.Errors()...)
+	if machineSetErrs != nil {
+		errs = append(errs, machineSetErrs.Errors()...)
+	}
 
 	capiMachineSet.Spec.Template.Spec = capiMachine.Spec
 
+	// We have to merge these two maps so that labels and annotations added to the template objectmeta are persisted
+	// along with the labels and annotations from the machine objectmeta.
+	capiMachineSet.Spec.Template.ObjectMeta.Labels = mergeMaps(capiMachineSet.Spec.Template.ObjectMeta.Labels, capiMachine.Labels)
```
nrb, line 178: As `mergeMaps` is written currently, if a label exists in both the template and the machine, the machine's value will override the template's.

If anything else is acting on the machine, this could cause drift. Should we be concerned about that, or are the corresponding VAP (and probably MAPs later) going to be enough?

pkg/conversion/mapi2capi/aws_fuzz_test.go
```diff
@@ -0,0 +1,215 @@
...
+	infraCluster := &capav1.AWSCluster{
+		Spec: capav1.AWSClusterSpec{
+			Region: "us-east-1",
+		},
+	}
+
+	Context("AWSMachine Conversion", func() {
+		fromMachineAndAWSMachineAndAWSCluster := func(machine *capiv1.Machine, infraMachine client.Object, infraCluster client.Object) capi2mapi.MachineAndInfrastructureMachine {
+			awsMachine, ok := infraMachine.(*capav1.AWSMachine)
+			Expect(ok).To(BeTrue(), "input infra machine should be of type %T, got %T", &capav1.AWSMachine{}, infraMachine)
```
nrb, line 59: I think the uses of `BeTrue` here are fine since we're outputting a message, but https://onsi.github.io/gomega/#beassignabletotypeofexpected-interface might be more appropriate in describing what we're checking.

pkg/conversion/mapi2capi/aws_fuzz_test.go
```diff
@@ -0,0 +1,215 @@
...
+*/
+package mapi2capi_test
+
+import (
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	fuzz "github.com/google/gofuzz"
```
nrb, line 24: This repo was archived on 10 Jan, 2024. They don't list a reason why, but my assumption is because Go 1.18 included fuzzing.

Did you evaluate Go's built in fuzzing? Or does apimachinery/pkg/api/apitesting/fuzzing use gofuzz?

pkg/conversion/test/fuzz.go
```diff
@@ -0,0 +1,331 @@
...
+// ObjectMetaFuzzerFuncs returns a set of fuzzer functions that can be used to fuzz ObjectMeta objects.
+// The namespace is forced to the provided namespace as the conversion always sets specific namespaces.
+// Fields that are not required for conversion are cleared.
+func ObjectMetaFuzzerFuncs(namespace string) fuzzer.FuzzerFuncs {
+	return func(codecs runtimeserializer.CodecFactory) []interface{} {
+		return []interface{}{
+			func(o *metav1.ObjectMeta, c fuzz.Continue) {
+				c.FuzzNoCustom(o)
+
+				// Force the namespace lse the conversion will fail as it always sets the namespaces deliberately.
```
nrb, line 219: lse?

pkg/conversion/test/fuzz.go
```diff
@@ -0,0 +1,331 @@
...
+				o.ResourceVersion = ""
+				o.Generation = 0
+				o.CreationTimestamp = metav1.Time{}
+				o.DeletionTimestamp = nil
+				o.DeletionGracePeriodSeconds = nil
+				o.Finalizers = nil // TODO(OCPCOUD-XXXX)
+				o.ManagedFields = nil
+
+				// Clear fields that are not currently supported in the conversion.
+				o.OwnerReferences = nil // TODO(OCPCLOUD-XXXX)
```
nrb, line 235: [OCPCLOUD-2716](https://issues.redhat.com/browse/OCPCLOUD-2716)?

pkg/conversion/mapi2capi/aws.go
JoelSpeed: If you were to look at L75 of this file, you can see where we are creating the Machine that gets converted here. We don't populate the Machine labels, so, there should be nothing in the list unless it is something we have moved from the Machine.Spec.ObjectMeta up to the Machine.ObjectMeta. That is, the labels that are managed by CAPI.

So, I think it's correct that these take priority, that said, it would find it weird if the MAPI MachineSet had a label at the Machine level that was then different at the node level right?

pkg/conversion/mapi2capi/aws_fuzz_test.go
JoelSpeed: I'm leveraging the archived library as it's used in the Kube and CAPI fuzzing. I expect at some point they will move away from it, but until they do I'm happy to keep this using the same library to be consistent.

Looking at the built in fuzzing, it looks nowhere near as advanced and doesn't look like we can even add custom fuzzer functions, so at the moment, what we have seems like the best choice.

I cannot seem to find any discussion on the K8s issues related to the fuzz library being archived, I'll poke around a bit more

pkg/conversion/mapi2capi/aws_fuzz_test.go
JoelSpeed: Yes, you're correct, but since I'm writing test code, which needs the converted type, I have to do the conversion. If I don't check the ok, the linter complains (as it should), so this is the best way to get around the linter complaining while still giving me the `awsMachine` variable of the correct type, or at least, that was my theory.

Open to other ideas though

pkg/conversion/test/fuzz.go
JoelSpeed: Should be `else`

pkg/conversion/mapi2capi/aws_fuzz_test.go
nrb: I did a quick comparison, I agree that the built in fuzz functionality isn't where we'd like it to be. Moving off the archived repo will likely require someone taking initiative upstream to bring it in tree or create a new repo.

pkg/conversion/mapi2capi/aws.go
nrb: That would be weird. As long as this is intentional 👍

pkg/conversion/mapi2capi/aws_fuzz_test.go
```diff
@@ -53,26 +53,52 @@ var _ = Describe("AWS Fuzz (mapi2capi)", func() {
...
+			return capi2mapi.FromMachineAndAWSMachineAndAWSCluster(machine, awsMachine, awsCluster)
+		}
+
+		conversiontest.MAPI2CAPIMachineRoundTripFuzzTest(
+			scheme,
+			infra,
+			infraCluster,
+			mapi2capi.FromAWSMachineAndInfra,
+			fromMachineAndAWSMachineAndAWSCluster,
+			conversiontest.ObjectMetaFuzzerFuncs("openshift-machine-api"),
```
damdo, line 73: Nit, could we set the namespace in a const so we avoid typing it every time?

pkg/conversion/mapi2capi/aws_fuzz_test.go
damdo: Also agree with using BeTrue here.

damdo (commented):
Great work here 🎉 
This is going to be very useful!

A minor nit but LGTM

pkg/conversion/mapi2capi/aws_fuzz_test.go
JoelSpeed: I have a follow up PR ready to open, can fix in that

JoelSpeed:
/override ci/prow/e2e-aws-capi-techpreview
/override ci/prow/e2e-aws-ovn
/override ci/prow/e2e-aws-ovn-techpreview
/override ci/prow/e2e-gcp-capi-techpreview 
/override ci/prow/e2e-vsphere-capi-techpreview

This is a testing only change, even the go.mod changes only affect test utilities, no point running E2E for unit changes

JoelSpeed:

And I missed this one 🤦

PR #212, opened 2024-09-27
OCPCLOUD-2643: MachineSyncReconciler - split reconciles

theobarberbany:
This change adds the logic to split the reconciliation depending on which MachineAuthority is set.

### DIFF
--- a/pkg/controllers/machinesync/machine_sync_controller.go
+++ b/pkg/controllers/machinesync/machine_sync_controller.go
@@ -21,9 +21,11 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	"github.com/openshift/cluster-capi-operator/pkg/util"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	awscapiv1beta1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta1"
@@ -32,18 +34,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const (
-	capiNamespace string = "openshift-cluster-api"
-	mapiNamespace string = "openshift-machine-api"
+	capiNamespace  string = "openshift-cluster-api"
+	mapiNamespace  string = "openshift-machine-api"
+	machineSetKind string = "MachineSet"
 )
 
 var (
 	// errPlatformNotSupported is returned when the platform is not supported.
 	errPlatformNotSupported = errors.New("error determining InfraMachine" +
 		" type, platform not supported")
+
+	// errAuthoritativeAPIUnexpected is returned when we receive an unexpected AuthoritativeAPI.
+	errAuthoritativeAPIUnexpected = errors.New("AuthoritativeAPI unexpected value")
 )
 
 // MachineSyncReconciler reconciles CAPI and MAPI machines.
@@ -100,6 +107,75 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Reconcile reconciles CAPI and MAPI machines for their respective namespaces.
 func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx, "namespace", req.Namespace, "name", req.Name)
+
+	logger.V(1).Info("Reconciling machine")
+	defer logger.V(1).Info("Finished reconciling machine")
+
+	var mapiMachineNotFound, capiMachineNotFound bool
+
+	// Get the MAPI Machine
+	mapiMachine := &machinev1beta1.Machine{}
+	if err := r.Get(ctx, req.NamespacedName, mapiMachine); apierrors.IsNotFound(err) {
+		logger.Info("MAPI Machine not found")
+
+		mapiMachineNotFound = true
+	} else if err != nil {
+		logger.Error(err, "Failed to get MAPI Machine")
+		return ctrl.Result{}, fmt.Errorf("failed to get MAPI machine: %w", err)
+	}
+
+	// Get the corresponding CAPI Machine
+	capiMachine := &capiv1beta1.Machine{}
+	capiNamespacedName := client.ObjectKey{
+		Namespace: r.CAPINamespace,
+		Name:      req.Name,
+	}
+
+	if err := r.Get(ctx, capiNamespacedName, capiMachine); apierrors.IsNotFound(err) {
+		logger.Info("CAPI Machine not found")
+
+		capiMachineNotFound = true
+	} else if err != nil {
+		logger.Error(err, "Failed to get CAPI Machine")
+		return ctrl.Result{}, fmt.Errorf("failed to get CAPI machine:: %w", err)
+	}
+
+	if mapiMachineNotFound && capiMachineNotFound {
+		logger.Info("CAPI and MAPI machines not found, nothing to do")
+		return ctrl.Result{}, nil
+	}
+
+	// We mirror if the CAPI machine is owned by a MachineSet which has a MAPI
+	// counterpart.
+	if mapiMachineNotFound {
+		if shouldReconcile, err := r.shouldMirrorCAPIMachineToMAPIMachine(ctx, logger, capiMachine); err != nil {
+			return ctrl.Result{}, err
+		} else if shouldReconcile {
+			return r.reconcileCAPItoMAPI(ctx, capiMachine, mapiMachine)
+		}
+	}
+
+	switch mapiMachine.Status.AuthoritativeAPI {
+	case machinev1beta1.MachineAuthorityMachineAPI:
+		return r.reconcileMAPIMachinetoCAPIMachine(ctx, mapiMachine, capiMachine)
+	case machinev1beta1.MachineAuthorityClusterAPI:
+		return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
+	case machinev1beta1.MachineAuthorityMigrating:
+		logger.Info("machine currently migrating", "machine", mapiMachine.GetName())
+		return ctrl.Result{}, nil
+	default:
+		return ctrl.Result{}, fmt.Errorf("%w: %s", errAuthoritativeAPIUnexpected, mapiMachine.Spec.AuthoritativeAPI)
+	}
+}
+
+// reconcileCAPItoMAPI reconciles a CAPI Machine to a MAPI Machine.
+func (r *MachineSyncReconciler) reconcileCAPItoMAPI(ctx context.Context, capiMachine *capiv1beta1.Machine, mapiMachine *machinev1beta1.Machine) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
+}
+
+// reconcileMAPItoCAPI reconciles a MAPI Machine to a CAPI Machine.
+func (r *MachineSyncReconciler) reconcileMAPItoCAPI(ctx context.Context, mapiMachine *machinev1beta1.Machine, capiMachine *capiv1beta1.Machine) (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
@@ -114,4 +190,47 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
 	default:
 		return nil, fmt.Errorf("%w: %s", errPlatformNotSupported, platform)
 	}
+}
+
+// shouldMirrorCAPIMachine takes a CAPI machine and determines if there should
+// be a MAPI mirror, it returns true only if:
+//
+// 1. The CAPI Machine is owned by a CAPI MachineSet,
+// 2. That owning CAPI MachineSet has a MAPI MachineSet Mirror.
+func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context.Context, logger logr.Logger, machine *capiv1beta1.Machine) (bool, error) {
+	logger.WithName("shouldMirrorCAPIMachineToMAPIMachine").
+		Info("checking if CAPI machine should be mirrored", "machine", machine.GetName())
+
+	// Check if owner refs point to CAPI MachineSet
+	for _, ref := range machine.ObjectMeta.OwnerReferences {
+		if ref.Kind != machineSetKind || ref.APIVersion != capiv1beta1.GroupVersion.String() {
+			continue
+		}
+
+		logger.Info("CAPI machine is owned by a machineset",
+			"machine", machine.GetName(), "machineset", ref.Name)
+		// Checks if the CAPI MS has a mirror in MAPI namespace
+		key := client.ObjectKey{
+			Namespace: r.MAPINamespace,
+			Name:      ref.Name,
+		}
+		mapiMachineSet := &machinev1beta1.MachineSet{}
+
+		if err := r.Get(ctx, key, mapiMachineSet); apierrors.IsNotFound(err) {
+			logger.Info("MAPI MachineSet mirror not found, nothing to do",
+				"machine", machine.GetName(), "machineset", ref.Name)
+
+			return false, nil
+		} else if err != nil {
+			logger.Error(err, "Failed to get MAPI MachineSet mirror")
+
+			return false, fmt.Errorf("failed to get MAPI MachineSet: %w", err)
+		}
+
+		return true, nil
+	}
+
+	logger.Info("CAPI machine is not owned by a machineset, nothing to do", "machine", machine.GetName())
+
+	return false, nil
 }
--- a/pkg/controllers/machinesync/machine_sync_controller.go
+++ b/pkg/controllers/machinesync/machine_sync_controller.go
@@ -106,6 +106,8 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // Reconcile reconciles CAPI and MAPI machines for their respective namespaces.
+//
+//nolint:funlen
 func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx, "namespace", req.Namespace, "name", req.Name)
 
@@ -114,9 +116,14 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 
 	var mapiMachineNotFound, capiMachineNotFound bool
 
-	// Get the MAPI Machine
+	// Get the MAPI Machine.
 	mapiMachine := &machinev1beta1.Machine{}
-	if err := r.Get(ctx, req.NamespacedName, mapiMachine); apierrors.IsNotFound(err) {
+	mapiNamespacedName := client.ObjectKey{
+		Namespace: r.MAPINamespace,
+		Name:      req.Name,
+	}
+
+	if err := r.Get(ctx, mapiNamespacedName, mapiMachine); apierrors.IsNotFound(err) {
 		logger.Info("MAPI Machine not found")
 
 		mapiMachineNotFound = true
@@ -125,7 +132,7 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 		return ctrl.Result{}, fmt.Errorf("failed to get MAPI machine: %w", err)
 	}
 
-	// Get the corresponding CAPI Machine
+	// Get the corresponding CAPI Machine.
 	capiMachine := &capiv1beta1.Machine{}
 	capiNamespacedName := client.ObjectKey{
 		Namespace: r.CAPINamespace,
@@ -147,12 +154,12 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 	}
 
 	// We mirror if the CAPI machine is owned by a MachineSet which has a MAPI
-	// counterpart.
+	// counterpart. This is because we want to be able to migrate in both directions.
 	if mapiMachineNotFound {
 		if shouldReconcile, err := r.shouldMirrorCAPIMachineToMAPIMachine(ctx, logger, capiMachine); err != nil {
 			return ctrl.Result{}, err
 		} else if shouldReconcile {
-			return r.reconcileCAPItoMAPI(ctx, capiMachine, mapiMachine)
+			return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -115,3 +194,39 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
...
 }
+
+// shouldMirrorCAPIMachine takes a CAPI machine and determines if there should
+// be a MAPI mirror, it returns true only if:
+//
+// 1. The CAPI Machine is owned by a CAPI MachineSet,
+// 2. That CAPI MachineSet has a MAPI MachineSet Mirror.
+func (r *MachineSyncReconciler) shouldMirrorCAPIMachine(ctx context.Context, machine *capiv1beta1.Machine, logger logr.Logger) (bool, error) {
+	// Checks if owner ref is present
+	if len(machine.ObjectMeta.OwnerReferences) > 0 {
```
JoelSpeed, line 205: Since you are ranging below, this is ineffective, you can get rid of this if condition

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -115,3 +194,39 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
...
+		// If present, checks if owner ref points to CAPI MachineSet
+		for _, ref := range machine.ObjectMeta.OwnerReferences {
+			if ref.Kind == "MachineSet" && ref.APIVersion == "cluster.x-k8s.io/v1beta1" {
```
JoelSpeed, line 208: How does this code block change if you were to invert this? Is it cleaner?
```suggestion
			if ref.Kind != "MachineSet" || ref.APIVersion != "cluster.x-k8s.io/v1beta1" {
```

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: 🤦 this occured to me and then I completely forgot to do it

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Ah yep nice, gets rid of the nasty nested ifs, ty! :)

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -100,6 +112,73 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Reconcile reconciles CAPI and MAPI machines for their respective namespaces.
 func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx, "namespace", req.Namespace, "name", req.Name)
```
damdo, line 115: Is req.Name/req.Namespace always for a MAPI Machine? Or does this also get triggered by reqs that contain a CAPI Machine?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -100,6 +112,73 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	// Get the MAPI Machine
+	mapiMachine := &machinev1beta1.Machine{}
+	if err := r.Get(ctx, req.NamespacedName, mapiMachine); apierrors.IsNotFound(err) {
+		logger.Info("MAPI Machine not found")
+
+		mapiMachineNotFound = true
+	} else if err != nil {
+		logger.Error(err, "Failed to get MAPI Machine")
+		return ctrl.Result{}, fmt.Errorf("%w: %w", errFailedToGetMAPIMachine, err)
```
damdo, line 133: Last time I did this double wrapping, @JoelSpeed wasn't too fond of it, is that still the case Joel?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -100,6 +112,73 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		logger.Info("CAPI Machine not found")
+
+		capiMachineNotFound = true
+	} else if err != nil {
+		logger.Error(err, "Failed to get CAPI Machine")
+		return ctrl.Result{}, fmt.Errorf("%w: %w", errFailedToGetCAPIMachine, err)
+	}
+
+	if mapiMachineNotFound && capiMachineNotFound {
+		logger.Info("CAPI and MAPI machines not found, nothing to do")
```
damdo, line 153: When do you envision hitting this case? When we get a req for a delete a MAPI/CAPI machine after we just deleted the other counterpart?

pkg/controllers/machinesync/machine_sync_controller.go
damdo: If this can come from either MAPI or CAPI Machines we need to change the namespace for getting the mapiMachine, otherwise if we use req.Namespace we can end up looking into the openshift-cluster-api ns for a MAPI Machine.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -100,6 +112,73 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return ctrl.Result{}, nil
+	}
+
+	// We mirror if the CAPI machine is owned by a MachineSet which has a MAPI
+	// counterpart.
```
damdo, line 158: We could add a bit more on the why for why this is.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -115,3 +194,38 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
 		return nil, fmt.Errorf("%w: %s", errPlatformNotSupported, platform)
 	}
 }
+
+// shouldMirrorCAPIMachine takes a CAPI machine and determines if there should
+// be a MAPI mirror, it returns true only if:
+//
+// 1. The CAPI Machine is owned by a CAPI MachineSet,
+// 2. That CAPI MachineSet has a MAPI MachineSet Mirror.
+func (r *MachineSyncReconciler) shouldMirrorCAPIMachine(ctx context.Context, machine *capiv1beta1.Machine, logger logr.Logger) (bool, error) {
```
damdo, line 203: I'd make it even more descriptive
```suggestion
func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context.Context, machine *capiv1beta1.Machine, logger logr.Logger) (bool, error) {
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -100,6 +112,73 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// If we have not set AuthoritativeAPI, we default to MachineAuthorityMachineAPI
+	if mapiMachine.Spec.AuthoritativeAPI == machinev1beta1.MachineAuthorityClusterAPI {
+		return r.reconcileCAPItoMAPI(ctx, capiMachine, mapiMachine)
+	} else {
+		return r.reconcileMAPItoCAPI(ctx, mapiMachine, capiMachine)
+	}
+}
+
+// reconcileCAPItoMAPI reconciles a CAPI Machine to a MAPI Machine.
+func (r *MachineSyncReconciler) reconcileCAPItoMAPI(ctx context.Context, capiMachine *capiv1beta1.Machine, mapiMachine *machinev1beta1.Machine) (ctrl.Result, error) {
```
damdo, line 176: I'd also make these more descriptive to avoid mixup :)
```suggestion
func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Context, capiMachine *capiv1beta1.Machine, mapiMachine *machinev1beta1.Machine) (ctrl.Result, error) {
```

damdo (commented):
Looks good up until now, left some minor comments

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -100,6 +112,73 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	logger.V(1).Info("Reconciling machine")
+	defer logger.V(1).Info("Finished reconciling machine")
+
+	var (
+		mapiMachineNotFound bool
+		capiMachineNotFound bool
+	)
+
+	// Get the MAPI Machine
```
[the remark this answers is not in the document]
theobarberbany: Oh weird - I think it is on (Ive had it complain at me) but it's not here for some reason...

Will change either way.

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Not 100% I'm following you here. 

> is req.Name/req.Namespace always for a MAPI Machine?

Nope, we can trigger reconciles from a CAPI machine or InfrastructureMachine we just rewrite the reconcile.Request to have the mapi namespace. 

https://github.com/openshift/cluster-capi-operator/blob/main/pkg/util/watch_filters.go#L41

Although, I agree - being explicit in looking in the MAPI namespace for the MAPI machine is less surprising :) Will update.

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: I'm not entirely sure... I was thinking about deletion, although I mostly threw this in for completeness.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Ah ok so we watch CAPI Machines but the req always corresponds to a MAPI Machine. req.Namespace is always the MAPI one then.
Ok cool, then either could've worked. NP

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Ok 👍

damdo (commented):
GG @theobarberbany 🙌 

/lgtm

Will leave the approval to @JoelSpeed

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Yeah, I think given it made you think twice keeping it as clear as possible is the way to go :)

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -99,7 +111,80 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+//nolint:funlen
 func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx, "namespace", req.Namespace, "name", req.Name)
+
+	logger.V(1).Info("Reconciling machine")
+	defer logger.V(1).Info("Finished reconciling machine")
+
+	var (
+		mapiMachineNotFound, capiMachineNotFound bool
+	)
```
JoelSpeed, line 124: Does the following not work? Wonder why you put it in parentheses?
```suggestion
	var mapiMachineNotFound, capiMachineNotFound bool
```

pkg/controllers/machinesync/machine_sync_controller.go
JoelSpeed: Yeah, I think for now lets keep to writing them out and only wrapping the base error. Until and unless we find some good reason we need to have a static error here
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to get MAPI machine: %w", err)
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -99,7 +111,80 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Name:      req.Name,
+	}
+
+	if err := r.Get(ctx, capiNamespacedName, capiMachine); apierrors.IsNotFound(err) {
+		logger.Info("CAPI Machine not found")
+
+		capiMachineNotFound = true
+	} else if err != nil {
+		logger.Error(err, "Failed to get CAPI Machine")
+		return ctrl.Result{}, fmt.Errorf("%w: %w", errFailedToGetCAPIMachine, err)
```
JoelSpeed, line 155:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to get CAPI machine: %w", errFailedToGetCAPIMachine, err)
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -99,7 +111,80 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
+		}
+	}
+
+	// If we have not set AuthoritativeAPI, we default to MachineAuthorityMachineAPI
+	if mapiMachine.Spec.AuthoritativeAPI == machinev1beta1.MachineAuthorityClusterAPI {
+		return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
+	} else {
+		return r.reconcileMAPIMachinetoCAPIMachine(ctx, mapiMachine, capiMachine)
+	}
```
JoelSpeed, line 178: Would probably prefer a switch here that explicitly checks the empty string, MAPI and CAPI string values, and then errors for an unknown value, else you're defaulting unknown values to a MAPI to CAPI direction here.

Is that intended?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -115,3 +200,38 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
 		return nil, fmt.Errorf("%w: %s", errPlatformNotSupported, platform)
 	}
 }
+
+// shouldMirrorCAPIMachineToMAPIMachine takes a CAPI machine and determines if there should
+// be a MAPI mirror, it returns true only if:
+//
+// 1. The CAPI Machine is owned by a CAPI MachineSet,
+// 2. That CAPI MachineSet has a MAPI MachineSet Mirror.
+func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context.Context, machine *capiv1beta1.Machine, logger logr.Logger) (bool, error) {
```
JoelSpeed, line 209: Idiomatically, the order of arguments should always be `ctx, logger, others`

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -115,3 +200,38 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
...
+	// Check if owner refs point to CAPI MachineSet
+	for _, ref := range machine.ObjectMeta.OwnerReferences {
+		if ref.Kind != "MachineSet" || ref.APIVersion != "cluster.x-k8s.io/v1beta1" {
```
JoelSpeed, line 212: Do we have constants we can use for this? We could use the group version stuff from `doc.go` in the CAPI api packages to start for API version

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -115,3 +200,38 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
...
+
+		// Checks if the CAPI MS has a mirror in MAPI namespace
+		key := client.ObjectKey{
+			Namespace: r.MAPINamespace,
+			Name:      ref.Name,
+		}
+		mapiMachineSet := &machinev1beta1.MachineSet{}
+
+		if err := r.Get(ctx, key, mapiMachineSet); apierrors.IsNotFound(err) {
+			logger.Info("MAPI MachineSet mirror not found, nothing to do", ref.Name)
```
JoelSpeed, line 224: Do we have a log that tells us that we are checking for an owned CAPI mirror machineset?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -115,3 +200,38 @@ func getInfraMachineFromProvider(platform configv1.PlatformType) (client.Object,
...
+		} else if err != nil {
+			logger.Error(err, "Failed to get MAPI MachineSet mirror")
+
+			return false, fmt.Errorf("%w: %w", errFailedToGetMAPIMachineSet, err)
+		}
+
+		return true, nil
+	}
+
+	return false, nil
```
JoelSpeed, line 236: What happens if the CAPI machine has no owner references at all? I think this is correct, that we don't reconcile, but have we double checked/should we add a comment?

JoelSpeed (commented):
Is there anything we can start testing here?

Perhaps factoring some of the logic out of Reconcile would help with testing the decision tree as to which reconcile to call? Though perhaps that's not currently possible given the nil implementations?

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Ah, I'd initially done them over two lines - didn't remember the parentheses

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Nope - good point!

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: >Is that intended?

Going from the comment, I was thinking so... but can't for the life of me remember why now. I think possibly I recall a webhook going in to ensure it's never unset, however that still doesn't mean this would make sense 🤔 

Either way, will update to a switch!

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: What about in cases where I don't have an error to return, then I tend to get linter errors :(  e.g: 

``` 
do not define dynamic errors, use wrapped static errors instead: "fmt.Errorf(\"unexpected value for spec.AuthoritativeAPI: %s\", mapiMachine.Spec.AuthoritativeAPI)" (err113)
                return ctrl.Result{}, fmt.Errorf("unexpected value for spec.AuthoritativeAPI: %s", mapiMachine.Spec.AuthoritativeAPI)
```

JoelSpeed: Yes in that case, create a static error, but when you're just wrapping an existing error there is no need to create additional ones unless you plan to unwrap them later. The idea is so that we have a base error that can be unwrapped to find the root cause

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: I can't find any `doc.go` with the API version, other than maybe building it from: https://github.com/kubernetes-sigs/cluster-api/blob/main/api/v1beta1/groupversion_info.go#L30 ?

Maybe just a `const` at the top of the file?

theobarberbany: Ah sweet! got it

damdo (commented):
/lgtm

Thanks for improving logging!

theobarberbany:
> Is there anything we can start testing here? 

I think we may be able to test that we `Expect` our empty `reconcileMAPIMachinetoCAPIMachine` `reconcileCAPIMachinetoMAPIMachine` get called with different inputs to test `shouldMirrorCAPIMachineToMAPIMachine` & our switch indirectly.. but that is testing the implementation rather than the behaviour.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -99,7 +106,83 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	switch mapiMachine.Spec.AuthoritativeAPI {
+	case machinev1beta1.MachineAuthorityMachineAPI:
+		return r.reconcileMAPIMachinetoCAPIMachine(ctx, mapiMachine, capiMachine)
+	case machinev1beta1.MachineAuthorityClusterAPI:
+		return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
+	case machinev1beta1.MachineAuthorityMigrating:
+		logger.Info("machine currently migrating", "machine", mapiMachine.GetName())
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
theobarberbany: Hmm I should've checked this - the linter gets sad if it's not in there, so I just threw it in :/ 

```
pkg/controllers/machinesync/machine_sync_controller.go:166:2: missing cases in switch of type v1beta1.MachineAuthority: v1beta1.MachineAuthorityMigrating (exhaustive)
        switch mapiMachine.Status.AuthoritativeAPI {
        ^
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -99,7 +106,83 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// counterpart. This is because we want to be able to migrate in both directions.
+	if mapiMachineNotFound {
+		if shouldReconcile, err := r.shouldMirrorCAPIMachineToMAPIMachine(ctx, logger, capiMachine); err != nil {
+			return ctrl.Result{}, err
+		} else if shouldReconcile {
+			return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
+		}
+	}
+
+	switch mapiMachine.Spec.AuthoritativeAPI {
```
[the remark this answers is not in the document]
theobarberbany: Ah yep this is me misremembering, it should be `Status`! 

I've had a read of the enhancement doc again, I think that this should be handled by a separate controller all together? The `migration-controller`

```
To allow the migration between Machine API and Cluster API resources and vice versa, we will implement a new controller machine-api-migration-controller. This controller will be responsible for co-ordinating the hand over of authority between the Machine API and Cluster API controllers.

The controller will observe the spec and status authoritative API fields and look for a difference between them. A difference indicating the cluster admin requested a migration between the two APIs.

When the spec and status differ, it will first check if the Synchronized condition is set to True, if it is not, this indicates an unmigratable resource and therefore should take no action.

If the condition allows, the controller will set the status.authoritativeAPI to the Migrating value and wait for the old authoritative resource to set the Paused condition to True.

Once the Paused condition is observed, the controller will move the status to match the spec allowing the new authoritative API to take over.

This is the controller that will be responsible for pausing and un-pausing Cluster API resources as part of the handover.
```


I'm not following this: 
>but, if the status is migrating, we need to know where we are coming from, so that we can complete the migration process 

If we're going MAPI -> CAPI, we'll just observe the change of `authoritativeAPI` to be CAPI and start syncing the other way.. or am I missing something?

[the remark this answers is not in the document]
JoelSpeed: No you're right, we split this out to be a separate controller, clever us! I think you just need to observe status here and handle each of the three cases appropriately. Migrating would just be a no-op :)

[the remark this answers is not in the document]
theobarberbany: Sweet - have updated to switch on `Status`

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -99,7 +106,83 @@ func (r *MachineSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	switch mapiMachine.Status.AuthoritativeAPI {
+	case machinev1beta1.MachineAuthorityMachineAPI:
+		return r.reconcileMAPIMachinetoCAPIMachine(ctx, mapiMachine, capiMachine)
+	case machinev1beta1.MachineAuthorityClusterAPI:
+		return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
+	case machinev1beta1.MachineAuthorityMigrating:
+		logger.Info("machine currently migrating", "machine", mapiMachine.GetName())
+		return ctrl.Result{}, nil
+	default:
+		return ctrl.Result{}, fmt.Errorf("%w: %s", errAuthoritativeAPIUnexpected, mapiMachine.Spec.AuthoritativeAPI)
```
JoelSpeed, line 175: Nit, this would go into a crash loop back-off, better to log the error and return a nil to prevent the back-off loop.

That said, the API validations should mean we never get here

JoelSpeed:
/lgtm
/approve

One nit which you can fix if you want to, not blocking though

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: TIL, would a single instance of this cause it to crash loop, even if all of the others were fine? I guess it would keep reconciling and keep erroring?

From the point of a user, if we are in a situation where we'd got past the API validations would we want this to surface as a log or as a more obvious failure? 🤔

JoelSpeed: Sorry, wording not great there, I don't actually mean pod level crash loop back off, I meant the reconcile would go into exponential backoff 🤦 

It will log  and get more infrequent, so, this is fine, but, I guess since it's not a transient error, retrying has no way to resolve, it would be better not to bother

theobarberbany: Ohh ok gotcha. Makes sense now!

PR #215, opened 2024-10-08
OCPCLOUD-2749: mapi2capi conversion pkg negative testing

damdo:
First part of the negative unit testing efforts for the conversion packages.
This is for the mapi2capi pkg.


pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,140 @@ limitations under the License.
...
+)
+
+var _ = DescribeTable("mapi2capi AWS convert MAPI MachineSet",
+	func(ms *mapiv1.MachineSet, i *configv1.Infrastructure, errorsMatcher types.GomegaMatcher, warningsMatcher types.GomegaMatcher) {
+		_, _, warns, err := FromAWSMachineSetAndInfra(ms, i).ToMachineSetAndMachineTemplate()
+		Expect(err).To(errorsMatcher, "should have been able to convert MAPI MachineSet to CAPI MachineSet")
+		Expect(warns).To(warningsMatcher, "should have not warned while converting MAPI MachineSet to CAPI MachineSet")
+	},
+
+	Entry("With a Base configuration", awsMAPIMachineSetBase.Build(), infra, BeNil(), BeEmpty()),
```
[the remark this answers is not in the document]
nrb: Got it, thanks.

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,140 @@ limitations under the License.
...
+	}))
+	awsMAPIMachineWithMissingEBSConfig = awsMAPIMachineBase.WithProviderSpecBuilder(awsBaseProviderSpec.WithBlockDevices([]mapiv1.BlockDeviceMappingSpec{
+		{DeviceName: ptr.To("/dev/sdb")},
+	}))
+	awsMAPIMachineWithRootVolumeNotDeleted = awsMAPIMachineBase.WithProviderSpecBuilder(awsBaseProviderSpec.WithBlockDevices([]mapiv1.BlockDeviceMappingSpec{
+		{EBS: &mapiv1.EBSBlockDeviceSpec{VolumeSize: ptr.To(int64(10)), DeleteOnTermination: ptr.To(false)}},
+	}))
+	awsMAPIMachineWithNoDeviceAndVirtualName = awsMAPIMachineBase.WithProviderSpecBuilder(awsBaseProviderSpec.WithBlockDevices([]mapiv1.BlockDeviceMappingSpec{
+		{VirtualName: ptr.To("test"), NoDevice: ptr.To("test"), EBS: &mapiv1.EBSBlockDeviceSpec{VolumeSize: ptr.To(int64(10))}},
+	}))
```
JoelSpeed, line 87: Can you add some whitespace in here, this becomes quite a wall as it stands

Out of interest, are they all based on a single base machine?

Would it be easier to create providerSpecs and pass a PS builder/built PS into the test case and have that substituted into the machine within the test case?

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,140 @@ limitations under the License.
...
-	awsMAPIMachine := machinebuilder.Machine().WithProviderSpecBuilder(machinebuilder.AWSProviderSpec()).WithProviderID("aws:///us-west-2a/i-05442bc41c3df969d").Build()
-	awsMAPIMachineSet := machinebuilder.MachineSet().WithProviderSpecBuilder(machinebuilder.AWSProviderSpec()).Build()
-	awsMAPIMachineSet.Spec.Template.Spec.ProviderID = ptr.To("aws:///us-west-2a/i-05442bc41c3df969d") // TODO: do this in machinebuilder.
-	infra := configv1.Infrastructure{
-		Spec: configv1.InfrastructureSpec{},
-		Status: configv1.InfrastructureStatus{
-			InfrastructureName: "sample-cluster-name",
-		},
+var _ = DescribeTable("mapi2capi AWS convert MAPI Machine",
+	func(m *mapiv1.Machine, i *configv1.Infrastructure, errorsMatcher types.GomegaMatcher, warningsMatcher types.GomegaMatcher) {
```
JoelSpeed, line 101: Please use an input struct rather than named positional arguments like this, it makes the entries _sooo_ much easier to read.

I'd expect the entries to look like
```
Entry("name", awsConversionInput{
  machine: awsMAPIMachineBase.Build(),
  Infra: infra,
  expectedError: "...", <- Typical pattern is to handle this within the table
  expectedWarnings: []string{},
}),
```

In the past we have even done the construction of the providerSpec differences within the Entry, so that you can see exactly the changes here, rather than having to refer up to a particular variable above

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,140 @@ limitations under the License.
...
+		_, _, warns, err := FromAWSMachineAndInfra(m, i).ToMachineAndInfrastructureMachine()
+		Expect(err).To(errorsMatcher, "should have been able to convert providerSpec to MachineTemplateSpec")
```
JoelSpeed, line 103: You're reducing the value of the table by having the matcher outside of this function, putting too much into the Entry IMO

I would have expected something more akin to
```
if in.expectedError != "" {
  Expect(err).To(MatchError(ContainSubstring(in.expectedError)), "...")
} else {
  Expect(err).To(Not(HaveOcurred()), "...")
}
```

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,140 @@ limitations under the License.
...
+		Expect(warns).To(warningsMatcher, "should have not warned while converting providerSpec to MachineTemplateSpec")
```
JoelSpeed, line 104: Perhaps so that ordering doesn't matter, you could use `ConsistOf` and pass in just a list of warnings? An empty list would still work for the cases where there are no errors

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,140 @@ limitations under the License.
...
+	},
+	// Base Case.
+	Entry("With a Base configuration", awsMAPIMachineBase.Build(), infra, BeNil(), BeEmpty()),
+	// Only Error.
+	Entry("With LoadBalancers", awsMAPIMachineWithLBs.Build(), infra, MatchError(ContainSubstring(errLoadbalancersNotSupported)), BeEmpty()),
```
JoelSpeed, line 109: I'm not a fan of the constants being used in this test case to match errors like this.

I'd prefer to see the raw string here as I can then see what an end user would see.

I also care about how the errors are constructed beyond the limit of the special text we add in. I want to be able to see that the `field.Invalid` constructed a proper message, something like
```
spec.foo.bar: Invalid Value: 123: value must be less than 100
```

Where currently all you would be checking is the final part of that, `value must be less than 100`

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,140 @@ limitations under the License.
...
+	Entry("With VirtualName specified", awsMAPIMachineWithVirtualName.Build(), infra, MatchError(ContainSubstring(errVirtualNameNotSupported)), BeEmpty()),
+	// Error + Warning.
+	Entry("With VirtualName specified and missing EBS configuration", awsMAPIMachineWithVirtualNameAndMissingEBSConfig.Build(), infra, MatchError(ContainSubstring(errVirtualNameNotSupported)), ConsistOf(ContainSubstring(warnMissingBlockDeviceConf))),
+	Entry("With VirtualName specified and root Volume not deleted on termination", awsMAPIMachineWithVirtualNameAndRootVolNotDeletedOnTerm.Build(), infra, MatchError(ContainSubstring(errVirtualNameNotSupported)), ConsistOf(ContainSubstring(warnRootVolumeMustBeDeletedOnTerminationIgnoringValueFalse))),
+	// Double Errors.
+	Entry("With NoDevice and VirtualName specified", awsMAPIMachineWithNoDeviceAndVirtualName.Build(), infra, ConsistOf(MatchError(ContainSubstring(errNoDeviceIsNotSupported)), MatchError(ContainSubstring(errVirtualNameNotSupported))), BeEmpty()),
+	// Only Warnings.
+	Entry("With missing EBS configuration", awsMAPIMachineWithMissingEBSConfig.Build(), infra, BeNil(), ConsistOf(ContainSubstring(warnMissingBlockDeviceConf))),
```
JoelSpeed, line 128: You call `Build()` on every entry for the Machine, have you considered passing a builder into the function and calling build on it within the describe table function?

pkg/conversion/util/util.go
```diff
@@ -30,3 +33,12 @@ func IsCAPIManagedLabel(key string) bool {
 		dnsSubdomainOrName == capiv1.NodeRestrictionLabelDomain || strings.HasSuffix(dnsSubdomainOrName, "."+capiv1.NodeRestrictionLabelDomain) ||
 		dnsSubdomainOrName == capiv1.ManagedNodeLabelDomain || strings.HasSuffix(dnsSubdomainOrName, "."+capiv1.ManagedNodeLabelDomain)
 }
+
+// MatchOptionalErrorDetail matches an actual error detail or a nil error detail.
+func MatchOptionalErrorDetail(errDetail string) types.GomegaMatcher {
```
JoelSpeed, line 38: Is this actually being used?

pkg/conversion/mapi2capi/aws.go
```diff
@@ -41,11 +41,26 @@ var (
...
+	errUnsupportedAuthenticationValue                                = "unsupported authentication value"
+	errNonRootVolumesMustBeDeletedOnTerminationUnsupportedValueFalse = "non-root volumes must be deleted on termination, unsupported value false"
+	errVolumeSizeRequiredButMissing                                  = "volumeSize is required, but is missing"
+	errPlacementRegionShouldMatchInfrastructureStatusValue           = "placement.region should match infrastructure status value"
+	errVirtualNameNotSupported                                       = "virtualName is not supported"
+	errNoDeviceIsNotSupported                                        = "noDevice is not supported"
+
+	// Warnings.
+	warnRootVolumeMustBeDeletedOnTerminationIgnoringValueFalse = "root volume must be deleted on termination, ignoring invalid value false"
+	warnMissingBlockDeviceConf                                 = "missing ebs configuration for block device"
```
JoelSpeed, line 63: I don't think extracting these out is the right move personally, I'm not sure it adds value and is making the errors harder to read personally, they aren't being re-used anywhere (other than in tests where I have a separate comment), I'd personally prefer these go back inline and that we match on full error strings rather than these substrings

pkg/conversion/mapi2capi/aws_test.go
JoelSpeed: I'm thinking we want these more like [this example](https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/8d85b2cb2c1dc8709fa49470b1df91640c642a59/pkg/controllers/controlplanemachineset/updates_test.go#L111) in CPMS where in Each `Entry` you have everything you need to be able to understand the complete test input, WDYT?

pkg/conversion/mapi2capi/aws_test.go
damdo: Yes let's do this in the `Entry` rather than here, it's easier to have these two details close together.

pkg/conversion/mapi2capi/aws_test.go
damdo: Yes good idea, done.

pkg/conversion/mapi2capi/aws_test.go
damdo: Yes, I agree, the matching needs to happen against multiple errors and warnings, so the logic will be slightly different, but let's do this inside the table setup function.

pkg/conversion/mapi2capi/aws.go
damdo: I tend to disagree, I dislike having to match things in text two times and having to amend both if the text needs changing.

I tried to make the var names as descriptive as possible so we shouldn't need to look up the whole text, every time we want to know what they throw.

pkg/conversion/mapi2capi/aws_test.go
damdo: Yeah I can see this can be desirable, I'll try to see if there is an easy way to do this while also retaining the err variables bit: https://github.com/openshift/cluster-capi-operator/pull/215#discussion_r1796823009

Otherwise we can switch to a full typed out match.

pkg/conversion/mapi2capi/aws.go
JoelSpeed: That's fair, lets see if you can find an alternative path in the middle somewhere

One benefit of typing it out twice though, is that I've picked up spelling mistakes that way many times 😅

pkg/conversion/mapi2capi/aws_test.go
JoelSpeed: Ooh good point, perhaps the expected errors should be a slice? Since we probably want to test that with multiple issues we can see multiple errors

damdo: Indeed!

pkg/conversion/mapi2capi/aws.go
damdo: Done! Taken an approach that uses fmt.Sprintf with a typed out approach and then uses the variable as the last bit of the error. This mimics the [CPMS approach](https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/8d85b2cb2c1dc8709fa49470b1df91640c642a59/pkg/controllers/controlplanemachineset/updates_test.go#L342).

Unfortunately I couldn't match certain warning strings composites because the warning string returned by the converter contains pointers that are not matchable from the testing side. So I kept those as matching only with the fixed string at the tail of the warning.

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -54,6 +57,13 @@ type awsMAPI2CAPIConversionInput struct {
 	expectedWarnings []string
 }
 
+type awsMAPI2CAPIMachinesetConversionInput struct {
+	machineSetBuilder machinebuilder.MachineSetBuilder
+	infra             *configv1.Infrastructure
+	expectedErrors    []string
+	expectedWarnings  []string
+}
+
```
JoelSpeed, line 66: Perhaps wrap the DescribeTable in an outer Describe and keep all the test bits within that? Else the scope leaks into the main package right?

pkg/conversion/mapi2capi/aws.go
damdo: Yeah it would yield errors like:

```
• [FAILED] [0.001 seconds]
mapi2capi AWS convert MAPI Machine [It] With VirtualName specified and missing EBS configuration
/Users/ddonati/src/cluster-capi-operator/pkg/conversion/mapi2capi/aws_test.go:247

  [FAILED] should match expected warnings while converting an AWS MAPI Machine to CAPI
  Expected
      <[]string | len:1, cap:1>: [
          "spec.providerSpec.value.blockDevices[0]: Invalid value: v1beta1.BlockDeviceMappingSpec{DeviceName:(*string)(nil), EBS:(*v1beta1.EBSBlockDeviceSpec)(nil), NoDevice:(*string)(nil), VirtualName:(*strin
g)(0xc000f85ea0)}: missing ebs configuration for block device",
      ]
  to consist of
      <[]*matchers.ContainSubstringMatcher | len:1, cap:1>: [
          {
              Substr: "spec.providerSpec.value.blockDevices[0]: Invalid value: v1beta1.BlockDeviceMappingSpec{DeviceName:(*string)(nil), EBS:(*v1beta1.EBSBlockDeviceSpec)(nil), NoDevice:(*string)(nil), Virtual
Name:(*string)(0xc000056d10)}: missing ebs configuration for block device",
              Args: nil,
          },
      ]
```
Where there is a pointer mismatch even if I declare them at the top like this:
```
	testValue                          = ptr.To[string]("test")
	blockDeviceMappingWithVirtualName  = &mapiv1.BlockDeviceMappingSpec{VirtualName: testValue}
	blockDeviceMappingWithoutEBSConfig = &mapiv1.BlockDeviceMappingSpec{DeviceName: ptr.To("/dev/sdb")}
```
And reference them in the inputs and in the expected output.

damdo: But I have a new idea on how to circumvent this.
```
- warnings = append(warnings, field.Invalid(fldPath.Index(i), mapping, warnMissingBlockDeviceConf).Error())
+ warnings = append(warnings, field.Invalid(fldPath.Index(i).Child("ebs"), mapping.EBS, warnMissingBlockDeviceConf).Error())
```
Change the warning throw like so, so we can just match against nil

damdo: WDYT?

JoelSpeed: I'm surprised it is rendering the pointers, rather than rendering something more akin to a `%+v`, but, ok, lets not stress too much on this then

damdo: I agree it is a bit overzelous in using `%#v`.
Ok then, I'll move to matching the lowest field to avoid the issue entirely. TY

pkg/conversion/mapi2capi/aws_test.go
damdo: Yes good point

theobarberbany:
Looking good! 

I agree that extracting all of those errors is a tad confusing.. but also the matching is not how I understood it to work either.... 

LGTM! :)

JoelSpeed:
@damdo Please squash the commit history and clean up here please.

I'm not sure how strongly I want to push on extracting the errors here, perhaps we should bring that up at parking lot? Seems both Theo and I prefer them inline, would be interested to see what others think

damdo:
@JoelSpeed Happy to go with what the majority prefers

pkg/conversion/mapi2capi/aws.go
JoelSpeed: The more I think about this thread, the less I like extracting these.

Think about a case where the error message gets confused. Lets say I accidentally update one of these vars to mean something else, now, when its returned in the original location, I get a nonsense message that doesn't mean what I wanted it to.

I may think it's harmless, and it probably won't get picked up in code review, because we likely aren't looking at the context in which the variable is used.

Since the test uses the same variable, the test will still pass, and this will only get picked up when someone hits the error, and we find and get confused by the nonsense string.

When the errors are typed out inline as they were before, changing them is obvious and the diff will show the context -> less likely to produce a nonsense message.

And then add to that, I have to go and update a test. That is a conscious effort to change the test expectation and we will clearly see that we are changing the expected string. I'm not a fan of calculating test output. Calculated output can change underneath us without us noticing, we should write the strings out in full to ensure that if anything does change, the tests are either broken, or updated, and we can use them as a safety net.

Note the CPMS example you chose is different because it is wrapping a specific static error that is used only in testing as a sentinel error returned by the mocking library implementing an interface, and not an error returned by the controller itself.

damdo: Had a chat offline, converged to inline errors. PR updated.

pkg/conversion/mapi2capi/aws.go
```diff
@@ -277,7 +269,7 @@ func (m *awsMachineAndInfra) toAWSMachine(providerSpec mapiv1.AWSMachineProvider
 		m.infrastructure.Status.PlatformStatus.AWS.Region != "" &&
 		providerSpec.Placement.Region != m.infrastructure.Status.PlatformStatus.AWS.Region {
 		// Assuming that the platform status has a region, we expect all MachineSets to match that region, if they don't, this is an error on the users part.
-		errs = append(errs, field.Invalid(fldPath.Child("placement", "region"), providerSpec.Placement.Region, fmt.Sprintf("placement.region should match infrastructure status value %q", m.infrastructure.Status.PlatformStatus.AWS.Region)))
+		errs = append(errs, field.Invalid(fldPath.Child("placement", "region"), providerSpec.Placement.Region, fmt.Sprintf("%s %q", "placement.region should match infrastructure status value", m.infrastructure.Status.PlatformStatus.AWS.Region)))
```
JoelSpeed, line 272: Odd change here? Probably want to revert this

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,332 @@ limitations under the License.
...
+			},
+			expectedWarnings: []string{},
+		}),
+		Entry("With DeviceIndex non-zero", awsMAPI2CAPIConversionInput{
+			machineBuilder: awsMAPIMachineBase.WithProviderSpecBuilder(
+				awsBaseProviderSpec.WithDeviceIndex(1),
+			),
+			infra: infra,
+			expectedErrors: []string{
+				fmt.Sprintf("spec.providerSpec.value.deviceIndex: Invalid value: %d: %s", 1, "deviceIndex must be 0 or unset"),
```
JoelSpeed, line 119: This doesn't need to be printed right?
```suggestion
				"spec.providerSpec.value.deviceIndex: Invalid value: 1: deviceIndex must be 0 or unset",
```

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,332 @@ limitations under the License.
...
+			expectedWarnings: []string{},
+		}),
+		Entry("With mismatched region", awsMAPI2CAPIConversionInput{
+			machineBuilder: awsMAPIMachineBase.WithProviderSpecBuilder(
+				awsBaseProviderSpec.WithRegion(awsNonDefaultRegion),
+			),
+			infra: infraWithRegion,
+			expectedErrors: []string{
+				fmt.Sprintf("spec.providerSpec.value.placement.region: Invalid value: %q: %s %q",
+					awsNonDefaultRegion, "placement.region should match infrastructure status value", infraWithRegion.Status.PlatformStatus.AWS.Region),
```
JoelSpeed, line 130: Can we avoid `fmt.Sprintf` where possible please, I don't think it's needed here for example

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,332 @@ limitations under the License.
...
+			},
+			expectedWarnings: []string{},
+		}),
+		Entry("With unsupported network interface type", awsMAPI2CAPIConversionInput{
+			machineBuilder: awsMAPIMachineBase.WithProviderSpecBuilder(
+				awsBaseProviderSpec.WithNetworkInterfaceType("unsupported-value"),
+			),
+			infra: infra,
+			expectedErrors: []string{
+				fmt.Sprintf("spec.providerSpec.value.networkInterfaceType: Invalid value: %q: %s", "unsupported-value", "networkInterface type must be one of ENA or omitted, unsupported value"),
```
JoelSpeed, line 154:
```suggestion
				"spec.providerSpec.value.networkInterfaceType: Invalid value: \"unsupported-value\" networkInterface type must be one of ENA or omitted, unsupported value",
```

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,332 @@ limitations under the License.
...
+		Entry("With AMI ARN reference", awsMAPI2CAPIConversionInput{
+			machineBuilder: awsMAPIMachineBase.WithProviderSpecBuilder(
+				awsBaseProviderSpec.WithAMI(mapiv1.AWSResourceReference{
+					ARN: ptr.To("arn:aws:ec2:us-east-1::image/ami-1234567890abcdef0"),
+				}),
+			),
+			infra: infra,
+			expectedErrors: []string{
+				fmt.Sprintf("spec.providerSpec.value.ami.arn: Invalid value: %q: %s",
+					"arn:aws:ec2:us-east-1::image/ami-1234567890abcdef0", "unable to convert AMI ARN reference. Not supported in CAPI"),
```
JoelSpeed, line 167:
```suggestion
				"spec.providerSpec.value.ami.arn: Invalid value: \"arn:aws:ec2:us-east-1::image/ami-1234567890abcdef0\" unable to convert AMI ARN reference. Not supported in CAPI",
```

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -16,43 +16,332 @@ limitations under the License.
...
+		Entry("With LoadBalancers", awsMAPI2CAPIConversionInput{
+			machineBuilder: awsMAPIMachineBase.WithProviderSpecBuilder(
+				awsBaseProviderSpec.WithLoadBalancers(
+					[]mapiv1.LoadBalancerReference{{Name: "a", Type: mapiv1.ClassicLoadBalancerType}},
+				),
+			),
+			infra: infra,
+			expectedErrors: []string{
+				fmt.Sprintf("spec.providerSpec.value.loadBalancers: Invalid value: %#v: %s",
+					[]mapiv1.LoadBalancerReference{{Name: "a", Type: mapiv1.ClassicLoadBalancerType}}, "loadBalancers are not supported"),
```
JoelSpeed, line 109: What does this actually look like when it's printed out?

PR #216, opened 2024-10-09
MULTIARCH-5043: Add Powervs support for CAPI-MAPI conversion

Karthik-K-N:


pkg/conversion/capi2mapi/powervs.go
```diff
@@ -0,0 +1,305 @@
...
+	if len(errors) > 0 {
+		return nil, warnings, utilerrors.NewAggregate(errors)
+	}
+
+	return mapiMachineSet, warnings, nil
+}
+
+// toProviderSpec converts a capi2mapi machineAndPowerVSMachineAndPowerVSCluster into a MAPI PowerVSMachineProviderConfig.
+//
+//nolint:funlen
```
JoelSpeed, line 161: Is this needed? The function doesn't look super long?

pkg/conversion/capi2mapi/powervs.go
```diff
@@ -0,0 +1,305 @@
...
+		ServiceInstance: serviceInstance,
+		Image:           image,
+		Network:         network,
+		KeyPairName:     m.powerVSMachine.Spec.SSHKey,
+		SystemType:      m.powerVSMachine.Spec.SystemType,
+		ProcessorType:   mapiv1.PowerVSProcessorType(m.powerVSMachine.Spec.ProcessorType),
+		Processors:      m.powerVSMachine.Spec.Processors,
+		MemoryGiB:       m.powerVSMachine.Spec.MemoryGiB,
+		//CredentialsSecret:
+		//LoadBalancers: TODO: Not supported for workers.
```
JoelSpeed, line 200: Are you creating Jira trackers for the feature gaps you've found so far?

pkg/conversion/capi2mapi/powervs.go
```diff
@@ -0,0 +1,305 @@
...
+	if network.Name != nil {
+		networkResource.Type = mapiv1.PowerVSResourceTypeName
+		networkResource.Name = network.Name
+		return networkResource, nil
+	}
+	if network.RegEx != nil {
+		networkResource.Type = mapiv1.PowerVSResourceTypeRegEx
+		networkResource.RegEx = network.RegEx
+		return networkResource, nil
+	}
```
JoelSpeed, line 248: Is this ordering the same as would happen within the provider code?

pkg/conversion/capi2mapi/powervs.go
```diff
@@ -0,0 +1,305 @@
...
+			return imageResource, nil
+		}
+		if image.RegEx != nil {
+			imageResource.Type = mapiv1.PowerVSResourceTypeRegEx
+			imageResource.RegEx = image.RegEx
+			return imageResource, nil
+		}
+	}
+	imageResource.Type = mapiv1.PowerVSResourceTypeName
+	imageResource.Name = &imageRef.Name
```
JoelSpeed, line 275: Wonder if refactoring this to a switch might make things simpler? 

I assume again the ordering here is similar to the implementation? You have no discriminator on the MAPI side right?

pkg/conversion/capi2mapi/powervs_fuzz_test.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+		func(imageRef *corev1.LocalObjectReference, c fuzz.Continue) {
+			imageRef.Name = c.RandString()
+
+		},
+		func(spec *capibmv1.IBMPowerVSMachineSpec, c fuzz.Continue) {
+			c.FuzzNoCustom(spec)
+
+			spec.ServiceInstanceID = ""
+			spec.Image = nil
```
JoelSpeed, line 147: Why clearing these?

pkg/conversion/mapi2capi/powervs.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	if !reflect.DeepEqual(providerSpec.ObjectMeta, metav1.ObjectMeta{}) {
+		// We don't support setting the object metadata in the provider spec.
+		// It's only present for the purpose of the raw extension and doesn't have any functionality.
+		errs = append(errs, field.Invalid(fldPath.Child("metadata"), providerSpec.ObjectMeta, "metadata is not supported"))
+	}
+
+	if len(providerSpec.LoadBalancers) > 0 {
+		// TODO: CAPIBM only applies load balancers to the control plane nodes. We should always reject LBs on non-control plane and work out how to connect the control plane LBs correctly otherwise.
```
JoelSpeed, line 240: Need to track this in Jira to follow up on right?

pkg/conversion/mapi2capi/powervs.go
```diff
@@ -0,0 +1,378 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      m.machine.Name,
+			Namespace: capiNamespace,
+		},
+		Spec: spec,
+	}, warnings, errs
+}
+
+// fromPowerVSMAPIMachineToCAPIMachine translates a PowerVS MAPI Machine to its Core CAPI Machine correspondent.
+// TODO: May be generalise to support different infrastructure machines
```
JoelSpeed, line 258: Hmm yes, making this more generally applicable seems like a good idea

pkg/conversion/mapi2capi/powervs_test.go
```diff
@@ -0,0 +1,142 @@
...
+		KeyPairName:   "ipi-sa04-418nig-jqw97-key",
+		SystemType:    "s922",
+		ProcessorType: mapiv1.PowerVSProcessorTypeShared,
+		Processors:    intstr.FromString("2"),
+		MemoryGiB:     32,
+	}
+}
+
+// rawExtensionFromPowerVSProviderSpec marshals the machine provider spec.
+func rawExtensionFromPowerVSProviderSpec(spec *mapiv1.PowerVSMachineProviderConfig) (*runtime.RawExtension, error) {
```
JoelSpeed, line 129: This keeps popping up, we need a generic util for this

pkg/conversion/test/fuzz.go
```diff
@@ -159,7 +161,7 @@ func CAPI2MAPIMachineSetRoundTripFuzzTest(scheme *runtime.Scheme, infra *configv
 	machineFuzzInputs := []TableEntry{}
 	fz := getFuzzer(scheme, fuzzerFuncs...)
 
-	for i := 0; i < 1; i++ {
+	for i := 0; i < 1000; i++ {
```
JoelSpeed, line 164: Good catch

pkg/conversion/mapi2capi/powervs.go
Karthik-K-N: Yeah, currently i made a note of it, I will create a Jira as well.

pkg/conversion/capi2mapi/powervs_fuzz_test.go
Karthik-K-N: ServiceInstanceID is deprecated, Image id identified by another filed ImageRef

pkg/conversion/capi2mapi/powervs.go
```diff
@@ -0,0 +1,305 @@
...
+			Name: userDataSecretName,
+		}
+	}
+
+	return &mapiProviderConfig, warnings, nil
+}
+
+// Conversion helpers.
+
+// TODO: May be we can use generics and support for all the platforms?
```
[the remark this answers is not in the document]
Karthik-K-N: Yeah, Shall I submit a follow up PR to do that same?

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: Its not needed.

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: Yeah, I will do

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: Yes, Id, Name followed by Regex.

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: updated it to use switch.

pkg/conversion/capi2mapi/powervs.go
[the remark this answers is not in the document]
JoelSpeed: Yeah I think that would be helpful

JoelSpeed:
Units suggest a suite timeout, can we extend the suite timeout from 10m to 30m to avoid that?

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: Added a comment.

pkg/conversion/capi2mapi/powervs_fuzz_test.go
JoelSpeed: Might be worth noting those here so we don't forget, assuming these will never be changed

pkg/conversion/mapi2capi/powervs.go
JoelSpeed: Please create the Jira and add the card number to the TODO as we did for AWS, I'd like to see that before we merge this please

pkg/conversion/mapi2capi/powervs.go
```diff
@@ -0,0 +1,371 @@
...
+
+	// mapiMachine.Spec.AuthoritativeAPI - Ignore as this is part of the conversion mechanism.
+
+	// metadata.labels - needs special handling
+	// metadata.annotations - needs special handling
+
+	errs = append(errs, handleUnsupportedMachineFields(mapiMachine.Spec)...)
+
+	return capiMachine, errs
+}
```
JoelSpeed, line 314: Apart from the APIVersion and Kind, everything here looks identical to the AWS version, can we deduplicate this easily in a new file?

pkg/conversion/mapi2capi/powervs_fuzz_test.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+		},
+		func(pc *mapiv1.PowerVSMachineProviderConfig, c fuzz.Continue) {
+			c.FuzzNoCustom(pc)
+
+			// The type meta is always set to these values by the conversion.
+			pc.Kind = "PowerVSMachineProviderConfig"
+			pc.APIVersion = mapiv1.SchemeGroupVersion.String()
+
+			pc.LoadBalancers = nil
```
JoelSpeed, line 176: Load balancers not handled yet? Do we have a todo and a jira for that somewhere?

pkg/conversion/mapi2capi/powervs_test.go
```diff
@@ -0,0 +1,140 @@
...
+		Expect(warns).To(BeEmpty(), "should have not warned while converting MAPI MachineSet to CAPI MachineSet")
+		Expect(capiMachineSet).To(Not(BeNil()), "should not have a nil CAPI MachineSet")
+		Expect(capiInfraMachineTemplate).To(Not(BeNil()), "should not have a nil CAPI MachineTemplate")
+	})
+
+})
+
+// getPowerVSProviderSpec builds and returns PowerVSProviderConfig.
+func getPowerVSProviderSpec() *mapiv1.PowerVSMachineProviderConfig {
+	// TODO: May be we can add this in machine builder?
```
JoelSpeed, line 96: Would be good to be consistent with how we've done this, the machine builder pattern is pretty nice for writing these kinds of tests

pkg/conversion/capi2mapi/powervs.go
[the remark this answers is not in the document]
Karthik-K-N: Just changed the type to interface{}

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: Created the task https://issues.redhat.com/browse/MULTIARCH-5041

pkg/conversion/capi2mapi/powervs_fuzz_test.go
Karthik-K-N: Added a comment

pkg/conversion/mapi2capi/powervs.go
Karthik-K-N: Done, Created a Jira task and added it in comment.

pkg/conversion/mapi2capi/powervs_test.go
Karthik-K-N: Just changed the param type to interface{}

pkg/conversion/mapi2capi/powervs.go
Karthik-K-N: Created util file and moved it there with kind as a parameter to identify the different types of machines.

pkg/conversion/mapi2capi/powervs_fuzz_test.go
Karthik-K-N: Done

pkg/conversion/mapi2capi/util.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// set appropriate api version
+	switch kind {
+	case awsMachineKind:
+		capiMachine.Spec.InfrastructureRef.Kind = kind
+		capiMachine.Spec.InfrastructureRef.APIVersion = awsMachineAPIVersion
+	case ibmPowerVSMachineKind:
+		capiMachine.Spec.InfrastructureRef.Kind = kind
+		capiMachine.Spec.InfrastructureRef.APIVersion = ibmPowerVSMachineAPIVersion
+	}
```
JoelSpeed, line 67: Do we need to create a switch here? May as well just pass the API version as an argument to save us having to implement this over and over?

JoelSpeed (commented):
I think lets drop this switch, it will just add unnecessary complexity I think, otherwise I think this is good, we can fix up other duplications with a refactor later

pkg/conversion/mapi2capi/util.go
Karthik-K-N: Updated, Removed the switch case , Initially added thinking in future we may need to do some machine kind specific changes , But that can be added later if required. Thank you.

pkg/conversion/mapi2capi/powervs_test.go
Karthik-K-N: I will follow up on this to add support for Power VS.

JoelSpeed:
Lets add nolint directives for the current `dupl` cases, we can look at resolving that again once this is merged

Karthik-K-N:
> `dupl` cases

Done, added nolint.

Karthik-K-N:
@JoelSpeed can this be merged or do you see any more changes required?

pkg/conversion/capi2mapi/powervs_test.go
```diff
@@ -0,0 +1,96 @@
...
+	})
+
+	It("should be able to convert a Power VS MAPI MachineSet to a Power VS CAPI MachineSet", func() {
+		mapiMachineSet, warns, err :=
+			FromMachineSetAndPowerVSMachineTemplateAndPowerVSCluster(powerVSMachineSet, powerVSMachineTemplate, capiPowerVSCluster).ToMachineSet()
+		Expect(mapiMachineSet).To(Not(BeNil()), "should not have a nil MAPI MachineSet")
+		Expect(err).ToNot(HaveOccurred(), "should have been able to convert CAPI MachineSet/PowerVSMachineTemplate to MAPI MachineSet")
+		Expect(warns).To(BeEmpty(), "should have not warned while converting CAPI MachineSet/PowerVSMachineTemplate to MAPI MachineSet")
+	})
+})
```
damdo, line 96: This has changed quite a bit since the first test. I suggest following this pattern: 
https://github.com/openshift/cluster-capi-operator/pull/222
https://github.com/openshift/cluster-capi-operator/pull/215

And maybe put together a resourcebuilder for CAPIBM, like so :)
https://github.com/openshift/cluster-api-actuator-pkg/pull/333

pkg/conversion/mapi2capi/powervs_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+var _ = Describe("mapi2capi PowerVS", func() {
```
damdo, line 34: Same as above

pkg/conversion/capi2mapi/powervs_test.go
Karthik-K-N: Should I update the this PR with similar change or Shall I submit a follow up PR?

also I will submit a PR for adding resourcebulder for CAPIBM as well soon.

damdo: I'd say let's keep it in one PR, to avoid conflict nerves and stuff :)

damdo: Thanks @Karthik-K-N

Karthik-K-N: sure will do that

Karthik-K-N: Added more tests.

Karthik-K-N:
Added more tests as suggested.

Karthik-K-N:
Updated tests witch test fixtures as well. Please take a look when possible. Thanks.

Karthik-K-N:
Re based with main, Ready for review.

pkg/conversion/capi2mapi/powervs.go
```diff
@@ -0,0 +1,305 @@
...
+		//LoadBalancers: TODO(MULTIARCH-5041): Not supported for workers.
+	}
+
+	userDataSecretName := ptr.Deref(m.machine.Spec.Bootstrap.DataSecretName, "")
+	if userDataSecretName != "" {
+		mapiProviderConfig.UserDataSecret = &mapiv1.PowerVSSecretReference{
+			Name: userDataSecretName,
+		}
+	}
+
```
JoelSpeed, line 201: Are there any fields that exist in the CAPI machine template, that are not consumer as part of the above conversion to a MAPI provider spec?

pkg/conversion/capi2mapi/powervs_fuzz_test.go
```diff
@@ -0,0 +1,178 @@
...
+				network.ID = ptr.To(c.RandString())
+			case 1:
+				network.Name = ptr.To(c.RandString())
+			case 2:
+				network.RegEx = ptr.To(c.RandString())
+			}
+		},
+		func(imageRef *corev1.LocalObjectReference, c fuzz.Continue) {
+			imageRef.Name = c.RandString()
+		},
```
JoelSpeed, line 141: Why does this one need custom fuzzing? What other fields exist inside this object?

pkg/conversion/mapi2capi/powervs.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	spec := capibmv1.IBMPowerVSMachineSpec{
+		//	ServiceInstanceID: Deprecated, Use ServiceInstance.
+		ServiceInstance: serviceInstance,
+		SSHKey:          providerSpec.KeyPairName,
+		Image:           image,
+		// ImageRef: Not required as image is set above
```
JoelSpeed, line 219: What if the source has ImageRef? Should there not be errors for folks trying to use this field?

pkg/conversion/mapi2capi/powervs.go
```diff
@@ -0,0 +1,307 @@
...
+		SystemType:    providerSpec.SystemType,
+		ProcessorType: capibmv1.PowerVSProcessorType(providerSpec.ProcessorType),
+		Processors:    providerSpec.Processors,
+		MemoryGiB:     providerSpec.MemoryGiB,
+		Network:       network,
+		// ProviderID. This is populated when this is called in higher level funcs (ToMachine(), ToMachineSet()).
+	}
+
```
JoelSpeed, line 227: Are there any fields from the providerSpec that aren't consumed in this conversion?

pkg/conversion/mapi2capi/powervs_fuzz_test.go
```diff
@@ -0,0 +1,186 @@
...
+			pc.ObjectMeta = metav1.ObjectMeta{}
+			pc.CredentialsSecret = nil
```
JoelSpeed, line 178: Should there be errors in the conversions if these fields are set? I don't see those

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: No, There are no any such fields. We almost have parity with capi and mapi machine specs.

pkg/conversion/capi2mapi/powervs_fuzz_test.go
Karthik-K-N: Not really required, I will remove.

pkg/conversion/mapi2capi/powervs.go
Karthik-K-N: In CAPI image can be specified either via imageRef its localobject reference or image its reference to name, id or regex, In mapi an Image can be only repressed with image its reference to name, id or regex.

So when converting from mapi to capi we directly set the mapi image to capi image.

When converting from capi to mapi , if imageref is set we convert it to image in mapi.

More info about imageref: there is an image controller in capi which imports image from public bucket to Power VS service instance so it can be consumed while creating a machine.

pkg/conversion/mapi2capi/powervs.go
Karthik-K-N: No there aren't any apart from Loadbalancers.

pkg/conversion/mapi2capi/powervs_fuzz_test.go
Karthik-K-N: I took AWS as reference I think we are yet to decide on the behavior wrt to credetialsecret right? https://github.com/Karthik-K-N/cluster-capi-operator/blob/ba985b0949437862a9b64d40852bccb16859abb8/pkg/conversion/mapi2capi/aws.go#L267

Karthik-K-N:
Addressed review comments, Took while as I was on vacation.

pkg/conversion/capi2mapi/powervs.go
JoelSpeed: How are you handling credentials secrets currently? Do you have a credentials secret within your providerSpec?

pkg/conversion/mapi2capi/powervs_fuzz_test.go
JoelSpeed: Right, so, do you have a TODO to fix the credentials secret somewhere? I'd expect to see a comment similar to the one you've linked on the AWS side?

Karthik-K-N: Oh, I missed to add before, Updated it now.

pkg/conversion/capi2mapi/powervs.go
Karthik-K-N: No, we don't have credential secret in provider spec, We usually mount the secret cotaining APIKEY to deployment spec https://github.com/openshift/cluster-api-provider-ibmcloud/blob/main/config/default/manager_credentials_patch.yaml, Internaly the controller code looks for it https://github.com/Karthik-K-N/cluster-api-provider-ibmcloud/blob/3add9b2939f3ec6105addb9a666f1e8aeb9d85bd/pkg/cloud/services/authenticator/authenticator.go#L29-L50

Please let me know If I misunderstood your question.

JoelSpeed: Is that the same for the machine API provider?

Karthik-K-N: Yeah, 
In MAPI we create secret: https://github.com/openshift/machine-api-provider-powervs/blob/main/examples/addons.yaml
In CAPIBM we create secret: https://github.com/openshift/cluster-api-provider-ibmcloud/blob/main/config/default/credentials.yaml

JoelSpeed: The machineset example still has a credentials secret field, which is what we need to handle https://github.com/openshift/machine-api-provider-powervs/blob/4308b9e8769dba57570c5b5ff5bc433f17fe4ef5/examples/machine-set.yaml#L41-L42

In theory, an end user could change the value to a different secret, and we need to make sure that if they do this, we handle that correctly. This is what the note on the AWS side is for, to work out how we handle this value when it is not the default

Karthik-K-N: Got the point, I think I misread MAPI provider spec to capi.
So how should I take this forward, Do you think we need to handle that now or can it be taken care later?

JoelSpeed: Add a comment similar to how the comment is in AWS in the correct places, and create a card to follow up later. I think once we've worked out how to get the AWS one moving forward, that will set a pattern for powervs

Karthik-K-N: Done, I have created a ticket [MULTIARCH-5195 ](https://issues.redhat.com//browse/MULTIARCH-5195) and added a TODO at both MAPI to CAPI and CAPI to MAPI conversion phases.

damdo: Thanks for tracking this

damdo (commented):
Amazing job @Karthik-K-N 
Thanks a lot for your hard work on this one 🎉 

/approve

Karthik-K-N:
Its a huge PR and Thanks to @JoelSpeed @damdo for continues review and guidance,  I will make sure to follow up on pending issues and keep track on upcoming changes to this conversion work and follow up on the same.

Karthik-K-N:
It need acknowledge-critical-fixes-only label, Not sure why. Should we wait for something?

JoelSpeed:

This operator is only deployed on tech preview, and this change should only affect powervs. E2E otherwise looking green

JoelSpeed:

Unrelated failure, this code wouldn't even run on this test suite

Karthik-K-N:
Thank you.

PR #217, opened 2024-10-09
OCPCLOUD-2643: MachineSetSync controller bootstrap

theobarberbany:
-> Bootstraps MachineSet Sync controller
-> Adds basic test suite & empty test.
-> Fetches resources + dynamic watching of InfraMachineTemplates

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package machinesetsync
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	"github.com/openshift/cluster-capi-operator/pkg/util"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


pkg/util/watch_filters.go
```diff
@@ -43,6 +46,60 @@ func CAPIMachineToMAPIMachine(namespace string) func(context.Context, client.Obj
 	}
 }
 
+// RewriteNamespace takes a client.Object and returns a reconcile.Request for
+// it in the namespace provided.
+//
+// It is intended for use with CAPI Machines and MachineSet requests, where we expect
```
theobarberbany, line 52: I think CAPIMachineToMAPIMachine above should be replaced by RewriteNamespace, however I'm not doing that just yet as I want https://github.com/openshift/cluster-capi-operator/pull/212 to merge first, which relies on the other naming.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	capiNamespace string = "openshift-cluster-api"
+	mapiNamespace string = "openshift-machine-api"
+)
+
+var (
+	// errPlatformNotSupported is returned when the platform is not supported.
+	errPlatformNotSupported = errors.New("error determining InfraMachineTemplate" +
+		" type, platform not supported")
```
JoelSpeed, line 46: Why split this over two lines?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -0,0 +1,117 @@
...
+	// default to the production namespaces.
+	if r.CAPINamespace == "" {
+		r.CAPINamespace = capiNamespace
+	}
+
+	if r.MAPINamespace == "" {
+		r.MAPINamespace = mapiNamespace
+	}
+
+	// TODO: InfraMachineTemplates
```
JoelSpeed, line 76: Still TODO?

pkg/util/watch_filters.go
JoelSpeed: 212 is now merged so you should be able to fix this up now

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: Ah - I was thinking about trying to keep everything wrapped to ~80 characters, but don't think it's worth it.

JoelSpeed (commented):
/lgtm
/approve

PR #219, opened 2024-10-15
OCPCLOUD-2643: Machinesetsync split reconcile

theobarberbany:
-> Updates MachineSetSync reconciler to not return an error if AuthoritativeAPI has an unexpected value
-> Adds split reconciliation logic to split the reconciliation depending on which MachineAuthority is set.
-> Reworks MachineSet test suite to use helper funcs from `/pkg/test`
-> Updates helper funcs in `/pkg/test` to contain fake MachineSet CRD

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -24,6 +24,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	"github.com/openshift/cluster-capi-operator/pkg/util"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	awscapiv1beta1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta1"
@@ -32,6 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -172,7 +168,8 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 		logger.Info("machine currently migrating", "machine", mapiMachine.GetName())
 		return ctrl.Result{}, nil
 	default:
-		return ctrl.Result{}, fmt.Errorf("%w: %s", errAuthoritativeAPIUnexpected, mapiMachine.Spec.AuthoritativeAPI)
+		logger.Info("machine AuthoritativeAPI has unexpected value", "AuthoritativeAPI", mapiMachine.Status.AuthoritativeAPI, "machine", mapiMachine.GetName())
```
theobarberbany, line 171: self-nit: I don't think I need the extra machine key here, as it should be in the logging context.... I tihnk. 

Leaving here as unsure.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -172,7 +168,8 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 		logger.Info("machine currently migrating", "machine", mapiMachine.GetName())
```
theobarberbany, line 168: Same here

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -97,7 +99,80 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // Reconcile reconciles CAPI and MAPI machines for their respective namespaces.
+//
+//nolint:funlen
```
JoelSpeed, line 103: Needed?

theobarberbany: Yep we're at 62 lines on this one :/

pkg/controllers/machinesync/machine_sync_controller.go
JoelSpeed: You have the request name and namespace, the `machine` value will be a duplicate of the `name` right?

theobarberbany: yep ill remove the additional key then

PR #221, opened 2024-10-18
OCPCLOUD-2646: CAPI to MAPI MachineSet synchronisation

theobarberbany:
These changes: 

-> Implements CAPI to MAPI sync
-> Builds out test suite for happy path

At the moment we only test the happy path in order to allow further work to build upon this (without branching off of this PR). More in depth testing to come in a later PR.


pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -188,3 +351,55 @@ func getInfraMachineTemplateFromProvider(platform configv1.PlatformType) (client
...
+func updateMAPIMachineSetSpecFromCAPI(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet, newProviderSpec machinev1beta1.ProviderSpec) {
+	logger := log.FromContext(ctx).WithValues("function", "updateMAPIMachineSetSpecFromCAPI")
+
+	// Update the ProviderSpec
+	mapiMachineSet.Spec.Template.Spec.ProviderSpec = newProviderSpec
+
+	// Update labels (CAPI labels take priority)
+	mapiMachineSet.Spec.Template.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.Labels, capiMachineSet.Spec.Template.Labels)
+
+	// Update taints ?? I can't find 'em on the CAPI MachineSet
```
theobarberbany, line 365: I'm not entirely sure how to handle taints here. 

IIUC (from grepping around the capi codebase) this may be taken care of by a bootstrap provider?

pkg/controllers/common_consts.go
```diff
@@ -32,4 +34,13 @@ const (
...
 	InfrastructureResourceName = "cluster"
+
+	// SynchronisedCondition is used to denote when a MAPI or CAPI resource is
+	// synchronised. This condition should only be true when a synchronisation
+	// controller has successfully synchronised a non-authoritative resource.
+	SynchronisedCondition machinev1beta1.ConditionType = "Synchronised"
+
+	// ReasonResourceSynchronised denotes that the resource is synchronised
+	// successfully.
+	ReasonResourceSynchronised = "ResourceSynchronised"
```
JoelSpeed, line 45: Now I hate to say this, as a brit, but, is this the spelling we should be using for this, given it is user facing? I believe the EP and existing API field use the US spelling

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -86,7 +93,8 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			handler.EnqueueRequestsFromMapFunc(util.ResolveCAPIMachineSetFromObject(r.MAPINamespace)),
 			builder.WithPredicates(util.FilterNamespace(r.CAPINamespace)),
 		).
-		Complete(r); err != nil {
+		Complete(r)
+	if err != nil {
```
JoelSpeed, line 97: Curious why this change, seems a backwards step in terms of clean code to me

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +106,243 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 }
 
-// Reconcile reconciles CAPI and MAPI machines for their respective namespaces.
-//
-//nolint:funlen
+// Reconcile reconciles CAPI and MAPI MachineSets for their respective namespaces.
 func (r *MachineSetSyncReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx, "namespace", req.Namespace, "name", req.Name)
+	logger := log.FromContext(ctx).WithValues("namespace", req.Namespace, "name", req.Name)
+	ctx = logr.NewContext(ctx, logger)
```
JoelSpeed, line 112: Why this? Is this the new shiny pattern rather than passing `ctx, logger` as the first two arguments?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +106,243 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
-	logger.V(1).Info("Reconciling machineset")
-	defer logger.V(1).Info("Finished reconciling machineset")
+	logger.V(1).Info("Reconciling MachineSet")
+	defer logger.V(1).Info("Finished reconciling MachineSet")
```
JoelSpeed, line 115: Do we want to have a conversation about consistency between capitalising these keywords?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +106,243 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
-	var mapiMachineSetNotFound, capiMachineSetNotFound bool
+	mapiMachineSet, capiMachineSet, err := r.fetchMachineSets(ctx, req.Name)
+	if err != nil {
+		return ctrl.Result{}, err
```
JoelSpeed, line 119: Error must be wrapped

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +106,243 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
-		mapiMachineSetNotFound = true
+// fetchMachineSets fetches both MAPI and CAPI MachineSets.
+func (r *MachineSetSyncReconciler) fetchMachineSets(ctx context.Context, name string) (*machinev1beta1.MachineSet, *capiv1beta1.MachineSet, error) {
+	logger := log.FromContext(ctx)
+
+	// Fetch MAPI MachineSet
+	mapiMachineSet := &machinev1beta1.MachineSet{}
+
+	err := r.Get(ctx, client.ObjectKey{Namespace: r.MAPINamespace, Name: name}, mapiMachineSet)
```
JoelSpeed, line 142: When there's a single return, why not inline into the `if`?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +106,243 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		r.updateSynchronisedConditionWithPatch(
+			ctx, mapiMachineSet, corev1.ConditionFalse,
+			"FailedToConvertMachineSet", err.Error(),
+		)
+
+		return ctrl.Result{}, fmt.Errorf("failed to convert CAPI MachineSet to MAPI MachineSet: %w", err)
+	}
+
+	for _, warning := range warns {
+		logger.Info("Warning during conversion", "warning", warning)
```
JoelSpeed, line 256: I wonder if there's a good way to expose these to end users. I would expect perhaps to use an event recorder to expose these at a minimum

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +106,243 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	oldMachineSet := mapiMachineSet.DeepCopy()
+
+	var newCondition *machinev1beta1.Condition
+	if status == corev1.ConditionTrue {
+		newCondition = conditions.TrueConditionWithReason(
+			consts.SynchronisedCondition, reason, message, //nolint:govet
+		)
+
+		// Update the synchronizedGeneration
+		mapiMachineSet.Status.SynchronizedGeneration++
```
JoelSpeed, line 328: The generation should come from the generation of the source machineset, it's not an internal counter

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +106,243 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		newCondition = conditions.FalseCondition(
+			consts.SynchronisedCondition, reason,
+			machinev1beta1.ConditionSeverityError, message, //nolint:govet
+		)
+	}
+
+	conditions.Set(&mapiMachineSet.Status.Conditions, newCondition)
+
+	patch := client.MergeFrom(oldMachineSet)
+	if err := r.Status().Patch(ctx, mapiMachineSet, patch); err != nil {
```
JoelSpeed, line 339: I think controller runtime has an apply style patch option, that we should be using here, to set the correct field owner for this particular condition

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -188,3 +351,55 @@ func getInfraMachineTemplateFromProvider(platform configv1.PlatformType) (client
 		return nil, fmt.Errorf("%w: %s", errPlatformNotSupported, platform)
 	}
 }
+
+// updateMAPIMachineSetSpecFromCAPI updates the MAPI MachineSet spec based on the CAPI MachineSet.
+func updateMAPIMachineSetSpecFromCAPI(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet, newProviderSpec machinev1beta1.ProviderSpec) {
```
JoelSpeed, line 356: Hmm, how much of this is needed? I expected most of this to be handled inside the conversion library 🤔 

I think there are a handful of things that need to be handled at a higher level, but not sure about these ones.

I think primarily the changes to be handled at a higher level are finalizers and owner references, so this may not be required at all

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -188,3 +351,55 @@ func getInfraMachineTemplateFromProvider(platform configv1.PlatformType) (client
...
+	mapiMachineSet.Spec.DeletePolicy = capiMachineSet.Spec.DeletePolicy
+
+	// Update Selector
+	mapiMachineSet.Spec.Selector = capiMachineSet.Spec.Selector
+
+	logger.Info("Updated MAPI MachineSet spec from CAPI MachineSet")
+}
+
+// updateMAPIMachineSetStatusFromCAPI updates the MAPI MachineSet status based on the CAPI MachineSet.
+func updateMAPIMachineSetStatusFromCAPI(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet) {
```
JoelSpeed, line 383: We haven't yet looked at converting status, we were focusing on spec first. When we do get to it, I think this will want to be in the conversion library itself.

Perhaps look at how this might integrate directly into the library?

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: From digging through the code, I think that it's nicer maybe.. WDYT?

It stores the logger with the keys `namespace` and `name` back in the context so any additional `log.FromContext(ctx)` get them for free... 

I thought it may be nicer / cleaner overall as I don't see us storing much in the context otherwise? Maybe it's a use?

JoelSpeed: Works for me, so long as we try to be consistent CC @damdo

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: Have discussed, will stash this until we get onto updating the conversion library.

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: Oh snap I forgot to revert this. Was playing with less nesting when refactoring other bits. Will change back.

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: I think I've tried to favour `MachineSet` over `machineset` in this. 

Happy to revert and favour `machineset` though.

JoelSpeed: CC @damdo any strong opinions?

I think for CPMS we were using the style of the product docs, so we always wrote out `control plane machine set`, maybe we want to follow suit here with that too?

theobarberbany: Hmm yep if we are using that elsewhere then we should do the same here!

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: I'll expose events alongside the log. 

The only other options I think we have are:

 - Adding an additional condition, but surfacing warnings in a condition feels like we're deviating from the intended use of conditions
 - Adding an additional field to the Machine/Machineset status, but that means additional API changes

JoelSpeed: I'm happy for warnings to stay as events for now until we find a need for them not to be

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	case machinev1beta1.MachineAuthorityClusterAPI:
-		return r.reconcileCAPIMachineSettoMAPIMachineSet(ctx, capiMachineSet, mapiMachineSet)
+		return r.reconcileCAPIMachineSetToMAPIMachineSet(ctx, capiMachineSet, mapiMachineSet)
 	case machinev1beta1.MachineAuthorityMigrating:
-		logger.Info("machine currently migrating", "machine", mapiMachineSet.GetName())
+		logger.Info("machine set is currently migrating", "machine set", mapiMachineSet.GetName())
 		return ctrl.Result{}, nil
 	default:
-		logger.Info("machine AuthoritativeAPI has unexpected value", "AuthoritativeAPI", mapiMachineSet.Status.AuthoritativeAPI)
+		logger.Info("Unexpected value for AuthoritativeAPI", "AuthoritativeAPI", mapiMachineSet.Status.AuthoritativeAPI)
```
damdo, line 178: Is name/namespace of the resource inherited and reported in the k/v for this "unexpected value" log line?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return nil
 }
 
-// Reconcile reconciles CAPI and MAPI machines for their respective namespaces.
-//
-//nolint:funlen
+// Reconcile reconciles CAPI and MAPI MachineSets for their respective namespaces.
 func (r *MachineSetSyncReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx, "namespace", req.Namespace, "name", req.Name)
+	logger := log.FromContext(ctx).WithValues("namespace", req.Namespace, "name", req.Name)
```
damdo, line 110: Could be useful to tell also the resource type (e.g. machineset or machine), as the log lines might intertwine

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -58,12 +64,13 @@ type MachineSetSyncReconciler struct {
...
 func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	infraMachineTemplate, err := getInfraMachineTemplateFromProvider(r.Platform)
 	if err != nil {
-		return fmt.Errorf("failed to get InfraMachine from Provider: %w", err)
+		return fmt.Errorf("failed to get InfraMachineTemplate from Provider: %w", err)
 	}
+
 	// Allow the namespaces to be set externally for test purposes, when not set,
 	// default to the production namespaces.
 	if r.CAPINamespace == "" {
```
damdo, line 76: We could use `DefaultManagedNamespace` and `DefaultMAPIManagedNamespace` consts for these

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// TODO: Put Gets() for Cluster + Template in helper func.
+func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx context.Context, capiMachineSet *capiv1beta1.MachineSet, mapiMachineSet *machinev1beta1.MachineSet) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithValues("function", "reconcileCAPIMachineSetToMAPIMachineSet")
+
+	// Get the AWSCluster
+	cluster := &awscapiv1beta1.AWSCluster{}
+	clusterNamespacedName := client.ObjectKey{
+		Namespace: capiMachineSet.GetNamespace(),
+		Name:      capiMachineSet.Spec.ClusterName,
+	}
```
damdo, line 203: This should be called infraCluster everywhere, for consistency with other controllers/reference terms.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	if err := r.Get(ctx, clusterNamespacedName, cluster); err != nil {
+		logger.Error(err, "Failed to get CAPI cluster", "cluster", clusterNamespacedName.Name)
```
damdo, line 206: CAPI InfraCluster

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+		r.updateSynchronizedConditionWithPatch(
+			ctx, mapiMachineSet, corev1.ConditionFalse,
+			"FailedToGetCluster", err.Error(),
```
damdo, line 210: We could go a bit more descriptive
```suggestion
			"FailedToGetCAPIInfraCluster", err.Error(),
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		)
+
+		return ctrl.Result{}, fmt.Errorf("failed to get CAPI cluster: %w", err)
```
damdo, line 213:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to get CAPI InfraCluster: %w", err)
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	// Get the AWSMachineTemplate
```
damdo, line 216:
```suggestion
	// Get the InfraMachineTemplate.
```

damdo: We could stay generic with the comments, so if we plug in generic types for these, we don't need to change them then.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	template := &awscapiv1beta1.AWSMachineTemplate{}
```
damdo, line 217: I'd refer to this as infraMachineTemplate, to know quickly what that is with a glance

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	templateNamespacedName := client.ObjectKey{
+		Namespace: capiMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace,
+		Name:      capiMachineSet.Spec.Template.Spec.InfrastructureRef.Name,
+	}
+
+	if err := r.Get(ctx, templateNamespacedName, template); err != nil {
+		logger.Error(err, "Failed to get AWSMachineTemplate", "name", templateNamespacedName.Name)
```
damdo, line 224: Also here since the Get accepts a client.Object, we could stay generic and say:
```suggestion
		logger.Error(err, "Failed to get InfraMachineTemplate", "name", templateNamespacedName.Name)
```
at the top of the function, when we switch over the providers, we could then plug in the providerName in the log lines, so we carry it with us all along.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+		r.updateSynchronizedConditionWithPatch(
+			ctx, mapiMachineSet, corev1.ConditionFalse,
+			"FailedToGetAWSMachineTemplate", err.Error(),
```
damdo, line 228:
```suggestion
			"FailedToGetCAPIInfraMachineTemplate", err.Error(),
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return ctrl.Result{}, fmt.Errorf("failed to get AWSMachineTemplate: %w", err)
+	}
+
+	// Convert the CAPI MachineSet and AWS resources to a MAPI MachineSet
+	convertedMachineSet, warns, err := capi2mapi.FromMachineSetAndAWSMachineTemplateAndAWSCluster(
+		capiMachineSet, template, cluster,
+	).ToMachineSet()
+
+	if err != nil {
+		logger.Error(err, "Failed to convert CAPI MachineSet to MAPI machine set")
```
damdo, line 240: Mix between MachineSet and machine set, we should probably settle with one across the board.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+		r.updateSynchronizedConditionWithPatch(
+			ctx, mapiMachineSet, corev1.ConditionFalse,
+			"FailedToConvertMachineSet", err.Error(),
```
damdo, line 244:
```suggestion
			"FailedToConvertCAPIMachineSetToMAPI", err.Error(),
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		)
+
+		return ctrl.Result{}, fmt.Errorf("failed to convert CAPI MachineSet to MAPI machine set: %w", err)
```
damdo, line 247: Same as above

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		r.updateSynchronizedConditionWithPatch(
+			ctx, mapiMachineSet, corev1.ConditionFalse,
+			"FailedToGetAWSMachineTemplate", err.Error(),
+		)
+
+		return ctrl.Result{}, fmt.Errorf("failed to get AWSMachineTemplate: %w", err)
+	}
+
+	// Convert the CAPI MachineSet and AWS resources to a MAPI MachineSet
+	convertedMachineSet, warns, err := capi2mapi.FromMachineSetAndAWSMachineTemplateAndAWSCluster(
```
damdo, line 235: I'd always mention the family of machine and machinesets in variable names.
So I'd go with `convertedMAPIMachineSet` or `newMAPIMachineSet`

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// Check if there are any changes after updating the spec
+	if !reflect.DeepEqual(convertedMachineSet.Spec, mapiMachineSet.Spec) {
+		logger.Info("Updating MAPI machine set spec")
+
+		if err := r.Update(ctx, convertedMachineSet); err != nil {
+			logger.Error(err, "Failed to update MAPI machine set")
+
+			r.updateSynchronizedConditionWithPatch(
+				ctx, mapiMachineSet, corev1.ConditionFalse,
+				"FailedToUpdateMAPI", err.Error(),
```
damdo, line 266:
```suggestion
				"FailedToUpdateMAPIMachineSet", err.Error(),
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return ctrl.Result{}, fmt.Errorf("failed to update MAPI machine set: %w", err)
+		}
+
+		logger.Info("Successfully updated MAPI machine set spec")
+	} else {
+		logger.Info("No changes detected in MAPI machine set spec")
+	}
+
+	r.updateSynchronizedConditionWithPatch(ctx, convertedMachineSet, corev1.ConditionTrue,
+		consts.ReasonResourceSynchronized, "Synchronized CAPI to MAPI", capiMachineSet.Generation)
```
damdo, line 278:
```suggestion
		consts.ReasonResourceSynchronized, "Successfully synchronized CAPI MachineSet to MAPI", capiMachineSet.Generation)
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	conditions.Set(&mapiMachineSet.Status.Conditions, newCondition)
+
+	// Create a Patch object using Apply
+	patch := client.MergeFrom(machinesetCopy)
+
+	data, err := patch.Data(mapiMachineSet)
+	if err != nil {
+		logger.Error(err, "failed serialising patch for log")
+	}
+	logger.Info("contents of patch", "data", data)
```
damdo, line 311: Do we always want to log this?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -106,15 +157,28 @@ var _ = Describe("MachineSetSync Reconciler", func() {
...
-			},
+			capiMachineSet = capiMachineSetBuilder.Build()
+
+			Expect(k8sClient.Create(ctx, capiMachineSet)).Should(Succeed())
+			Expect(k8sClient.Create(ctx, mapiMachineSet)).Should(Succeed())
 		})
-		Expect(err).ToNot(HaveOccurred())
+
+		// For now only happy path
+		It("should update the synchronized condition", func() {
```
damdo, line 173:
```suggestion
		It("should update the synchronized condition on the MAPI MachineSet", func() {
```

damdo (commented):
Thanks for your work on this @theobarberbany 🙌 

I left many comments but most of them are all the same things 😬
Mainly some linting heads ups and some personal preferences on verbosity and naming.

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: Oop - I"m still working out SSA Patching, this is a debug line :P

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	if err := r.Get(ctx, mapiNamespacedName, mapiMachineSet); apierrors.IsNotFound(err) {
-		logger.Info("MAPI MachineSet not found")
+	return r.syncMachineSets(ctx, mapiMachineSet, capiMachineSet)
+}
+
+// fetchMachineSets fetches both MAPI and CAPI MachineSets.
+func (r *MachineSetSyncReconciler) fetchMachineSets(ctx context.Context, name string) (*machinev1beta1.MachineSet, *capiv1beta1.MachineSet, error) {
+	logger := log.FromContext(ctx)
+
+	// Fetch MAPI MachineSet
```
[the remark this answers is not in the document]
theobarberbany: Draft PRs don't run the linter CI Job (and I've not been running locally as it's still WIP :) )

Also IIUC That linter only applies to comments describing functions (not in line comments)

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: Yup :)

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	if err := r.Get(ctx, client.ObjectKey{Namespace: r.MAPINamespace, Name: name}, mapiMachineSet); apierrors.IsNotFound(err) {
+		logger.Info("MAPI machine set not found", "name", name)
 
-		mapiMachineSetNotFound = true
+		mapiMachineSet = nil
 	} else if err != nil {
-		logger.Error(err, "Failed to get MAPI MachineSet")
-		return ctrl.Result{}, fmt.Errorf("failed to get MAPI MachineSet: %w", err)
+		logger.Error(err, "Failed to get MAPI machine set", "name", name)
```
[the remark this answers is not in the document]
theobarberbany: Hmm - I don't have a strongly held opinion here. 

I think in this case it could be redundant, although the log line will contain all of the k/v pairs we add in context to get to that failure.

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: Hmm. I'd not been thinking of using generics here when we need to add support for other platforms, although I've not given it a tonne of thought. 

I was thinking we'd have a library and a switch statement on platform that loads the correct helper/ reconcile func. In this instance, I think keeping the errors and comments provider specific make sense (i.e this code will ultimately live in `/lib/aws` or some such)

 Do we think generics would be a better approach?

JoelSpeed: I don't think Dam necessarily meant generics as in the concept in computing, but more, keeping the code generic enough since you won't always have AWS specific types here

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +105,218 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	switch mapiMachineSet.Status.AuthoritativeAPI {
 	case machinev1beta1.MachineAuthorityMachineAPI:
-		return r.reconcileMAPIMachineSettoCAPIMachineSet(ctx, mapiMachineSet, capiMachineSet)
+		return r.reconcileMAPIMachineSetToCAPIMachineSet(ctx, mapiMachineSet, capiMachineSet)
 	case machinev1beta1.MachineAuthorityClusterAPI:
-		return r.reconcileCAPIMachineSettoMAPIMachineSet(ctx, capiMachineSet, mapiMachineSet)
+		return r.reconcileCAPIMachineSetToMAPIMachineSet(ctx, capiMachineSet, mapiMachineSet)
 	case machinev1beta1.MachineAuthorityMigrating:
-		logger.Info("machine currently migrating", "machine", mapiMachineSet.GetName())
+		logger.Info("machine set is currently migrating", "machine set", mapiMachineSet.GetName())
```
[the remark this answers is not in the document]
theobarberbany: Hmm.

I'm not sure if there's an easy way to determine what direction we're going. We will have a MAPI MachineSet with `AuthoritataiveAPI=Migrating`, but we can't nessecarily rely on observing it change.

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: We get this for free somehow that I don't entirely understand right now: 

```
E1028 17:15:16.233708   14061 controller.go:316] "Reconciler error" err="failed to get CAPI InfraCluster: AWSCluster.infrastructure.cluster.x-k8s.io \"\" not found" controller="machineset" controllerGroup="machine.openshift.io" controllerKind="MachineSet" MachineSet="openshift-machine-api-rk7pk/foo" namespace="openshift-machine-api-rk7pk" name="foo" reconcileID="3d65de5d-360d-4cb2-b835-b3c7b3c82a15"
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: Ah yep - we had a sync on this yesterday, I completely misunderstood dam's comment here! 🤦 

I think we're aligned now though!

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
damdo: Ok gotcha, thanks!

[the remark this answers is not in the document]
JoelSpeed: The direction would be worked out by looking at the spec authoritative API, but I don't think we need to log that necessarily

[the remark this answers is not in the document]
theobarberbany: Ah ofc, yep that works! 

I'll still leave it out for now unless we think there's a need for it :)

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+		return ctrl.Result{}, fmt.Errorf("failed to convert CAPI MachineSet to MAPI machine set: %w", err)
+	}
+
+	for _, warning := range warns {
+		logger.Info("Warning during conversion", "warning", warning)
+		r.Recorder.Event(mapiMachineSet, corev1.EventTypeWarning, "ConversionWarning", warning)
+	}
+
+	newMapiMachineSet.Spec.Template.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.Labels, newMapiMachineSet.Spec.Template.Labels)
```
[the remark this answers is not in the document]
JoelSpeed: This also seems to be redundant since we don't persist the newMapiMachineSet, only the spec from it

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
+// updateSynchronizedConditionWithPatch updates the synchronized condition
+// using a server side apply patch. We do this to force ownership of the
+// 'Synchronized' condition and 'SynchronizedGeneration'.
+func (r *MachineSetSyncReconciler) updateSynchronizedConditionWithPatch(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, status corev1.ConditionStatus, reason, message string, generation ...int64) {
+	logger := log.FromContext(ctx).WithValues("function", "updateSynchronizedConditionWithPatch")
+
+	var conditionAc *machinev1applyconfigs.ConditionApplyConfiguration
+
+	var statusAc *machinev1applyconfigs.MachineSetStatusApplyConfiguration
```
JoelSpeed, line 287: Why not declare these inline with `:=`?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		WithStatus(status).
+		WithReason(reason).
+		WithMessage(message).
+		WithSeverity(machinev1beta1.ConditionSeverityNone)
+
+	statusAc = machinev1applyconfigs.MachineSetStatus()
+
+	if status == corev1.ConditionTrue {
+		setLastTransitionTime(consts.SynchronizedCondition, mapiMachineSet.Status.Conditions, conditionAc)
+		statusAc.WithSynchronizedGeneration(generation[0])
```
JoelSpeed, line 300: Why is generation a variadic argument?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	} else {
+		conditionAc.WithSeverity(machinev1beta1.ConditionSeverityError)
+		setLastTransitionTime(consts.SynchronizedCondition, mapiMachineSet.Status.Conditions, conditionAc)
+	}
+
+	// We need to add the conditions to the statusAc
+	statusAc.WithConditions(conditionAc)
+
+	msAc := machinev1applyconfigs.MachineSet(mapiMachineSet.GetName(), mapiMachineSet.GetNamespace()).WithStatus(statusAc)
+	if err := r.Status().Patch(ctx, mapiMachineSet, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner("machineset-sync-controller")); err != nil {
```
JoelSpeed, line 310: Why does this failure not get bubbled up? Wouldn't we want to retry if there was an error here?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -188,3 +322,33 @@ func getInfraMachineTemplateFromProvider(platform configv1.PlatformType) (client
...
+
+// hasSameState returns true if a condition has the same state as a condition
+// apply config; state is defined by the union of following fields: Type,
+// Status, Reason, Severity and Message (it excludes LastTransitionTime).
+func hasSameState(i *machinev1beta1.Condition, j *machinev1applyconfigs.ConditionApplyConfiguration) bool {
+	return i.Type == *j.Type &&
+		i.Status == *j.Status &&
+		i.Reason == *j.Reason &&
+		i.Severity == *j.Severity &&
+		i.Message == *j.Message
```
JoelSpeed, line 353: Was this copied from somewhere? Not sure I agree with message being included here, in fact, why is it any more than type and status?

We should look up the definition of the last transition time 🤔 

My concern is that this will constantly tick up for messages that change only in trivial details, like timestamps

pkg/util/ssa.go
```diff
@@ -0,0 +1,46 @@
...
+	return types.ApplyPatchType
+}
+
+// Data implements Patch.
+func (p applyConfigPatch) Data(_ client.Object) ([]byte, error) {
+	return json.Marshal(p.applyConfig) //nolint:wrapcheck
+}
+
+// ApplyConfigPatch is a helper function for using applyConfigPatch.
+func ApplyConfigPatch(applyConfig interface{}) client.Patch {
```
JoelSpeed, line 42: Why not use the one from controller runtime?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -21,30 +21,48 @@ import (
...
 	var mgrDone chan struct{}
 	var mgr manager.Manager
+	var k komega.Komega
 	var reconciler *MachineSetSyncReconciler
 
-	var namespace *corev1.Namespace
-	var namespaceName string
+	var syncControllerNamespace *corev1.Namespace
+	var CapiNamespace *corev1.Namespace
+	var MapiNamespace *corev1.Namespace
```
JoelSpeed, line 53:
```suggestion
	var capiNamespace *corev1.Namespace
	var mapiNamespace *corev1.Namespace
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
theobarberbany: Ugh - I changed this today and forgot to come back to it. 

The API Server rejects the output from the conversion library, which is why we don't just pass `newMapiMachineSet`. 

I've got a TODO in my notes to dig into this - as it's weird. The error is a 'not found' - despite the test env creating & observing updates (e.g. to `status.AuthoritativeAPI`)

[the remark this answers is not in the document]
theobarberbany: > This also seems to be redundant since we don't persist the newMapiMachineSet, only the spec from it

I think this is included? We are updating `.Spec.Template.Labels` and persisting everything under `.Spec`?

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: > Was this copied from somewhere?

Yep, the MAO conditions package. I figured we should keep it consistent with that. https://github.com/openshift/machine-api-operator/blob/master/pkg/util/conditions/gettersetter.go#L160-L168 

I'm not sure if we're including timestamps in messages? Will outputs from `err.Error()` contain timestamps?

pkg/util/ssa.go
theobarberbany:
```go
// applyPatch uses server-side apply to patch the object.
type applyPatch struct{}

// Type implements Patch.
func (p applyPatch) Type() types.PatchType {
	return types.ApplyPatchType
}

// Data implements Patch.
func (p applyPatch) Data(obj Object) ([]byte, error) {
	// NB(directxman12): we might technically want to be using an actual encoder
	// here (in case some more performant encoder is introduced) but this is
	// correct and sufficient for our uses (it's what the JSON serializer in
	// client-go does, more-or-less).
	return json.Marshal(obj)
}

```

```go
	if err := r.Status().Patch(ctx, mapiMachineSet, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner("machineset-sync-controller")); err != nil {

```

IIUC we can't pass an apply config to controller runtime Apply patch implementation. 

- The Apply configs don't implement `.DeepCopy()` so don't satisfy `client.Object` (we can't use Data() then)
- We can't actually pass a patch made using `applyPatch` without specifying a raw patch and separate data. 

The if we just passed the type `client.Apply` to `Status().Patch()` the logic just passes the thing we're trying to update as `obj` to `Data()`, so we can't use ApplyConfigs - and if we don't use them we'd own all fields without `omit empty`.

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: > We should look up the definition of the last transition time 🤔

For Pods we define it: Timestamp for when the Pod last transitioned from one status to another. 

So, I think it depends on how we define a distinct status, and if we only care about True/False or if messages / reasons matter?

pkg/util/ssa.go
JoelSpeed: Ok, lets keep this

I wonder what the point of their implementation is then, if it clearly doesn't work 🤔

pkg/controllers/machinesetsync/machineset_sync_controller.go
JoelSpeed: Returning an aggregate would be good yes, but not necessary, you already log the previous error as far as I can tell. Have a see how hard working an aggregate in might be

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: >Have you looked at all at how CAPI does this?
Nope, will take a look now

theobarberbany: Discussing elsewhere: 

In library-go we only use `Status`, so let's keep in line with that. 
https://github.com/openshift/library-go/blame/master/pkg/operator/v1helpers/canonicalize.go#L46


>  My thought here is that you're passing API errors into the condition message, and I'm pretty sure there could be a timestamp somewhere in there, especially if it's an error about updating something that has a timestamp in

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: I don't think it's too bad - have done: 

```go
		mapiMachineSet.Spec = newMapiMachineSet.Spec

		if err := r.Update(ctx, mapiMachineSet); err != nil {
			logger.Error(err, "Failed to update MAPI machine set")

			updateErr := fmt.Errorf("failed to update MAPI machine set: %w", err)

			if condErr := r.updateSynchronizedConditionWithPatch(
				ctx, mapiMachineSet, corev1.ConditionFalse,
				"FailedToUpdateMAPIMachineSet", err.Error(), nil); condErr != nil {
				return ctrl.Result{}, utilerrors.NewAggregate([]error{updateErr, condErr})
			}

```

pkg/util/ssa.go
theobarberbany: I think it will work, but it take ownership of any fields in the resource being patched without an `omitempty` :/

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
theobarberbany: Screenshot 2024-10-30 at 15 30 45

It was the namespace having come from the CAPI MachineSet giving a not found, which makes sense 😂 copying over the resource version was also required, but now means we can apply the output of the conversion library! :D

pkg/util/ssa.go
JoelSpeed: Why is our implementation any different then?

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
JoelSpeed: Hmm, I would expect the conversion library to set the appropriate namespaces on the way out, perhaps that's a bug?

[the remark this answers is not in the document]
theobarberbany: Yeah, it feels like it although I'm not sure how we could currently: 

```go
newMapiMachineSet, warns, err := capi2mapi.FromMachineSetAndAWSMachineTemplateAndAWSCluster(
		capiMachineSet, infraMachineTemplate, infraCluster).ToMachineSet()
```

We don't pass the existing MAPI MachineSet, so unless we passed the namespace (which is totally an option...) then how would we know what to set it to?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		r.Recorder.Event(mapiMachineSet, corev1.EventTypeWarning, "ConversionWarning", warning)
+	}
+
+	newMapiMachineSet.Spec.Template.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.Labels, newMapiMachineSet.Spec.Template.Labels)
+
+	// Check if there are any changes after updating the spec
+	if !reflect.DeepEqual(newMapiMachineSet.Spec, mapiMachineSet.Spec) {
+		logger.Info("Updating MAPI machine set")
+
+		// The MachineSet that comes out of the conversion package is rejected by the APIServer (we don't set anything other than `.spec` currently).
```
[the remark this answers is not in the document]
theobarberbany: Fubar'd here - see other comments where we'll handle this. Resolving for now.

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
theobarberbany: I've opened https://github.com/openshift/cluster-capi-operator/issues/229 as I don't think we should try fixing the conversion namespace bug in this PR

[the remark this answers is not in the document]
JoelSpeed: Hmm yes, fair point, will need to look into this, as there are definitely areas within the conversion library where it does force the namespace. Seems like that's being done inconsistently right now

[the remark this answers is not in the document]
theobarberbany: yeah, also this controller allows setting of custom namespaces (e.g for testing), if we want to keep that option then that'll inform how we modify the conversion library

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+// fetchMachineSets fetches both MAPI and CAPI MachineSets.
+func (r *MachineSetSyncReconciler) fetchMachineSets(ctx context.Context, name string) (*machinev1beta1.MachineSet, *capiv1beta1.MachineSet, error) {
+	logger := log.FromContext(ctx)
+
+	mapiMachineSet := &machinev1beta1.MachineSet{}
 
-		mapiMachineSetNotFound = true
+	if err := r.Get(ctx, client.ObjectKey{Namespace: r.MAPINamespace, Name: name}, mapiMachineSet); apierrors.IsNotFound(err) {
+		logger.Info("MAPI machine set not found", "name", name)
```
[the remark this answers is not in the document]
theobarberbany: I wasn't - good catch!

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// reconcileCAPIMachineSetToMAPIMachineSet reconciles a CAPI MachineSet to a
+// MAPI MachineSet.
+//
+// TODO: Platform specific implementation (currently this works only for AWS,
+// we want a switch on platform somewhere).
+// TODO: Put Gets() for Cluster + Template in helper func.
+//
+//nolint:funlen
+func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx context.Context, capiMachineSet *capiv1beta1.MachineSet, mapiMachineSet *machinev1beta1.MachineSet) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithValues("function", "reconcileCAPIMachineSetToMAPIMachineSet")
```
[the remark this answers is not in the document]
theobarberbany: nope - just experimenting with the logging + debugging reasons, gonna remove. 

I don't think it adds much in terms of assisting with debugging from logs.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +103,217 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	for _, warning := range warns {
+		logger.Info("Warning during conversion", "warning", warning)
+		r.Recorder.Event(mapiMachineSet, corev1.EventTypeWarning, "ConversionWarning", warning)
+	}
+
+	newMapiMachineSet.Spec.Template.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.Labels, newMapiMachineSet.Spec.Template.Labels)
+
+	// Check if there are any changes after updating the spec
+	if !reflect.DeepEqual(newMapiMachineSet.Spec, mapiMachineSet.Spec) {
```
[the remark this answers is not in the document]
theobarberbany: Have added a helper func to only compare the parts of `ObjectMeta` we care about.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +104,224 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
-// reconcileMAPIMachineSettoCAPIMachineSet MAPI MachineSet to a CAPI MachineSet.
-func (r *MachineSetSyncReconciler) reconcileMAPIMachineSettoCAPIMachineSet(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet) (ctrl.Result, error) {
-	return ctrl.Result{}, nil
+// reconcileCAPIMachineSetToMAPIMachineSet reconciles a CAPI MachineSet to a
+// MAPI MachineSet.
+//
+// TODO: Platform specific implementation (currently this works only for AWS,
+// we want a switch on platform somewhere).
+// TODO: Put Gets() for Cluster + Template in helper func.
```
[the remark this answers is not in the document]
theobarberbany: yes :)

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +104,224 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Namespace: capiMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace,
+		Name:      capiMachineSet.Spec.Template.Spec.InfrastructureRef.Name,
+	}
+
+	if err := r.Get(ctx, templateNamespacedName, infraMachineTemplate); err != nil {
+		getErr := fmt.Errorf("failed to get InfraMachineTemplate: %w", err)
+
+		if condErr := r.updateSynchronizedConditionWithPatch(
+			ctx, mapiMachineSet, corev1.ConditionFalse,
+			"FailedToGetCAPIInfraMachineTemplate", err.Error(), nil); condErr != nil {
```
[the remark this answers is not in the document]
theobarberbany: oop yep 🤦

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -21,30 +21,48 @@ import (
...
 	"github.com/openshift/cluster-capi-operator/pkg/test"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/config"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-var _ = Describe("MachineSetSync Reconciler", func() {
+var _ = Describe("With a running controller", func() {
```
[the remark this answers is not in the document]
theobarberbany: ah yep - i was copying CPMS for this

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -58,20 +78,21 @@ type MachineSetSyncReconciler struct {
 	MAPINamespace string
 }
 
-// SetupWithManager sets the CoreClusterReconciler controller up with the given manager.
+// SetupWithManager sets up the controller with the Manager.
 func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	infraMachineTemplate, err := getInfraMachineTemplateFromProvider(r.Platform)
 	if err != nil {
-		return fmt.Errorf("failed to get InfraMachine from Provider: %w", err)
+		return fmt.Errorf("failed to get InfraMachineTemplate from Provider: %w", err)
```
JoelSpeed, line 85: Should this be `infrastructure machine template`?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +119,250 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Namespace: r.CAPINamespace,
-		Name:      req.Name,
+	return mapiMachineSet, capiMachineSet, nil
+}
+
+// fetchCAPIInfraResources fetches the provider specific infrastructure resources depending on which provider is set.
+func (r *MachineSetSyncReconciler) fetchCAPIInfraResources(ctx context.Context, capiMachineSet *capiv1beta1.MachineSet) (client.Object, client.Object, error) {
+	var infraCluster client.Object
+
+	var infraMachineTemplate client.Object
```
JoelSpeed, line 179: Maybe?
```suggestion
	var infraCluster, infraMachineTemplate client.Object
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -98,88 +119,250 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		}
+
+		awsCluster, ok := infraCluster.(*awscapiv1beta1.AWSCluster)
+		if !ok {
+			return nil, nil, fmt.Errorf("%w, expected AWSCluster, got %T", errUnexpectedInfraClusterType, infraCluster)
+		}
+
+		converter = capi2mapi.FromMachineSetAndAWSMachineTemplateAndAWSCluster(
+			capiMachineSet, awsMachineTemplate, awsCluster,
+		)
```
JoelSpeed, line 318: Why not just return directly here?
```suggestion
		return capi2mapi.FromMachineSetAndAWSMachineTemplateAndAWSCluster(
			capiMachineSet, awsMachineTemplate, awsCluster,
		).ToMachineSet()
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: yup, good catch :D

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: I think my reasoning here was that we can assign in the switch, and return outside of it at the end. 

Not strongly holding this opinion though, is returning from within the switch better? I guess this way we have fewer `nolint` comments ?

damdo (commented):
Some non-blocking nits, but it is ready to go I think :)

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -68,32 +86,82 @@ var _ = Describe("MachineSetSync Reconciler", func() {
...
+		Expect(k8sClient.Create(ctx, syncControllerNamespace)).To(Succeed())
+
+		mapiNamespace = corev1resourcebuilder.Namespace().
+			WithGenerateName("openshift-machine-api-").Build()
+		Expect(k8sClient.Create(ctx, mapiNamespace)).To(Succeed())
+
+		capiNamespace = corev1resourcebuilder.Namespace().
+			WithGenerateName("openshift-cluster-api-").Build()
+		Expect(k8sClient.Create(ctx, capiNamespace)).To(Succeed())
+
```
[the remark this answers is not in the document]
theobarberbany: Good shout - but can these (and the others) be bundled in with the PR for testing I'm working on (that branches off of this?) might help us get this in sooner :D

damdo (commented):
Fine with (optionally) having the nits I added fixed in the upcoming test PR:

/lgtm

Will leave approval to @JoelSpeed

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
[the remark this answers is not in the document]
damdo: Yes sure, all these nits can go in the coming one :)

JoelSpeed:
Go mod conflict to resolve, but otherwise LGTM

/approve

damdo (commented):
/lgtm

PR #222, opened 2024-10-19
OCPCLOUD-2749: capi2mapi conversion pkg negative testing

damdo:
Second part (following up from the first https://github.com/openshift/cluster-capi-operator/pull/215) of the negative unit testing efforts for the conversion packages.
This is for the capi2mapi pkg.

### DIFF
--- a/pkg/conversion/capi2mapi/aws_test.go
+++ b/pkg/conversion/capi2mapi/aws_test.go
@@ -18,75 +18,240 @@ package capi2mapi
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
+	capibuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/cluster-api/core/v1beta1"
+	capabuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/cluster-api/infrastructure/v1beta2"
+	"github.com/openshift/cluster-capi-operator/pkg/conversion/test/matchers"
 	"k8s.io/utils/ptr"
 	capav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
-	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-var _ = Describe("capi2mapi AWS", func() {
+var _ = Describe("capi2mapi AWS conversion", func() {
+	var (
+		awsCAPIMachineBase    = capibuilder.Machine()
+		awsCAPIAWSMachineBase = capabuilder.AWSMachine()
+		awsCAPIAWSClusterBase = capabuilder.AWSCluster()
+	)
 
-	awsMachineTemplate := &capav1.AWSMachineTemplate{
-		Spec: capav1.AWSMachineTemplateSpec{
-			Template: capav1.AWSMachineTemplateResource{
-				Spec: capav1.AWSMachineSpec{
-					ProviderID: ptr.To("test-123"),
-					InstanceID: ptr.To("test-123"),
-					PublicIP:   ptr.To(false),
-				},
-			},
-		},
-		Status: capav1.AWSMachineTemplateStatus{},
+	type awsCAPI2MAPIMachineConversionInput struct {
+		machineBuilder    capibuilder.MachineBuilder
+		awsMachineBuilder capabuilder.AWSMachineBuilder
+		awsClusterBuilder capabuilder.AWSClusterBuilder
+		expectedErrors    []string
+		expectedWarnings  []string
 	}
 
-	awsMachine := &capav1.AWSMachine{
-		Spec: capav1.AWSMachineSpec{
-			ProviderID: ptr.To("test-123"),
-			InstanceID: ptr.To("test-123"),
-			PublicIP:   ptr.To(false),
-		},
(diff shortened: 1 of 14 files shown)


pkg/conversion/capi2mapi/machine_test.go
```diff
@@ -0,0 +1,82 @@
...
+			expectedErrors:   []string{"spec.nodeDrainTimeout: Invalid value: v1.Duration{Duration:1000000000}: nodeDrainTimeout is not supported"},
+			expectedWarnings: []string{},
+		}),
+		Entry("With unsupported NodeVolumeDetachTimeout", capi2MAPIMachineConversionInput{
+			machineBuilder:   capiMachineBase.WithNodeVolumeDetachTimeout(ptr.To(metav1.Duration{Duration: 1 * time.Second})),
+			expectedErrors:   []string{"spec.nodeVolumeDetachTimeout: Invalid value: v1.Duration{Duration:1000000000}: nodeVolumeDetachTimeout is not supported"},
+			expectedWarnings: []string{},
+		}),
+		Entry("With unsupported NodeDeletionTimeout", capi2MAPIMachineConversionInput{
+			machineBuilder:   capiMachineBase.WithNodeDeletionTimeout(ptr.To(metav1.Duration{Duration: 1 * time.Second})),
```
theobarberbany, line 77: tiniest nit: we could use a const for the timeout

pkg/conversion/capi2mapi/aws_test.go
```diff
@@ -18,75 +18,240 @@ package capi2mapi
...
+			awsMachineBuilder: awsCAPIAWSMachineBase.
+				WithInstanceMetadataOptions(&capav1.InstanceMetadataOptions{
+					HTTPEndpoint:            capav1.InstanceMetadataEndpointStateDisabled,
+					HTTPPutResponseHopLimit: 2,
+					HTTPTokens:              "unsupported",
+					InstanceMetadataTags:    capav1.InstanceMetadataEndpointStateEnabled,
+				}),
+			machineBuilder: awsCAPIMachineBase,
+			expectedErrors: []string{
+				"spec.instanceMetadataOptions.httpTokens: Invalid value: \"unsupported\": unable to convert httpTokens state, unknown value, " +
```
theobarberbany, line 227: nit: why are these not separate string slices?

pkg/conversion/capi2mapi/aws_test.go
```diff
@@ -18,75 +18,240 @@ package capi2mapi
...
+	)
+
+	var _ = DescribeTable("capi2mapi AWS convert CAPI MachineSet/InfraMachineTemplate/InfraCluster to MAPI MachineSet",
+		func(in awsCAPI2MAPIMachinesetConversionInput) {
+			_, warns, err := FromMachineSetAndAWSMachineTemplateAndAWSCluster(
+				in.machineSetBuilder.Build(),
+				in.awsMachineTemplateBuilder.Build(),
+				in.awsClusterBuilder.Build(),
+			).ToMachineSet()
+			Expect(err).To(matchers.ConsistOfMatchErrorSubstrings(in.expectedErrors),
```
theobarberbany, line 242: another nit: we could helper-func this expect block, as we use it twice

theobarberbany (commented):
/lgtm! 

A few non blocking nits  / questions from me :)

pkg/conversion/capi2mapi/aws_test.go
damdo: It must be a unique string because instanceMetadataOptions are aggregated on their own in the first place https://github.com/openshift/cluster-capi-operator/blob/3c1f8bc317280313eb3d9ae988525b66fba767f9/pkg/conversion/capi2mapi/aws.go#L274, differently from all the other errors, which are only aggregated once at the end.

pkg/conversion/capi2mapi/aws_test.go
damdo: The failure comment is different though, not sure it is worth it :)

pkg/conversion/capi2mapi/machine_test.go
damdo: Talking to @JoelSpeed regarding test cases, my understanding is that the preferred way of setting this up is with inline values rather than variables, to have a gauge at the input and output without extra lookups/search of values in variables, especially if they are tiny values like this and not big structs & such.
Which I think I kind of agree with.

JoelSpeed: Yeah I think I'd keep this inline so it's easy to gauge (and change) individual test cases when needed

pkg/conversion/capi2mapi/aws_test.go
JoelSpeed: > are aggregated on their own

Perhaps that is a bug that we should fix 🤔

pkg/conversion/capi2mapi/aws_test.go
JoelSpeed: I'd be inclined to keep the duplication on this one

pkg/conversion/capi2mapi/machine_test.go
theobarberbany: Ah ok - that makes sense! :)

JoelSpeed (commented):
/approve

PR #230, opened 2024-11-05
OCPCLOUD-2646: MachineSetSync controller: tests

theobarberbany:
This PR adds tests outside of the happy path for the MachineSetSync controller going in the direction from capi -> mapi 

#221 should merge before this.

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -62,6 +62,7 @@ const (
 	reasonFailedToGetCAPIInfraResources       = "FailedToGetCAPIInfraResources"
 	reasonFailedToConvertCAPIMachineSetToMAPI = "FailedToConvertCAPIMachineSetToMAPI"
 	reasonFailedToUpdateMAPIMachineSet        = "FailedToUpdateMAPIMachineSet"
+	reasonFailedToGetCAPIMachineSet           = "FailedToGetCAPIMachineSet"
 	reasonResourceSynchronized                = "ResourceSynchronized"
 
 	messageSuccessfullySynchronized = "Successfully synchronized CAPI MachineSet to MAPI"
@@ -210,18 +211,31 @@ func (r *MachineSetSyncReconciler) fetchCAPIInfraResources(ctx context.Context, 
 func (r *MachineSetSyncReconciler) syncMachineSets(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
-	switch mapiMachineSet.Status.AuthoritativeAPI {
-	case machinev1beta1.MachineAuthorityMachineAPI:
+	authoritativeAPI := mapiMachineSet.Status.AuthoritativeAPI
+
+	switch {
+	case authoritativeAPI == machinev1beta1.MachineAuthorityMachineAPI:
 		return r.reconcileMAPIMachineSetToCAPIMachineSet(ctx, mapiMachineSet, capiMachineSet)
-	case machinev1beta1.MachineAuthorityClusterAPI:
+	case authoritativeAPI == machinev1beta1.MachineAuthorityClusterAPI && capiMachineSet == nil:
+		// We want to create a new CAPI MachineSet from the MAPI one.
+		// I think we may want to call a lot of the same logic we'll need for reconciling MAPI -> CAPI,
+		// as such I think we should hold off on implementing this until that logic is worked out
+		// (and hopefully it's just calling some of the same helper funcs)
+		// TODO: Implementation
+	case authoritativeAPI == machinev1beta1.MachineAuthorityClusterAPI && capiMachineSet != nil:
 		return r.reconcileCAPIMachineSetToMAPIMachineSet(ctx, capiMachineSet, mapiMachineSet)
-	case machinev1beta1.MachineAuthorityMigrating:
-		logger.Info("machine set is currently being migrated", "machine set", mapiMachineSet.GetName())
+
+	case authoritativeAPI == machinev1beta1.MachineAuthorityMigrating:
+		logger.Info("machine set is currently being migrated")
 		return ctrl.Result{}, nil
+
 	default:
 		logger.Info("unexpected value for authoritativeAPI", "AuthoritativeAPI", mapiMachineSet.Status.AuthoritativeAPI)
+
 		return ctrl.Result{}, nil
 	}
+
+	return ctrl.Result{}, nil
 }
 
 // reconcileMAPIMachineSetToCAPIMachineSet reconciles a MAPI MachineSet to a CAPI MachineSet.
--- a/pkg/controllers/machinesetsync/machineset_sync_controller_test.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller_test.go
@@ -29,11 +29,13 @@ import (
 	corev1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/core/v1"
 	machinev1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	capav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+
 	capiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
 
+	"github.com/openshift/cluster-api-actuator-pkg/testutils"
 	consts "github.com/openshift/cluster-capi-operator/pkg/controllers"
-	"github.com/openshift/cluster-capi-operator/pkg/test"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/config"
@@ -88,17 +90,16 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 		By("Setting up a namespace for the test")
 		syncControllerNamespace = corev1resourcebuilder.Namespace().
 			WithGenerateName("machineset-sync-controller-").Build()
-		Expect(k8sClient.Create(ctx, syncControllerNamespace)).To(Succeed())
+		Expect(k8sClient.Create(ctx, syncControllerNamespace)).To(Succeed(), "sync controller namespace should be able to be created")
 
 		mapiNamespace = corev1resourcebuilder.Namespace().
 			WithGenerateName("openshift-machine-api-").Build()
-		Expect(k8sClient.Create(ctx, mapiNamespace)).To(Succeed())
+		Expect(k8sClient.Create(ctx, mapiNamespace)).To(Succeed(), "mapi namespace should be able to be created")
 
 		capiNamespace = corev1resourcebuilder.Namespace().
 			WithGenerateName("openshift-cluster-api-").Build()
-		Expect(k8sClient.Create(ctx, capiNamespace)).To(Succeed())
+		Expect(k8sClient.Create(ctx, capiNamespace)).To(Succeed(), "capi namespace should be able to be created")
 
-		By("Setting up the builders")
 		mapiMachineSetBuilder = machinev1resourcebuilder.MachineSet().
 			WithNamespace(mapiNamespace.GetName()).
 			WithName("foo").
@@ -108,10 +109,9 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 		// reference it on the CAPI MachineSet
 		capaMachineTemplateBuilder = capav1builder.AWSMachineTemplate().
 			WithNamespace(capiNamespace.GetName()).
-			WithGenerateName("machine-template-")
+			WithName("machine-template")
 
 		capaMachineTemplate = capaMachineTemplateBuilder.Build()
-		Expect(k8sClient.Create(ctx, capaMachineTemplate)).To(Succeed())
 
 		capiMachineTemplate := capiv1beta1.MachineTemplateSpec{
 			Spec: capiv1beta1.MachineSpec{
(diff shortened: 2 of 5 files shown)


pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -56,12 +56,16 @@ var (
 
 	// errUnexpectedInfraClusterType is returned when we receive an unexpected InfraCluster type.
 	errUnexpectedInfraClusterType = errors.New("unexpected InfraCluster type")
+
+	// errCAPIMachineSetNotFound is returned when we have only a MAPI MachineSet with AuthorititaveAPI set to CAPI.
+	errCAPIMachineSetNotFound = errors.New("error synchronizing CAPI to MAPI, CAPI machineset not found")
```
JoelSpeed, line 61:
```suggestion
	errCAPIMachineSetNotFound = errors.New("error synchronizing CAPI to MAPI, CAPI machine set not found")
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -121,7 +125,7 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Reconcile reconciles CAPI and MAPI MachineSets for their respective namespaces.
 func (r *MachineSetSyncReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx).WithValues("namespace", req.Namespace, "name", req.Name)
+	logger := log.FromContext(ctx)
```
JoelSpeed, line 128: Why not include the name and namespace on every log?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -214,9 +218,16 @@ func (r *MachineSetSyncReconciler) syncMachineSets(ctx context.Context, mapiMach
 	case machinev1beta1.MachineAuthorityMachineAPI:
 		return r.reconcileMAPIMachineSetToCAPIMachineSet(ctx, mapiMachineSet, capiMachineSet)
 	case machinev1beta1.MachineAuthorityClusterAPI:
+		if capiMachineSet == nil {
+			condErr := r.updateSynchronizedConditionWithPatch(
+				ctx, mapiMachineSet, corev1.ConditionFalse, reasonFailedToGetCAPIMachineSet, errCAPIMachineSetNotFound.Error(), nil)
+
+			return ctrl.Result{}, condErr
+		}
+
```
JoelSpeed, line 227: So in this case, the MAPI machine set exists, and the CAPI machine set does not?

And the authority means that the CAPI machine set, that doesn't currently exist, should be authoritative?

The intention here is that the controller creates a CAPI machine set to, become the authoritative API, based on the MAPI machine set, and then, on the next reconcile, things go back to normal

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,280 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+				})
+
+				It("should update the replica count", func() {
+					Eventually(k.Object(mapiMachineSet), timeout).Should(
+						HaveField("Spec.Replicas", Equal(ptr.To(int32(4)))),
+					)
+				})
+			})
+
+			Context("when the object meta differs", func() {
```
JoelSpeed, line 227: What about when the object meta differs for cases that we don't care about?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,280 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+								HaveField("Type", Equal(consts.SynchronizedCondition)),
+								HaveField("Status", Equal(corev1.ConditionFalse)),
+								HaveField("Severity", Equal(machinev1beta1.ConditionSeverityError)),
+								HaveField("Reason", Equal("FailedToGetCAPIMachineSet")),
+								HaveField("Message", Equal("error synchronizing CAPI to MAPI, CAPI machineset not found")),
+							))),
+					))
+
+			})
+
```
JoelSpeed, line 324: Left a comment on this, but this isn't actually the intended behaviour in this case, the sync should create the CAPI MachineSet based on the MAPI MachineSet

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,280 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+				Expect(k8sClient.Create(ctx, capiMachineSet)).Should(Succeed())
+				Expect(k8sClient.Create(ctx, mapiMachineSet)).Should(Succeed())
+
+				By("Setting the AuthoritativeAPI to Migrating")
+				Eventually(k.UpdateStatus(mapiMachineSet, func() {
+					mapiMachineSet.Status.AuthoritativeAPI = machinev1beta1.MachineAuthorityMigrating
+				})).Should(Succeed())
+			})
+
+			It("should not make any changes", func() {
```
JoelSpeed, line 342: What's this test actually doing? Are we creating a scenario where it would normally be trying to update one side or the other, and now it's not? How is it obvious that it should only have been trying to update the mapi side?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,280 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+					mapiMachineSet.Status.AuthoritativeAPI = ""
+				})).Should(Succeed())
+			})
+
+			// I think there may be another case here, where we have had it previously set - synchronized and then it gets unset.
+			// We may want to ensure that we don't change resourceVersion, and that the condition does not change?
+			// Changes to false with a failure reason?
+			//
+			// I'm not sure if we have an API validation that stops this.
+			// TODO: clarify / work out
```
JoelSpeed, line 370: I believe we have validation to say that it cannot be cleared once populated, but lets check that

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,280 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+
+			It("should not make any changes", func() {
+				resourceVersion := mapiMachineSet.GetResourceVersion()
+				Consistently(k.Object(mapiMachineSet), timeout).Should(
+					HaveField("ResourceVersion", Equal(resourceVersion)),
+				)
```
JoelSpeed, line 376: Why is it only the MAPI side that we need to check?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,280 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+			It("should not make any changes to the CAPI MachineSet", func() {
+				resourceVersion := capiMachineSet.GetResourceVersion()
+				Consistently(k.Object(capiMachineSet), timeout).Should(
+					HaveField("ResourceVersion", Equal(resourceVersion)),
+				)
+			})
+
+			// Do we have another case here where we consistently want to not see a MAPI MachineSet created?
+			// Not sure if this is overkill
+
```
JoelSpeed, line 397: I'd say consistently no new MAPI MachineSets seems reasonable yes, in the future we may decide to change that behaviour so we should test it, as we have decided it for now

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,280 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+		})
+
 	})
+
+	Context("when the infrastructure resources don't exist", func() {
+		Context("when the MAPI machine set has MachineAuthority set to Cluster API", func() {
+			BeforeEach(func() {
+				By("Creating the CAPI and MAPI MachineSets")
+				mapiMachineSet = mapiMachineSetBuilder.Build()
+				capiMachineSet = capiMachineSetBuilder.WithReplicas(int32(4)).Build()
```
JoelSpeed, line 407: Why are we not checking that the replicas count is copied over? I agree not to do in-depth testing of the conversion since that's handled elsewhere, but, this seems like an easy smoke test to show that it is actually employing conversion

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: So I noticed when looking at test output we're already getting them from somewhere - i'm not sure where it's set up but this ends up being extra noise :/ 

```
I1107 15:12:35.353217   96821 machineset_sync_controller.go:161] "MAPI machine set not found"
controller="machineset" controllerGroup="machine.openshift.io" controllerKind="MachineSet"
MachineSet="openshift-machine-api-v4752/foo" namespace="openshift-machine-api-v4752" name="foo"
reconcileID="152f0deb-bb27-4ba5-a542-2e3e9ca914cc"
```

theobarberbany: I do wonder if this is coming from how we bootstrap envtest,  but haven't seen anything obvious.

JoelSpeed: Looks like controller runtime internals 🤔 I'd potentially still include the name and namespace tbh

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: I think we do further up - I just forgot to change when copypasta: 

```go
				It("should update the replica count", func() {
					Eventually(k.Object(mapiMachineSet), timeout).Should(
						HaveField("Spec.Replicas", Equal(ptr.To(int32(4)))),
					)
				})
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: I've updated the main `switch` in the reconcile func, and put a comment in. 

I think we should handle this after we've written MAPI -> CAPI as maybe we call a subset of that logic :)

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: I don't think we have a validation currently, here's where I've looked: 

https://github.com/openshift/api/blob/master/machine/v1beta1/types_machineset.go#L158-L167 

looks like we've got one that forces moving through Migrating on the status. 

https://github.com/openshift/api/blob/master/machine/v1beta1/types_machineset.go#L64-L77 

on the spec we say it must be one of CAPI or MAPI. 

I'm not sure I see anything stopping a particularly determined user from manually editing the status? Or am I missing something

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: yep good catch, we need both :D

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: Ah I'd forgotten to add differences between the two machinesets. 🤦 

I've updated this so we have n replicas on one, and n+3 on the other. In migrating we should see both remain the same, but if we reconcile in either direction we'll get a different resource version and fail the test.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: yep forgot this case, will add a 'when we don't care about the change' context :P

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
JoelSpeed: Perhaps we should add a validation to make sure the field cannot be removed once set 🤔

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,56 +155,326 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+
+			// I can't work out a nice way to write this as I'm tired :)
+
+			// It("should not create a MAPI MachineSet", func() {
+			// 	Consistently(k.ObjectList(&machinev1beta1.MachineSetList{}), timeout).ShouldNot(
+			// 		ContainElement(WithTransform(
+			// 			func(ms machinev1beta1.MachineSet) string {
+			// 				return ms.GetName()
+			// 			}, Equal(capiMachineSet.GetName()))))
+			// })
```
JoelSpeed, line 442: I think something along the lines of
```suggestion
			 It("should not create a MAPI MachineSet", func() {
			 	Consistently(k.ObjectList(&machinev1beta1.MachineSetList{}), timeout).ShouldNot(
			 			ContainElement(HaveField("ObjectMeta.Name", Equal(capiMachineSet.GetName()))),
			 		),
			 	)
			 })
```

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
damdo: So adding something like:
```
// +kubebuilder:validation:XValidation:rule="self != '' || oldSelf == ''",message="Once authoritativeAPI is set, it cannot be unset."
```
to the API?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
damdo: Ended up with:
```go
It("should not create a MAPI MachineSet", func() {
  Consistently(k.ObjectList(&machinev1beta1.MachineSetList{}), timeout).ShouldNot(HaveField("Items",
	  ContainElement(HaveField("ObjectMeta.Name", Equal(capiMachineSet.GetName()))),
  ))
})
```

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
damdo: I see so setting this at the to of `MachineSetStatus`
and:

```
// +kubebuilder:validation:XValidation:rule="self.authoritativeAPI != '' || oldSelf.authoritativeAPI == ''",message="The authoritativeAPI field cannot be unset once it has been set."
```
?

And if this is the right approach, will this also pass validation in case of creation without authoritativeAPI set?

JoelSpeed: Yes, something like that should work. Since it includes `oldSelf`, it's considered a transition rule, and therefore doesn't execute on create. So a status without a valid value on create will be accepted still, which is what we need

PR #233, opened 2024-11-06
MULTIARCH-5151:  Generate IBMPowerVSCluster resource for CAPI clusters

Karthik-K-N:
Generate IBMPowerVSCluster resource

PR #234, opened 2024-11-06
MULTIARCH-5152: Update Power VS e2e tests to remove use of deprecated fields

Karthik-K-N:
Remove creatingIBMPowerVS cluster as its already created via https://github.com/openshift/cluster-capi-operator/pull/233
Remove deprecated ServiceInstanceID field usage.


Karthik-K-N:
/close 
Combined the changes along with https://github.com/openshift/cluster-capi-operator/pull/233

PR #236, opened 2024-11-13
OCPCLOUD-2202: Create and manage core cluster object

racheljpg:
Hello! This is a PR to ensure that each cluster should have a cluster object in the `openshift-cluster-api` namespace, as well as updating the unit/e2e tests to reflect this. Thank you!


pkg/controllers/corecluster/corecluster.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return nil
+}
+
+// Reconcile reconciles the core cluster object for the openshift-cluster-api namespace.
+func (r *CoreClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx, "namespace", req.Namespace, "name", req.Name)
+
+	// Still have to patch the condition onto a newly created cluster
+	cluster, err := CreateCoreCluster(r.Client, r.Cluster.Name, r.Infra.Kind)
```
racheljpg, line 64: I moved the creation into it's own function for readability - but maybe that's unnecessary?

pkg/controllers/corecluster/corecluster.go
```diff
@@ -0,0 +1,145 @@
...
+				APIVersion: "infrastructure.cluster.x-k8s.io/v1beta1",
+				Kind:       infraClusterKind,
+				Name:       clusterName,
+				Namespace:  CAPINamespace,
+			},
+		},
+	}
+
+	if infraClusterKind == "VSphereCluster" {
+		host, port, err := GetControlPlaneHostAndPort(cl)
```
racheljpg, line 133: I'm a bit confused on the checking of the host and port for VSphereClusters - [is it necessary to take this function from the e2e tests](https://github.com/openshift/cluster-capi-operator/blob/ec9120bb11ace966a9046ad8e698216f8abf5c42/e2e/framework/util.go#L12), or is there a more succinct version I can do here?

racheljpg (commented):
It's a bit easier to ask my questions as a review, thanks :)

pkg/controllers/corecluster/corecluster.go
damdo: That could work, but I see this actually doesn't contain a create call, so it shouldn't probably be called "Create". If you want to actually perform the creation in there then it is probably best to make it a `r.ensureCoreCluster()` similarly to what we did for the `r.ensureInfraCluster()`, but without the extra layer for providers. So basically end up with having the `r.ensureCoreCluster()` have a structure similar to https://github.com/openshift/cluster-capi-operator/blob/ec9120bb11ace966a9046ad8e698216f8abf5c42/pkg/controllers/infracluster/aws.go#L34-L86.

Then later followed by the status patching logic.

pkg/controllers/corecluster/corecluster.go
damdo: Let's not branch on infraClusterKind from here.
Let's do it in the same way as https://github.com/openshift/cluster-capi-operator/blob/ec9120bb11ace966a9046ad8e698216f8abf5c42/pkg/controllers/infracluster/aws.go#L49-L61

racheljpg: Thanks, I've updated it to look more like that now

pkg/controllers/corecluster/corecluster.go
racheljpg: Let me know what you think of it now? I left the creation out of the ensure function but as you're saying here maybe it's best to keep it together to be tidier

pkg/controllers/corecluster/corecluster.go
```diff
@@ -0,0 +1,160 @@
...
+	if !cluster.DeletionTimestamp.IsZero() {
+		if err := r.SetStatusAvailable(ctx, ""); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	logger.V(1).Info("Reconciling core cluster")
+	defer logger.V(1).Info("Finished reconciling core cluster")
```
damdo, line 94: This will need to be moved to the bottom

pkg/controllers/corecluster/corecluster.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if !cluster.DeletionTimestamp.IsZero() {
+		if err := r.SetStatusAvailable(ctx, ""); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	logger.V(1).Info("Reconciling core cluster")
```
damdo, line 93: This will need to be moved to the top.

pkg/controllers/corecluster/corecluster.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if err := r.Create(ctx, cluster); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to create Core Cluster: %w", err)
+	}
+
+	logger.Info(fmt.Sprintf("InfraCluster '%s/%s' successfully created", CAPINamespace, r.Infra.Status.InfrastructureName))
+
+	if err := r.Client.Get(ctx, req.NamespacedName, cluster); err != nil && !errors.IsNotFound(err) {
+		return ctrl.Result{}, fmt.Errorf("failed to get core cluster: %w", err)
+	}
```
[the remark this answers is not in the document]
damdo: This can be removed now that you moved it into the ensure func

pkg/controllers/corecluster/corecluster.go
```diff
@@ -0,0 +1,173 @@
...
+
+	defer logger.V(1).Info("Finished reconciling core cluster")
+
+	return ctrl.Result{}, nil
+}
+
+// CreateCluster creates a cluster with the given name and returns the cluster object.
+func (r *CoreClusterReconciler) ensureCoreCluster(ctx context.Context, clusterName, infraClusterKind string, logger logr.Logger) (*clusterv1.Cluster, error) {
+	cluster := &clusterv1.Cluster{}
```
damdo, line 127: You'll need to set the namespace and name for the cluster if you want to use it with `client.ObjectKeyFromObject`

damdo: Otherwise you can use 
```
clusterKey := client.ObjectKey{
		Namespace: CAPINamespace,
		Name:      clusterName,
	}
```

racheljpg: Missed that, thank you!

pkg/controllers/corecluster/corecluster.go
```diff
@@ -0,0 +1,192 @@
...
+			ControlPlaneEndpoint: clusterv1.APIEndpoint{
+				Host: apiURL.Hostname(),
+				Port: int32(port),
+			},
+		},
+	}
+
+	clusterCopy := cluster.DeepCopy()
+
+	conditions.MarkTrue(cluster, clusterv1.ControlPlaneInitializedCondition)
```
racheljpg, line 170: I think this whole chunk setting the available condition may be nicer in it's own function

racheljpg: (I've done that now)

Makefile
```diff
@@ -40,7 +40,7 @@ migration:
 	go build -o bin/machine-api-migration cmd/machine-api-migration/main.go
 
 unit:
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin --index https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml)" ./hack/test.sh "./pkg/... ./assets/..." 5m
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin --index https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml)" ./hack/test.sh "./pkg/controllers/cluster" 5m
```
damdo, line 43: Let's not check this in

pkg/controllers/cluster/corecluster_test.go
```diff
@@ -0,0 +1,195 @@
...
+					ClusterOperatorStatusClient: operatorstatus.ClusterOperatorStatusClient{
+						Client: cl,
+					},
+					Cluster:  &capiv1.Cluster{},
+					Infra:    ocpInfrastructureCopy.DeepCopy(),
+					Platform: ocpInfrastructureCopy.Status.PlatformStatus.Type,
+				}
+			})
+
+			PContext("When there is no Infra Cluster", func() {
```
damdo, line 90: This `P` should be removed if we want to exercise it

racheljpg:
@JoelSpeed would appreciate a review from you when you can! Thank you :)

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	capiInfraClusterAPIVersionV1Beta1 = "infrastructure.cluster.x-k8s.io/v1beta1"
+	capiInfraClusterAPIVersionV1Beta2 = "infrastructure.cluster.x-k8s.io/v1beta2"
+	clusterOperatorName               = "cluster-api"
+)
+
+var (
+	errPlatformStatusShouldNotBeNil                   = errors.New("infrastructure PlatformStatus should not be nil")
+	errMappingOCPPlatformToInfraClusterKindAndVersion = errors.New("failed to map OpenShift Infrastructure PlatformType to a Cluster API infra cluster kind, platform not supported")
+	errOCPInfraShouldNotBeNil                         = errors.New("ocp infrastructure object should not be nil")
+	errOCPInfrastructureNameShouldNotBeEmpty          = errors.New("ocp infrastructure object's status.infrastructureName should not be empty")
```
JoelSpeed, line 55: Should we use `openshift` rather than `ocp`? In fact, better, I'd drop the `ocp` prefix. Bear in mind that openshift has many different names and `ocp` isn't always accurate

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if err := r.ensureCoreClusterControlPlaneInitializedCondition(ctx, cluster); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to ensure core cluster has the ControlPlaneInitializedCondition: %w", err)
+	}
+
+	if err := r.SetStatusAvailable(ctx, ""); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
+	}
+
+	defer logger.V(1).Info("Finished reconciling core cluster")
```
JoelSpeed, line 114: A defer at the end here is kinda weird, should it not be immediately after the start log on line 86?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,243 @@
...
+
+	logger.Info(fmt.Sprintf("core cluster %s/%s does not exist, creating it", clusterObjectKey.Namespace, clusterObjectKey.Name))
+
+	if r.Infra.Status.PlatformStatus == nil {
+		return nil, errPlatformStatusShouldNotBeNil
+	}
+
+	infraClusterKind, infraClusterAPIVersion, err := mapOCPPlatformToInfraClusterKindAndVersion(r.Platform)
+	if err != nil {
+		return nil, fmt.Errorf("unable to map OpenShift Infrastucture PlatformType to Cluster API infra cluster kind: %w", err)
```
JoelSpeed, line 136: I'd drop `OpenShift` from errors and also try to use the language we would see in product docs, I believe that would be more like `Infrastructure resource platform type`
```suggestion
		return nil, fmt.Errorf("unable to map Infrastructure resource platform type to Cluster API infra cluster kind: %w", err)
```

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -0,0 +1,244 @@
...
+	var ocpInfrastructureCopy *configv1.Infrastructure
+
+	Context("With a supported platform", func() {
+		BeforeEach(func() {
+			By("creating the testing namespace")
+			namespace := corev1resourcebuilder.Namespace().WithGenerateName("test-capi-corecluster-").Build()
+			Expect(cl.Create(context.Background(), namespace)).To(Succeed())
+			testNamespaceName = namespace.Name
+
+			//TODO: implement unit tests across multiple providers
```
JoelSpeed, line 56: When do we want to handle this?

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -0,0 +1,244 @@
...
+			Context("When there is no Infra Cluster", func() {
+				It("should error saying there is no infra cluster", func() {
+					By("running the core cluster controller reconciliation")
+					_, err := r.Reconcile(ctx, reconcile.Request{
+						NamespacedName: types.NamespacedName{
+							Namespace: testNamespaceName,
+							Name:      testInfraName,
+						},
+					})
+					Expect(err).To(HaveOccurred()) // TODO match error.
```
JoelSpeed, line 101: Can we fix error matching before we merge?

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -0,0 +1,244 @@
...
+			Context("When there is an infra cluster", func() {
+				var infraCluster *capav1.AWSCluster
+				BeforeEach(func() {
+					infraCluster = capabuilder.AWSCluster().WithName(testInfraName).WithNamespace(testNamespaceName).Build()
+					Expect(cl.Create(ctx, infraCluster)).To(Succeed())
+				})
+
+				It("should create the core cluster", func() {
+					By("running the core cluster controller reconciliation")
+					_, err := r.Reconcile(ctx, reconcile.Request{
```
JoelSpeed, line 128: Do we also need to have tests where the controller is running and reporting errors in another way? Rather than calling reconcile directly?

pkg/controllers/corecluster/corecluster_controller_test.go
racheljpg: Done :) (for now, have to add a test or two more)

pkg/controllers/corecluster/corecluster_controller_test.go
racheljpg: Looking into this

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -85,35 +84,46 @@ var _ = Describe("Reconcile Core cluster", func() {
...
+					Platform: openshiftInfrastructureCopy.Status.PlatformStatus.Type,
 				}
 			})
 
 			Context("When there is no Infra Cluster", func() {
 				It("should error saying there is no infra cluster", func() {
 					By("running the core cluster controller reconciliation")
+					expectedErrors := struct {
+						infraClusterNotound string
+						coreClusterNotFound string
```
damdo, line 97:
```suggestion
						infraClusterNotFound string
						coreClusterNotFound string
```

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -230,18 +239,34 @@ var _ = Describe("Reconcile Core cluster", func() {
 
 		It("should fail the reconciliation", func() {
 			By("running the core cluster controller reconciliation")
+			expectedErrors := struct {
+				platformTypeNotSupported string
+			}{
+				platformTypeNotSupported: "failed to map OpenShift Infrastructure PlatformType to a Cluster API infra cluster kind, platform not supported: \"None\"",
```
damdo, line 245:
```suggestion
				platformTypeNotSupported: "failed to map Infrastructure PlatformType to a Cluster API infra cluster kind, platform not supported: \"None\"",
```

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -49,10 +49,10 @@ const (
 )
 
 var (
-	errPlatformStatusShouldNotBeNil                   = errors.New("infrastructure PlatformStatus should not be nil")
-	errMappingOCPPlatformToInfraClusterKindAndVersion = errors.New("failed to map OpenShift Infrastructure PlatformType to a Cluster API infra cluster kind, platform not supported")
-	errOCPInfraShouldNotBeNil                         = errors.New("ocp infrastructure object should not be nil")
-	errOCPInfrastructureNameShouldNotBeEmpty          = errors.New("ocp infrastructure object's status.infrastructureName should not be empty")
+	errPlatformStatusShouldNotBeNil                         = errors.New("infrastructure PlatformStatus should not be nil")
+	errMappingOpenshiftPlatformToInfraClusterKindAndVersion = errors.New("failed to map OpenShift Infrastructure PlatformType to a Cluster API infra cluster kind, platform not supported")
```
damdo, line 53:
```suggestion
	errMappingOpenshiftPlatformToInfraClusterKindAndVersion = errors.New("failed to map Infrastructure PlatformType to a Cluster API infra cluster kind, platform not supported")
```

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -0,0 +1,413 @@
...
+					}{
+						infraClusterNotFound: "awsclusters.infrastructure.cluster.x-k8s.io \"test-ocp-infrastructure-name\" not found",
+						coreClusterNotFound:  "clusters.cluster.x-k8s.io \"test-ocp-infrastructure-name\" not found",
+					}
+
+					_, err := r.Reconcile(ctx, reconcile.Request{
+						NamespacedName: types.NamespacedName{
+							Namespace: testNamespaceName,
+							Name:      testInfraName,
+						},
```
damdo, line 165: We should get rid of these now that we are setting up `startManager()` in the top level BeforeEach

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -0,0 +1,413 @@
...
+
+				Expect(testCoreCluster.Status.Conditions).ToNot(BeEmpty())
+				Expect(testCoreCluster.Status.Conditions[0].Type).To(Equal(capiv1.ControlPlaneInitializedCondition))
+				Expect(testCoreCluster.Status.Conditions[0].Status).To(Equal(corev1.ConditionTrue))
+			})
+		})
+	})
+})
+
+var _ = Describe("Reconcile Core cluster on unsupported platform", func() {
```
damdo, line 289: Ideally this could be a Context forking off in the main Describe rather than being its own one.

damdo (commented):
I think you are going in the right direction with the test refactor to use the manager instead of the reconciler :)

Left a couple of things we should improve

pkg/controllers/corecluster/corecluster_controller_test.go
racheljpg: Couldn't wrap my head around how to lay that out since the top describe's BeforeEach is creating an AWS cluster and for this context we want a cluster on an unsupported platform! Is there some way to clarify I want a cluster on a different platform than what's being created in the top BeforeEach?
When I had it in the same describe, it was failing, because it couldn't create the cluster on an unknown platform since an AWS one already existed :) but it's possible I was missing something and had it laid out incorrectly

racheljpg: (we resolved this, there is in fact a tidy way to do this without two describes 🤓)

damdo:
/assign @JoelSpeed 
For re-review

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -0,0 +1,250 @@
...
+			mgrCancel, mgrDone = startManager(infra)
+		})
+
+		JustAfterEach(func() {
+			stopManager()
+		})
+
+		Context("When there is no core cluster", func() {
+			It("should not create a core cluster", func() {
+				By("starting IT here")
```
JoelSpeed, line 243: Is this for debugging?

pkg/controllers/corecluster/corecluster_controller_test.go
```diff
@@ -0,0 +1,250 @@
...
+				testCoreCluster := capibuilder.Cluster().WithName(testInfraName).WithNamespace(testNamespaceName).Build()
+				Consistently(komega.Get(testCoreCluster), "1s").Should(MatchError("clusters.cluster.x-k8s.io \"test-ocp-infrastructure-name\" not found"))
+				By("ending IT here")
```
JoelSpeed, line 246: Debugging?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,242 @@
...
+	capiInfraClusterAPIVersionV1Beta1 = "infrastructure.cluster.x-k8s.io/v1beta1"
+	capiInfraClusterAPIVersionV1Beta2 = "infrastructure.cluster.x-k8s.io/v1beta2"
+	clusterOperatorName               = "cluster-api"
+)
+
+var (
+	errPlatformStatusShouldNotBeNil                         = errors.New("infrastructure PlatformStatus should not be nil")
+	errMappingOpenshiftPlatformToInfraClusterKindAndVersion = errors.New("failed to map Infrastructure PlatformType to a Cluster API infra cluster kind, platform not supported")
+	errOpenshiftInfraShouldNotBeNil                         = errors.New("infrastructure object should not be nil")
+	errOpenshiftInfrastructureNameShouldNotBeEmpty          = errors.New("infrastructure object's status.infrastructureName should not be empty")
```
JoelSpeed, line 55: There's inconsistencies in these messages, perhaps we should
```suggestion
	errPlatformStatusShouldNotBeNil                         = errors.New("infrastructure platformStatus should not be nil")
	errMappingOpenshiftPlatformToInfraClusterKindAndVersion = errors.New("failed to map infrastructure platformType to infrastructure cluster kind, platform not supported")
	errOpenshiftInfraShouldNotBeNil                         = errors.New("infrastructure object should not be nil")
	errOpenshiftInfrastructureNameShouldNotBeEmpty          = errors.New("infrastructure object's infrastructureName should not be empty")
```

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,242 @@
...
+
+// Reconcile reconciles the core cluster object for the openshift-cluster-api namespace.
+func (r *CoreClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithName(controllerName)
+	logger.Info("Reconciling core cluster")
+	defer logger.Info("Finished reconciling core cluster")
+
+	ocpInfrastructureName, err := getOCPInfrastructureName(r.Infra)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to obtain ocp infrastructure name: %w", err)
```
JoelSpeed, line 91: Lets drop the `ocp` from the logs

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,242 @@
...
+		return cluster, nil
+	}
+
+	if r.Infra.Status.PlatformStatus == nil {
+		return nil, errPlatformStatusShouldNotBeNil
+	}
+
+	infraClusterKind, infraClusterAPIVersion, err := mapOCPPlatformToInfraClusterKindAndVersion(r.Platform)
+	if err != nil {
+		return nil, fmt.Errorf("unable to map Infrastucture resource platform type to Cluster API infra cluster kind: %w", err)
```
JoelSpeed, line 133: Should be
```suggestion
		return nil, fmt.Errorf("unable to map infrastucture resource platform type to infrastructure cluster kind: %w", err)
```
We are duplicating a lot of the error message here, perhaps we don't need map in both places

Perhaps the inner error should just be `unsupported platform type` and be less detailed?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,242 @@
...
+
+	infraCluster := &unstructured.Unstructured{}
+	infraCluster.SetKind(infraClusterKind)
+	infraCluster.SetAPIVersion(infraClusterAPIVersion)
+
+	if err := r.Client.Get(ctx, clusterObjectKey, infraCluster); err != nil {
+		return nil, fmt.Errorf("failed to get infra cluster %s/%s: %w", clusterObjectKey.Namespace, clusterObjectKey.Name, err)
+	}
+
+	logger.Info(fmt.Sprintf("core cluster %s/%s does not exist, creating it", clusterObjectKey.Namespace, clusterObjectKey.Name))
```
JoelSpeed, line 144: Other info logs are being capitalised
```suggestion
	logger.Info(fmt.Sprintf("Core cluster %s/%s does not exist, creating it", clusterObjectKey.Namespace, clusterObjectKey.Name))
```

pkg/controllers/corecluster/corecluster_controller_test.go
damdo: Yup, ty

pkg/controllers/corecluster/corecluster_controller.go
damdo: Agree

damdo:
Thanks for the review @JoelSpeed feedback included.

pkg/controllers/corecluster/corecluster_controller.go
JoelSpeed: Not fixed

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -0,0 +1,242 @@
...
+		return cluster, nil
+	}
+
+	if r.Infra.Status.PlatformStatus == nil {
+		return nil, errPlatformStatusShouldNotBeNil
+	}
+
+	infraClusterKind, infraClusterAPIVersion, err := mapOCPPlatformToInfraClusterKindAndVersion(r.Platform)
+	if err != nil {
+		return nil, fmt.Errorf("unable to map infrastucture resource platform type to infrastructure cluster kind: %w", err)
```
JoelSpeed, line 133: Do we want to simplify/deduplicate this one?

JoelSpeed (commented):
Two bits from the last review appear to have been missed

damdo:
> Two bits from the last review appear to have been missed

Sorry I had pushed the wrong branch.
Should be fixed now :) Thanks!

PR #237, opened 2024-11-15
OCPCLOUD-2644: MAPI to CAPI MachineSet synchronization

damdo:
These change:

-> Implements MAPI to CAPI sync
-> Builds out test suite


pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -333,10 +383,24 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 				// We could also mock the conversion interface.
 			})
 
-			PContext("when the CAPI machine set does not exist", func() {
-				// The current test is a placeholder and should cover the functionality in,
-				// case authoritativeAPI == machinev1beta1.MachineAuthorityClusterAPI && capiMachineSet == nil,
-				// once that is implemented.
+			Context("when the CAPI machine set does not exist", func() {
+				It("should create the CAPI machine set", func() {
+					capiMs := capiv1resourcebuilder.MachineSet().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
```
theobarberbany, line 388: more a question than a nit, but wouldn't this make sense in a `BeforeEach` ? for this context?

theobarberbany (commented):
Overall, on first pass LGTM! :D 

Will wait on Joel review as well :)

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
damdo: Here we are not actually "creating it", we are building an object so that we can do k.Get() on it and check if the controller/reconciler created it for us :) Or am I missing your point here?

theobarberbany: ah yep I got that, my point was just we could still do it in a `BeforeEach()` and that makes it somewhat less surprising when reading through the code :)

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -415,3 +577,41 @@ func objectMetaIsEqual(a, b metav1.ObjectMeta) bool {
...
+}
+
+// getCAPIMSResourceVersion returns the CAPI machine set ResourceVersion or the zero value for it.
+func getCAPIMSResourceVersion(obj *capiv1beta1.MachineSet) string {
+	if obj == nil {
+		return "0"
+	}
+
+	return obj.GetResourceVersion()
+}
```
JoelSpeed, line 617: Why is this called out separately, the `getResourceVersion` should function the same no?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -202,8 +216,44 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
-				// Here we expect the controller to create the CAPI MachineSet from the MAPI one.
+			Context("when the CAPI machine set does not exist", func() {
+				It("should create the CAPI machine set", func() {
+					capiMs := capiv1resourcebuilder.MachineSet().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
+					Eventually(k.Get(capiMs)).Should(Succeed())
+				})
+
+				It("should update the synchronized condition on the MAPI machine set to True", func() {
+					Eventually(k.Object(mapiMachineSet), timeout).Should(
+						SatisfyAll(
```
JoelSpeed, line 227: Redundant SatisfyAll?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -413,37 +477,116 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
-								HaveField("Type", Equal(consts.SynchronizedCondition)),
-								HaveField("Status", Equal(corev1.ConditionFalse)),
-								HaveField("Severity", Equal(machinev1beta1.ConditionSeverityError)),
-								HaveField("Reason", Equal("FailedToGetCAPIInfraResources")),
-							))),
-					))
+			Context("when the CAPI machine set does not exist", func() {
+				It("should update the synchronized condition on the MAPI machine set to True", func() {
+					Eventually(k.Object(mapiMachineSet), timeout).Should(
+						SatisfyAll(
```
JoelSpeed, line 580: Redundant?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -413,37 +477,116 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
-					SatisfyAll(
-						HaveField("Status.Conditions", ContainElement(
-							SatisfyAll(
-								HaveField("Type", Equal(consts.SynchronizedCondition)),
-								HaveField("Status", Equal(corev1.ConditionFalse)),
-								HaveField("Severity", Equal(machinev1beta1.ConditionSeverityError)),
-								HaveField("Reason", Equal("FailedToGetCAPIInfraResources")),
-							))),
-					))
+			Context("when the CAPI machine set does not exist", func() {
```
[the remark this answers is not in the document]
damdo: Good catch

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -146,7 +145,26 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 		Expect(err).ToNot(HaveOccurred(), "Manager should be able to be created")
 
 		reconciler = &MachineSetSyncReconciler{
-			Client:        mgr.GetClient(),
+			Client: mgr.GetClient(),
+			Infra: &configv1.Infrastructure{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "cluster",
+				},
+				Spec: configv1.InfrastructureSpec{
```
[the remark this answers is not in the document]
damdo: We would need https://github.com/openshift/cluster-api-actuator-pkg/pull/348

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -234,14 +239,56 @@ func (r *MachineSetSyncReconciler) syncMachineSets(ctx context.Context, mapiMach
...
-	return ctrl.Result{}, nil
 }
 
 // reconcileMAPIMachineSetToCAPIMachineSet reconciles a MAPI MachineSet to a CAPI MachineSet.
 func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet) (ctrl.Result, error) {
-	// This function is currently a placeholder.
-	return ctrl.Result{}, nil
+	logger := log.FromContext(ctx)
+
+	newCAPIMachineSet, newCAPIInfraMachineTemplate, warns, err := r.convertMAPIToCAPIMachineSet(mapiMachineSet, r.Infra)
```
theobarberbany, line 248: Are we setting `r.Infra` anywhere? A quick cmd+f and I can't see where we would be, and this would panic otherwise.

theobarberbany: Also is there a reason this stores the OCP Infra on the reconciler struct, rather than using a helper func e.g `fetchOCPInfra`? (where the CAPI counterpart is `fetchCAPIInfraResources`? )

It feels like two different ways of doing things, one for CAPI and one for MAPI. Maybe we should standardise to one?

damdo: Yes we do se it in `cmd/machine-api-migration/main.go` of this PR

```
	machineSyncReconciler := machinesync.MachineSyncReconciler{
		Infra:    infra,
```

And it stores the OCP Infra object in the reconciler because we are already getting it in the main.go file to extract the provider from it, and so we can just plug it in that way and use it in the various controllers that require it.

This object is not expected to be changed while our controller is running. This though is not the case for the CAPI Infra resources, which will change during the controller normal runtime.

damdo: So considering they are different things and we have different expectations and uses for them, I would keep them like they are now.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -369,6 +426,111 @@ func (r *MachineSetSyncReconciler) updateSynchronizedConditionWithPatch(ctx cont
 	return nil
 }
 
+// createOrUpdateCAPIInfraMachineTemplate creates a CAPI infra machine template from a MAPI machine set, or updates if it exists and it is out of date.
+func (r *MachineSetSyncReconciler) createOrUpdateCAPIInfraMachineTemplate(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, infraMachineTemplate client.Object, newCAPIInfraMachineTemplate client.Object) (ctrl.Result, error) { //nolint:unparam
```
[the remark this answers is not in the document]
damdo: It is because in the returned `(ctrl.Result, error)`
we always come back with `return ctrl.Result{}, createErr` so the Result is always nil and golanci-lint complains about it.
I'm under the impression that every time we define a method for the reconciler receiver, it is wise to return `(ctrl.Result, error)` as the returning values. Hence I haven't removed `ctrl.Result` but put the no-lint instead.

pkg/controllers/machinesetsync/machineset_sync_controller.go
theobarberbany: > Yes we do se it in cmd/machine-api-migration/main.go of this PR

Ah I'd missed that - nice! 

> This object is not expected to be changed while our controller is running. This though is not the case for the CAPI Infra resources, which will change during the controller normal runtime.

Cool - that makes sense to me! :) Cheers for clearing up! 


After chatting on slack we're not too sure what the nil poniter deref the QE team have managed to create. Will need some digging in order to repro.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -413,37 +477,116 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+			})
+
+			Context("when the CAPI machine set does not exist", func() {
+				It("should create the CAPI machine set and CAPI infra machine template", func() {
+					capiMs := capiv1resourcebuilder.MachineSet().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
+					Eventually(k.Get(capiMs)).Should(Succeed())
+
+					capiInfraMachineTemplate := capav1builder.AWSMachineTemplate().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
+					Eventually(k.Get(capiInfraMachineTemplate)).Should(Succeed())
+				})
```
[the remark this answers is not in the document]
damdo: Thanks, added.

pkg/controllers/machinesetsync/machineset_sync_controller.go
damdo: That requires casting *capiv1beta1.MachineSet to client.Object() before passing that into getResourceVersion and also using reflect.ValueOf to check the inner value of the interface-wrapped type.
I had a go at it and it seems to work 👍

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -369,6 +426,111 @@ func (r *MachineSetSyncReconciler) updateSynchronizedConditionWithPatch(ctx cont
...
+		return ctrl.Result{}, updateErr
+	}
+
+	logger.Info("Successfully updated CAPI infra machine template")
+
+	return ctrl.Result{}, nil
+}
+
+// createOrUpdateCAPIMachineSet creates a CAPI machine set from a MAPI one, or updates if it exists and it is out of date.
+func (r *MachineSetSyncReconciler) createOrUpdateCAPIMachineSet(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet, newCAPIMachineSet *capiv1beta1.MachineSet) (ctrl.Result, error) { //nolint:unparam
```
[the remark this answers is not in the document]
damdo: see https://github.com/openshift/cluster-capi-operator/pull/237#discussion_r1871428526

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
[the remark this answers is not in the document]
damdo: 348 merged and I now bumped this PR to use that.

Makefile
```diff
@@ -40,7 +40,7 @@ migration:
 	go build -o bin/machine-api-migration cmd/machine-api-migration/main.go
 
 unit:
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin --index https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml)" ./hack/test.sh "./pkg/... ./assets/..." 5m
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin --index https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml)" ./hack/test.sh "./pkg/controllers/machinesetsync/..." 5m
```
JoelSpeed, line 43: Don't think this was meant to be committed?

damdo: Thanks for catching. Fixed.

JoelSpeed: Now it's looking at manifests gen and not assets? What are we expecting this to be?

damdo: Yes this was pointing to `assets/` which was the old name for the `manifests-gen`, when I refactored I forgot to repoint this, hence I fixed it now.

JoelSpeed:
/approve

Though I think @theobarberbany was looking into a panic from pre-merge testing, will let Theo handle the LGTM

cmd/machine-api-migration/main.go
```diff
@@ -230,6 +231,7 @@ func main() {
 
 	machineSetSyncReconciler := machinesetsync.MachineSetSyncReconciler{
 		Platform: provider,
+		Infra:    infra,
```
theobarberbany, line 234: The source of the panic 😂  cc/ @JoelSpeed

PR #242, opened 2024-12-04
OCPCLOUD-2787: Updates migration cmd to change metrics port

theobarberbany:
This change:

Removes the defaulting in the code where we set DiagnosticsAddress. It was being overwritten by --diagnostics-address in capiflags.AddManagerOptions() and setting the default value to :8443.

Updates the deployment manifests to pass --diagnostics-address to the migration controller with a value of :8442, so we don't clash with the cluster-capi-operator container running in the pod.

### DIFF
--- a/manifests/0000_30_cluster-api_11_deployment.yaml
+++ b/manifests/0000_30_cluster-api_11_deployment.yaml
@@ -31,12 +31,16 @@ spec:
         - ./cluster-capi-operator
         args:
           - --images-json=/etc/cluster-api-config-images/images.json
+          - --diagnostics-address=:8443
         env:
         - name: RELEASE_VERSION
           value: "0.0.1-snapshot"
         ports:
         - containerPort: 9443
           name: webhook-server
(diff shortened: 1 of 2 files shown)


theobarberbany:
/hold 

whilst i work out if we want to expose metrics with ports in the deployment yaml

manifests/0000_30_cluster-api_11_deployment.yaml
```diff
@@ -38,6 +38,9 @@ spec:
         - containerPort: 9443
           name: webhook-server
           protocol: TCP
+        - containerPort: 8443
```
JoelSpeed, line 41: This is the default? I wonder if we should be explicit anyway and add the flag?

theobarberbany: Yeah it's default. Can't hurt to be explicit, will update.

damdo (commented):
/approve
/lgtm

PR #243, opened 2024-12-09
OCPCLOUD-2787: Updates migration controller cmd

theobarberbany:
Fixes broken import of capa v1beta2, removes capav1beta1.


cmd/machine-api-migration/main.go
```diff
@@ -29,7 +29,8 @@ import (
 	"github.com/openshift/cluster-capi-operator/pkg/controllers/machinesetsync"
 	"github.com/openshift/cluster-capi-operator/pkg/controllers/machinesync"
 	"github.com/openshift/cluster-capi-operator/pkg/util"
-	capav1beta1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	capav1beta1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta1"
```
damdo, line 32: Do we need the v1beta1 anywhere?

theobarberbany: Hmm we currently use v1beta1 in the machine-sync controller `return &awscapiv1beta1.AWSMachine{}`, but if we can get the same from v1beta2, I don't see why we couldn't remove it..

theobarberbany: Ah that makes sense, I have updated :) 

> v1beta2, as that's what we work with everywhere,

Must have missed this on the machineset sync reviews before, oops

damdo (commented):
/approve
/lgtm

PR #246, opened 2024-12-13
Aggregate controllers conditions under ClusterOperator

damdo:
- standardize setting of controllers normal status conditions
- repurpose unsupported controller to a broader clusteroperator one

TODO:
- [ ] add proper aggregation logic
- [ ] figure out what to do with errors in the controllers WRT available=false or degraded=true (Do we normally want to only set available=false on errors and eventually check if we ended up in a loop of errors and set degraded=true after a while?)
- [ ] add the conditions setting for errors in the various controllers' reconcile loops


cmd/cluster-capi-operator/main.go
```diff
@@ -266,8 +266,9 @@ func setupPlatformReconcilers(mgr manager.Manager, infra *configv1.Infrastructur
 		azureCloudEnvironment := getAzureCloudEnvironment(infra.Status.PlatformStatus)
 		if azureCloudEnvironment == configv1.AzureStackCloud {
 			klog.Infof("Detected Azure Cloud Environment %q on platform %q is not supported, skipping capi controllers setup", azureCloudEnvironment, platform)
-			setupUnsupportedController(mgr, managedNamespace)
+			setupClusterOperatorController(mgr, managedNamespace, true)
 		} else {
+			// The ClusterOperator Controller must run in all cases.
```
nrb, line 298: I think this comment should be above the `setupClusterOperatorController`

cmd/cluster-capi-operator/main.go
```diff
@@ -298,24 +298,24 @@ func setupReconcilers(mgr manager.Manager, infra *configv1.Infrastructure, platf
 		Platform:                    platform,
 		Infra:                       infra,
 	}).SetupWithManager(mgr); err != nil {
-		klog.Error(err, "unable to create controller", "controller", "CoreCluster")
+		klog.Error(err, "unable to create core cluster controller", "controller", "CoreCluster")
```
nrb, line 329: I'm probably missing context, but each of the `klog.Error` calls here repeat the controller's name and `controller`, which seems redundant. Do we have a special handler installed? I don't see it in this file.

cmd/cluster-capi-operator/main.go
```diff
@@ -109,6 +109,11 @@ func main() {
 		controllers.DefaultManagedNamespace,
 		"The namespace where CAPI components will run.",
 	)
+	sourceNamespace := flag.String(
+		"source-namespace",
+		controllers.DefaultMAPIManagedNamespace,
+		"The namespace where MAPI components will run.",
```
nrb, line 123: The code appears to only use it for secrets; are we actually getting other MAPI components with it? If not, the variable name could be a little more descriptive up here.

pkg/controllers/infracluster/infracluster_controller.go
```diff
@@ -227,26 +218,49 @@ func (r *InfraClusterController) ensureInfraCluster(ctx context.Context, log log
...
+	if err := r.SyncStatus(ctx, co, conds); err != nil {
+		return fmt.Errorf("failed to sync cluster operator status: %w", err)
+	}
+
+	return nil
+}
+
+// setControllerConditionDegraded sets the InfraClusterController conditions to a degraded state.
+//
+//nolint:unused
```
nrb, line 246: I'm reviewing commit-wise, so this might be used later in the PR, but if not, we should delete it and add it when it's actually called.

pkg/controllers/clusteroperator/clusteroperator_controller_test.go
```diff
@@ -68,62 +71,295 @@ var _ = Describe("ClusterOperator controller", func() {
...
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.MachineSyncControllerAvailableCondition, Status: configv1.ConditionTrue, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.MachineSyncControllerDegradedCondition, Status: configv1.ConditionFalse, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.KubeconfigControllerAvailableCondition, Status: configv1.ConditionTrue, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.KubeconfigControllerDegradedCondition, Status: configv1.ConditionFalse, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.CapiInstallerControllerAvailableCondition, Status: configv1.ConditionTrue, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.CapiInstallerControllerDegradedCondition, Status: configv1.ConditionTrue, Reason: "Error", Message: "reconciler error"},
```
nrb, line 228: Maybe put a comment at the end of this line to indicate it's the one that's degraded, since these lines are all so similar. Otherwise it's commented 24 lines away.

pkg/controllers/clusteroperator/clusteroperator_controller_test.go
```diff
@@ -68,62 +71,295 @@ var _ = Describe("ClusterOperator controller", func() {
...
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.MachineSyncControllerDegradedCondition, Status: configv1.ConditionFalse, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.KubeconfigControllerAvailableCondition, Status: configv1.ConditionTrue, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.KubeconfigControllerDegradedCondition, Status: configv1.ConditionFalse, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.CapiInstallerControllerAvailableCondition, Status: configv1.ConditionTrue, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.CapiInstallerControllerDegradedCondition, Status: configv1.ConditionFalse, Reason: operatorstatus.ReasonAsExpected},
+				{LastTransitionTime: metav1.Now(), Type: operatorstatus.SecretSyncControllerAvailableCondition, Status: configv1.ConditionFalse, Reason: "Error", Message: "persistent reconcile error"},
```
nrb, line 273: Similar here, perhaps mark the line that's expected to trigger the desired test state

nrb (commented):
Overall, it looks good. Thanks for the detailed tests!

I would strongly suggest removing any function that's got `no-lint: unused`, even though it might be useful in the future. https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

cmd/cluster-capi-operator/main.go
```diff
@@ -243,79 +268,88 @@ func main() {
...
 	case configv1.GCPPlatformType:
-		setupReconcilers(mgr, infra, platform, &gcpv1.GCPCluster{}, containerImages, applyClient, apiextensionsClient, managedNamespace)
+		setupReconcilers(mgr, infra, platform, &gcpv1.GCPCluster{}, containerImages, applyClient, apiextensionsClient, currentFeatureGates, managedNamespace, sourceNamespace)
 		setupWebhooks(mgr)
 	case configv1.AzurePlatformType:
 		azureCloudEnvironment := getAzureCloudEnvironment(infra.Status.PlatformStatus)
 		if azureCloudEnvironment == configv1.AzureStackCloud {
 			klog.Infof("Detected Azure Cloud Environment %q on platform %q is not supported, skipping capi controllers setup", azureCloudEnvironment, platform)
-			setupUnsupportedController(mgr, managedNamespace)
+			setupClusterOperatorController(mgr, managedNamespace, currentFeatureGates, true)
```
JoelSpeed, line 296: Would it make sense to set up the clusteroperator controller before the switch, and remove it from `setupReconcilers`? Then we don't need this else, or the default case down below?

cmd/cluster-capi-operator/main.go
```diff
@@ -243,79 +268,88 @@ func main() {
...
 	}
 }
 
-func setupReconcilers(mgr manager.Manager, infra *configv1.Infrastructure, platform configv1.PlatformType, infraClusterObject client.Object, containerImages map[string]string, applyClient *kubernetes.Clientset, apiextensionsClient *apiextensionsclient.Clientset, managedNamespace string) {
-	if err := (&corecluster.CoreClusterReconciler{
-		ClusterOperatorStatusClient: getClusterOperatorStatusClient(mgr, "cluster-capi-operator-cluster-resource-controller", managedNamespace),
+func setupReconcilers(mgr manager.Manager, infra *configv1.Infrastructure, platform configv1.PlatformType, infraClusterObject client.Object, containerImages map[string]string,
+	applyClient *kubernetes.Clientset, apiextensionsClient *apiextensionsclient.Clientset, currentFeatureGates featuregates.FeatureGate, managedNamespace, sourceNamespace string) {
+	// The ClusterOperator Controller must run under all circumstances as it manages the ClusterOperator object for this operator.
+	setupClusterOperatorController(mgr, managedNamespace, currentFeatureGates, false)
```
JoelSpeed, line 321: Yeah, as with previous comment, maybe move this up a level and before the switch to avoid calling it in three different places if it must always run

cmd/cluster-capi-operator/main.go
JoelSpeed: It's a structured logging format here, so `controller` is adding a new key`, and the next string is the value. The logging through this function makes sense given the structured logging context

cmd/cluster-capi-operator/main.go
```diff
@@ -383,14 +406,48 @@ func getAzureCloudEnvironment(ps *configv1.PlatformStatus) configv1.AzureCloudEn
...
 		IsUnsupportedPlatform:       isUnsupportedPlatform,
 	}).SetupWithManager(mgr); err != nil {
 		klog.Error(err, "unable to create clusteroperator controller", "controller", "ClusterOperator")
 		os.Exit(1)
 	}
 }
+
+// getFeatureGates is used to fetch the current feature gates from the cluster.
+// We use this to check if the machine api migration is actually enabled or not.
+func getFeatureGates(mgr ctrl.Manager) (featuregates.FeatureGateAccess, error) {
```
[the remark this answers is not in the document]
JoelSpeed: Perhaps we should 🤔

cmd/cluster-capi-operator/main.go
```diff
@@ -372,13 +406,48 @@ func getAzureCloudEnvironment(ps *configv1.PlatformStatus) configv1.AzureCloudEn
...
+	desiredVersion := util.GetReleaseVersion()
+	missingVersion := "0.0.1-snapshot"
+
+	configClient, err := configv1client.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return nil, fmt.Errorf("failed to create config client: %w", err)
+	}
+
+	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
```
JoelSpeed, line 432: Minor: This probably doesn't need to resync so quickly

cmd/machine-api-migration/main.go
```diff
@@ -216,29 +219,34 @@ func main() {
 		os.Exit(1)
 	}
 
-	machineSyncReconciler := machinesync.MachineSyncReconciler{
+	// ClusterOperator watches and keeps the cluster-api ClusterObject up to date.
+	if err := (&clusteroperator.ClusterOperatorController{
```
JoelSpeed, line 223: Is this the same controller as the other manager? Is it safe to have it running twice?

pkg/controllers/clusteroperator/clusteroperator_controller.go
```diff
@@ -24,37 +24,47 @@ import (
...
 
-// UnsupportedController on unsupported platforms watches and keeps the cluster-api ClusterObject up to date.
-type UnsupportedController struct {
+// ClusterOperatorController watches and keeps the cluster-api ClusterObject up to date.
+type ClusterOperatorController struct {
 	operatorstatus.ClusterOperatorStatusClient
-	Scheme *runtime.Scheme
+	Scheme                *runtime.Scheme
+	IsUnsupportedPlatform bool
+	FeatureGates          featuregates.FeatureGate
```
JoelSpeed, line 44: Where are the gates used?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -215,6 +219,52 @@ func (r *CoreClusterReconciler) ensureCoreClusterControlPlaneInitializedConditio
...
+// setControllerConditionsToNormal sets the CoreClusterController conditions to the normal state.
+func (r *CoreClusterController) setControllerConditionsToNormal(ctx context.Context, log logr.Logger) error {
+	co, err := r.GetOrCreateClusterOperator(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster operator: %w", err)
+	}
+
+	conds := []configv1.ClusterOperatorStatusCondition{
+		operatorstatus.NewClusterOperatorStatusCondition(operatorstatus.CoreClusterControllerAvailableCondition, configv1.ConditionTrue, operatorstatus.ReasonAsExpected,
+			"Core Cluster Controller works as expected"),
```
JoelSpeed, line 231: What if it's not available? AFAICT there's no way for this controller to say, oops, I have an error, I'm not available.

Is there a reason, after initial setup, where going available false is valid?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -215,6 +219,52 @@ func (r *CoreClusterReconciler) ensureCoreClusterControlPlaneInitializedConditio
...
+		operatorstatus.NewClusterOperatorStatusCondition(operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionTrue, operatorstatus.ReasonAsExpected,
+			fmt.Sprintf("Core Cluster Controller is degraded: %s", reconcileErr.Error())),
```
JoelSpeed, line 256: If on the first ever reconcile in a cluster, we have an error, and set ourselves degraded, should this really be Available=True?

I'm wondering if managing the available condition separately would be a better way to handle this?

pkg/controllers/infracluster/infracluster_controller.go
JoelSpeed: Yeah, if this isn't used, lets dump it

pkg/controllers/kubeconfig/kubeconfig.go
```diff
@@ -198,3 +188,51 @@ func newKubeConfigSecret(clusterName string, data []byte) *corev1.Secret {
...
+	if err := r.SyncStatus(ctx, co, conds); err != nil {
+		return fmt.Errorf("failed to sync cluster operator status: %w", err)
+	}
+
+	return nil
+}
+
+// setControllerConditionDegraded sets the KubeconfigController conditions to the normal state.
+//
+//nolint:unused
```
JoelSpeed, line 217: If there's no way for this controller to be degraded, lets get rid of this

pkg/controllers/kubeconfig/kubeconfig_test.go
```diff
@@ -26,94 +27,185 @@ import (
...
+		return mgrCancel, mgrDone
+	}
+
+	stopManager := func() {
+		By("Stopping the manager")
+		mgrCancel()
+		// Wait for the mgrDone to be closed, which will happen once the mgr has stopped.
+		<-mgrDone
+
+		Eventually(mgrDone).Should(BeClosed())
```
JoelSpeed, line 89: Aren't these doing the same thing?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -642,3 +663,51 @@ func getResourceVersion(obj client.Object) string {
...
+	if err := r.SyncStatus(ctx, co, conds); err != nil {
+		return fmt.Errorf("failed to sync cluster operator status: %w", err)
+	}
+
+	return nil
+}
+
+// setControllerConditionDegraded sets the MachineSetSyncController conditions to a degraded state.
+//
+//nolint:unused
```
JoelSpeed, line 692: If this isn't used, drop it

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -244,3 +277,51 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+	if err := r.SyncStatus(ctx, co, conds); err != nil {
+		return fmt.Errorf("failed to sync cluster operator status: %w", err)
+	}
+
+	return nil
+}
+
+// setControllerConditionDegraded sets the MachineSyncController conditions to a degraded state.
+//
+//nolint:unused
```
JoelSpeed, line 306: If it isn't used, drop it

pkg/controllers/secretsync/secret_sync_controller.go
```diff
@@ -176,59 +158,59 @@ func (r *UserDataSecretController) SetupWithManager(mgr ctrl.Manager) error {
...
+		return fmt.Errorf("failed to sync cluster operator status: %w", err)
 	}
 
 	return nil
 }
 
-func (r *UserDataSecretController) setDegradedCondition(ctx context.Context, log logr.Logger) error {
+// setControllerConditionDegraded sets the UserDataSecretController conditions to a degraded state.
+//
+//nolint:unused
```
JoelSpeed, line 196: If it's not used, drop it

Looks like perhaps previously it was used? So why do we no longer have a degraded state for this controller?

pkg/controllers/clusteroperator/clusteroperator_controller_test.go
```diff
@@ -0,0 +1,349 @@
...
+	return mgrCancel, mgrDone
+}
+
+func stopManager() {
+	By("Stopping the manager")
+	mgrCancel()
+	// Wait for the mgrDone to be closed, which will happen once the mgr has stopped.
+	<-mgrDone
+
+	Eventually(mgrDone).Should(BeClosed())
```
JoelSpeed, line 348: These appear to do the same thing

pkg/operatorstatus/operator_status.go
```diff
@@ -46,6 +46,54 @@ const (
...
 	ReasonSyncFailed = "SyncingFailed"
+
+	// ReasonControllersNotAvailable is the reason for the condition when one or more of the operator's controllers are not available.
+	ReasonControllersNotAvailable = "ControllersNotAvailable"
+
+	// ReasonControllersDegraded is the reason for the condition when one or more of the operator's controllers are degraded.
+	ReasonControllersDegraded = "ControllersDegraded"
+
+	// CapiInstallerControllerAvailableCondition is the condition type that indicates the CapiInstaller controller is available.
+	CapiInstallerControllerAvailableCondition = "CapiInstallerControllerAvailable"
```
JoelSpeed, line 57: `s/Capi/CAPI/` IMO

pkg/operatorstatus/operator_status.go
```diff
@@ -198,13 +200,133 @@ func NewClusterOperatorStatusCondition(conditionType configv1.ClusterStatusCondi
...
+
+	// Evaluate each controller's conditions.
+	for _, conditions := range controllerConditions {
+		evaluateCondition(co, conditions.available, &anyAvailableMissing, &availableMsg, false)
+		evaluateCondition(co, conditions.degraded, &anyDegradedMissing, &degradedMsg, true)
+	}
+
+	availableCondition := newAvailableCondition(anyAvailableMissing, availableMsg, releaseVersion)
+	degradedCondition := newDegradedCondition(anyDegradedMissing, degradedMsg)
+	progressingCondition := newCondition(configv1.OperatorProgressing, configv1.ConditionFalse, "", "")
```
JoelSpeed, line 244: At some point the CO should be progressing, during an upgrade of operands, I wonder why this isn't represented here?

pkg/operatorstatus/operator_status.go
```diff
@@ -198,13 +200,133 @@ func NewClusterOperatorStatusCondition(conditionType configv1.ClusterStatusCondi
...
+		degraded  configv1.ClusterStatusConditionType
+	}
+
+	// Define the controller conditions
+	controllerConditions := []controllerCondition{
+		{CapiInstallerControllerAvailableCondition, CapiInstallerControllerDegradedCondition},
+		{CoreClusterControllerAvailableCondition, CoreClusterControllerDegradedCondition},
+		{InfraClusterControllerAvailableCondition, InfraClusterControllerDegradedCondition},
+		{KubeconfigControllerAvailableCondition, KubeconfigControllerDegradedCondition},
+		{SecretSyncControllerAvailableCondition, SecretSyncControllerDegradedCondition},
```
JoelSpeed, line 224: Since not all of these have any reason to be degraded, do we require them all to always set a degraded condition?

Would it be better to factor this some way so that we only actually have conditions that are likely to change over time?

pkg/operatorstatus/operator_status.go
```diff
@@ -198,13 +200,133 @@ func NewClusterOperatorStatusCondition(conditionType configv1.ClusterStatusCondi
...
+	progressingCondition := newCondition(configv1.OperatorProgressing, configv1.ConditionFalse, "", "")
+
+	upgradeableStatus := configv1.ConditionUnknown
+
+	switch {
+	case availableCondition.Status == "True" && degradedCondition.Status == "False":
+		upgradeableStatus = configv1.ConditionTrue
+	case availableCondition.Status == "False" || degradedCondition.Status == "True":
+		upgradeableStatus = configv1.ConditionFalse
+	}
```
JoelSpeed, line 253: Is this based on some document explaining how the conditions should be working?

pkg/operatorstatus/operator_status.go
```diff
@@ -198,13 +200,133 @@ func NewClusterOperatorStatusCondition(conditionType configv1.ClusterStatusCondi
...
+func newAvailableCondition(anyMissing bool, messages []string, releaseVersion string) configv1.ClusterOperatorStatusCondition {
+	if anyMissing {
+		return newCondition(configv1.OperatorAvailable, configv1.ConditionUnknown, "", "")
+	}
+
+	if len(messages) == 0 {
+		return newCondition(configv1.OperatorAvailable, configv1.ConditionTrue, ReasonAsExpected, fmt.Sprintf("Cluster CAPI Operator is available at %s", releaseVersion))
+	}
+
+	return newCondition(configv1.OperatorAvailable, configv1.ConditionFalse, "ControllersNotAvailable", fmt.Sprintf("The following controllers available conditions are not as expected: %s", strings.Join(messages, ", ")))
```
JoelSpeed, line 294: As far as I could tell from reviewing the other controllers, we will never report available false, which I think is probably an error right?

pkg/operatorstatus/operator_status.go
```diff
@@ -198,13 +200,133 @@ func NewClusterOperatorStatusCondition(conditionType configv1.ClusterStatusCondi
...
+}
+
+// newCondition helps to create a new condition with optional message.
+func newCondition(conditionType configv1.ClusterStatusConditionType, status configv1.ConditionStatus, reason, message string) configv1.ClusterOperatorStatusCondition {
+	return configv1.ClusterOperatorStatusCondition{
+		Type:               conditionType,
+		Status:             status,
+		Reason:             reason,
+		Message:            message,
+		LastTransitionTime: metav1.Now(),
```
JoelSpeed, line 317: Does the code applying these take care of making sure `LastTransitionTime` doesn't tick up on every reconcile?

JoelSpeed (commented):
Reading through this, it appears that several of our controller can never go degraded, is that really true?

I'm also not sure I like the way we are tying the available and degraded conditions together. It appears that there is no way for us to ever say that we are not available? I would expect that after cluster bootstrap, that's probably true, but, if we are degraded=true on first reconcile, we also claim we are available, which is likely misleading.

I would expect a flow through the controller that mimics what CAPI does to be more accurate

Eg

```
1. Fetch clusteroperator at top of Reconcile
2. Extract conditions owned by our SSA manager name
3. Set up defer to update (via SSA) the conditions on the clusteroperator
4. If Available condition is not set, set it to false
5. As we traverse the function, update the list of extracted conditions with individual and specific conditions
6. Only once we know we are available, set available true
```
With this, we would have an accurate view of the available condition, and the degraded condition could be handled separately where appropriate.

It would also make it easier in the future to add upgradeable into the mix for the controllers if and when we need that

cmd/cluster-capi-operator/main.go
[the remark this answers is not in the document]
damdo: I thought the same when touching this function. And I agree :D

damdo:
> Reading through this, it appears that several of our controller can never go degraded, is that really true?
I'm also not sure I like the way we are tying the available and degraded conditions together. It appears that there is no way for us to ever say that we are not available? I would expect that after cluster bootstrap, that's probably true, but, if we are degraded=true on first reconcile, we also claim we are available, which is likely misleading.

I agree with your points here. In fact this was my goal when writing this code. It was to put in place the logic for setting of controller level conditions and aggregation for them at the operator level (this is why for example I temporarily left in `setControllerConditionDegraded` even if they are not used ATM).

I didn't want to necessarily decide myself where we should be degraded, available, or their counterparts in each controller.
That's why, for example, I didn't set degraded or unavailable conditions in each controller, but only the happy ones.
I would like to have an open discussion on where to set which conditions in each controller.

I'm also happy to pull in folks from the CVO/upgrades team if we want to get their opinion on what's the best posture in terms of conditions at various moments of the operator/cluster lifecycle.

damdo:
> Reading through this, it appears that several of our controller can never go degraded, is that really true?
I'm also not sure I like the way we are tying the available and degraded conditions together. It appears that there is no way for us to ever say that we are not available? I would expect that after cluster bootstrap, that's probably true, but, if we are degraded=true on first reconcile, we also claim we are available, which is likely misleading.

@JoelSpeed I agree with your points here. In fact this was my goal when writing this code. It was to put in place the logic for setting of controller level conditions and aggregation for them at the operator level (this is why for example I temporarily left in `setControllerConditionDegraded` even if they are not used ATM).

I didn't want to necessarily decide myself where we should be degraded, available, or their counterparts in each controller.
That's why, for example, I didn't set degraded or unavailable conditions in each controller, but only the happy ones.
I would like to have an open discussion on where to set which conditions in each controller.

I'm also happy to pull in folks from the CVO/upgrades team if we want to get their opinion on what's the best posture in terms of conditions at various moments of the operator/cluster lifecycle.

PR #248, opened 2025-01-10
OCPBUGS-29815: fix: always update clusteroperator status versions when differing

damdo:
For manual testing of this fix, see: https://github.com/openshift/cluster-capi-operator/pull/248#issuecomment-2592344932

**Problem:**

Prior to this change, the operator's controller responsible for keeping the ClusterObject up to date had an issue when invoking r.SetStatusAvailable(...) with a non-empty availableConditionMsg string. Specifically:

- The condition NewClusterOperatorStatusCondition(configv1.OperatorAvailable, configv1.ConditionTrue, ReasonAsExpected, availableConditionMsg) would remain unchanged.
- This left the conds []configv1.ClusterOperatorStatusCondition unmodified compared to the current ClusterObject.Status.Conditions. 
- As a result, the ClusterObject.Status was not resynced.
 
This behavior caused a critical issue during cluster version upgrades:
- The operator failed to propagate changes to ClusterObject.Status.Versions, as the logic did not detect any updates requiring a resync.
- This prevented the cluster version upgrade process from progressing, effectively blocking the upgrade, since the Cluster Version Operator awaited the cluster-api operator to sync the new ClusterObject.Status.Versions.

**Solution:**

With this change, the criteria for resyncing the ClusterObject.Status have been updated. The resync logic now also triggers when only the desired operandVersions are modified, ensuring that:
- The ClusterObject.Status is correctly updated even if no other conditions change.
- Cluster version upgrades can proceed without unnecessary blockages.

### DIFF
--- a/pkg/controllers/unsupported/unsupported_controller_test.go
+++ b/pkg/controllers/unsupported/unsupported_controller_test.go
@@ -24,37 +24,48 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-api-actuator-pkg/testutils"
+	configv1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/config/v1"
+	corev1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/core/v1"
+	"github.com/openshift/cluster-capi-operator/pkg/controllers"
 	"github.com/openshift/cluster-capi-operator/pkg/operatorstatus"
-	"github.com/openshift/cluster-capi-operator/pkg/test"
 )
 
 var _ = Describe("CAPI unsupported controller", func() {
 	ctx := context.Background()
+	desiredOperatorReleaseVersion := "this-is-the-desired-release-version"
 	var r *UnsupportedController
 	var capiClusterOperator *configv1.ClusterOperator
-	capiClusterOperatorKey := client.ObjectKey{Name: "cluster-api"}
+	var testNamespaceName string
 
 	BeforeEach(func() {
 		r = &UnsupportedController{
 			ClusterOperatorStatusClient: operatorstatus.ClusterOperatorStatusClient{
-				Client: cl,
+				Client:         cl,
+				ReleaseVersion: desiredOperatorReleaseVersion,
 			},
 		}
 
+		By("Creating the cluster-api ClusterOperator")
 		capiClusterOperator = &configv1.ClusterOperator{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "cluster-api",
 			},
 		}
-
 		Expect(cl.Create(ctx, capiClusterOperator)).To(Succeed(), "should be able to create the 'cluster-api' ClusterOperator object")
+
+		By("Creating the testing namespace")
+		namespace := corev1resourcebuilder.Namespace().WithGenerateName("test-capi-corecluster-").Build()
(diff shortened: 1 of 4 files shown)


pkg/operatorstatus/operator_status.go
```diff
@@ -214,3 +206,25 @@ func printOperandVersions(versions []configv1.OperandVersion) string {
 
 	return strings.Join(versionsOutput, ", ")
 }
+
+func clusterObjectNeedsUpdating(co *configv1.ClusterOperator, conds []configv1.ClusterOperatorStatusCondition, desiredVersions []configv1.OperandVersion, desiredRelatedObjects []configv1.ObjectReference) (*configv1.ClusterOperator, bool) {
```
nrb, line 210: It looks like we only use the `.Status` field on `co`. Could we switch to passing that in?

damdo: We might want to update the status based on the values on spec, I think it is worth keeping this way

pkg/controllers/unsupported/unsupported_controller_test.go
```diff
@@ -63,19 +73,35 @@ var _ = Describe("CAPI unsupported controller", func() {
...
 				ContainElement(And(HaveField("Type", Equal(configv1.OperatorAvailable)), HaveField("Status", Equal(configv1.ConditionTrue)), HaveField("Message", Equal(capiUnsupportedPlatformMsg)))),
 				ContainElement(And(HaveField("Type", Equal(configv1.OperatorProgressing)), HaveField("Status", Equal(configv1.ConditionFalse)))),
 				ContainElement(And(HaveField("Type", Equal(configv1.OperatorDegraded)), HaveField("Status", Equal(configv1.ConditionFalse)))),
+				ContainElement(And(HaveField("Type", Equal(configv1.OperatorUpgradeable)), HaveField("Status", Equal(configv1.ConditionTrue)))),
 			),
 		), "should match the expected ClusterOperator status conditions")
+
+	})
+
+	It("should update the ClusterOperator status version to the desired one", func() {
```
JoelSpeed, line 90: Do we want/have a test that changes the value to something incorrect, and then see if it gets update back to the correct thing?

JoelSpeed (commented):
LGTM apart from a potential test update

pkg/controllers/unsupported/unsupported_controller_test.go
JoelSpeed: Yep, exactly

damdo: Ah yes, we don't have that at the moment. Updated.
Also something to take into account is that this controller is going to disappear in the next PR #246 as it is going to get merged into a different one.
For that one I am planning more exhaustive integration tests and them starting from the manager rather than invoking the r.Reconcile() directly.
So any extra testing changes I'd suggest we could do there.

damdo:
Conducted manual testing doing a micro-upgrade on 4.17 azure and verified that this fix works.

Started with a clusterbot cluster:
```
launch 4.17.0-0.nightly-2025-01-11-012032 techpreview,azure
```


Built a payload with the fix:
```
build 4.17.0-0.nightly-2025-01-11-095851,https://github.com/openshift/cluster-capi-operator/pull/250
# this produced the following image: registry.build05.ci.openshift.org/ci-ln-40m7t02/release:latest
```

Triggered an upgrade from the live cluster into the fixed version
```
[~] $ oc adm upgrade --to-image="registry.build05.ci.openshift.org/ci-ln-40m7t02/release:latest" --allow-explicit-upgrade --force
```

The upgrade finished successfully:
```
[~] $ oc adm upgrade                                                                                                                                                                                                                                                                                      

Cluster version is 4.17.0-0.test-2025-01-14-164031-ci-ln-40m7t02-latest

Upgradeable=False

  Reason: FeatureGates_RestrictedFeatureGates_TechPreviewNoUpgrade
  Message: Cluster operator config-operator should not be upgraded between minor versions: FeatureGatesUpgradeable: "TechPreviewNoUpgrade" does not allow updates

warning: Cannot display available updates:
  Reason: NoChannel
  Message: The update channel has not been configured.
```

damdo:
This bugfix was manually tested by myself, automatically tested by integration tests and is also behind tech preview feature
Hence labelling it with

damdo:
The serial failure is unrelated and the operator is not even running in that test.

PR #254, opened 2025-01-29
OCPCLOUD-2792: Repurpose unsupported controller to a broader clusteroperator one

damdo:
Repurpose unsupported controller to a broader clusteroperator one.
- remove the usupported controller
- create the new clusteroperator controller
- always run the new clusteroperator controller and behave differently depending on whether the controller is running on a supported vs unsupported platform
- add improved testing on the supported vs unsupported situation, starting a full manager rather than solely running the reconcile function

This is the first step of a series of PRs meant to work towards a unified controller responsible for aggregating all controllers conditions into centrailsed Available, Progressing, Upgradeable, and Degraded operator conditions under the ClusterOperator object.


pkg/controllers/clusteroperator/clusteroperator_controller_test.go
```diff
@@ -0,0 +1,203 @@
...
+	return mgrCancel, mgrDone
+}
+
+func stopManager() {
+	By("Stopping the manager")
+	mgrCancel()
+	// Wait for the mgrDone to be closed, which will happen once the mgr has stopped.
+	<-mgrDone
+
+	Eventually(mgrDone).Should(BeClosed())
```
JoelSpeed, line 202: Nit, I think these will do the same thing no? So, do we need line 200?

JoelSpeed (commented):
One small thing but otherwise makes sense

pkg/controllers/clusteroperator/clusteroperator_controller_test.go
damdo: Yeah you are right, good point. Updated.

JoelSpeed (commented):
/lgtm
/approve

damdo:
This change was tested via units and E2Es, and is in TPNU/CNU, hence adding the label for allowing it to merge.

PR #255, opened 2025-01-30
NO-JIRA: Remove unused Dockerfile

stephenfin:
ART uses the `Dockerfile.rhel` file [1]. You cannot use the `Dockerfile` file currently due to missing libraries.

[1] https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.18/images/ose-cluster-capi-operator.yml


stephenfin:
fwiw, these are the errors I see if I build and deploy using the `Dockerfile`:

```
❯ oc logs -n openshift-cluster-api cluster-capi-operator-f784f8546-mhpc9
Defaulted container "cluster-capi-operator" out of: cluster-capi-operator, machine-api-migration
./cluster-capi-operator: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./cluster-capi-operator)
./cluster-capi-operator: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./cluster-capi-operator)
```

This image was built with:

```
❯ podman build -t ghcr.io/stephenfin/cluster-capi-operator . -f Dockerfile
```

damdo (commented):
/retest

/lgtm

@JoelSpeed IIRC this is not in use and safe to remove.

PR #256, opened 2025-02-03
OCPCLOUD-2792: corecluster: set controller level conditions + tests

damdo:
This PR is part of a wider effort to try and rework the controller/operator level condition setting and aggregation for the cluster-capi-operator (see: https://issues.redhat.com/browse/OCPCLOUD-2792)

This PR specifically, builds on top of #254 and:
- defines controller level conditions for the CoreCluster controller, setting them accordingly
- tests that the conditions have been set under different scenarios

More PRs will follow for each controller once the approach proposed in this one is agreed upon.

### DIFF
--- a/pkg/controllers/corecluster/corecluster_controller.go
+++ b/pkg/controllers/corecluster/corecluster_controller.go
@@ -86,33 +86,60 @@ func (r *CoreClusterController) Reconcile(ctx context.Context, req reconcile.Req
 	logger.Info("Reconciling core cluster")
 	defer logger.Info("Finished reconciling core cluster")
 
+	var err error
+
+	conditions := &[]configv1.ClusterOperatorStatusCondition{}
+
+	// We optimistically set Degraded=false at the beginning.
+	operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionFalse,
+		operatorstatus.ReasonAsExpected, "controller works as expected")
+
+	// We optimistically set Available=true at the beginning.
+	operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerAvailableCondition, configv1.ConditionTrue,
+		operatorstatus.ReasonAsExpected, "controller is available")
+
+	defer func() {
+		// This function runs before every return, to ensure the controller conditions
+		// for that codepath are correctly synced to the ClusterOperator status.
+		if syncErr := r.syncControllerConditions(ctx, conditions); syncErr != nil {
+			if err != nil {
+				// The sync of the controller conditions failed but there is also a non nil reconcile error.
+				// Log the sync error and pass down the reconciler one.
+				logger.Error(syncErr, "failed to sync cluster operator status")
+			} else {
+				err = syncErr
+			}
+		}
+	}()
+
 	ocpInfrastructureName, err := getOCPInfrastructureName(r.Infra)
 	if err != nil {
+		operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionTrue,
+			operatorstatus.ReasonSyncFailed, fmt.Sprintf("controller is degraded: %s", err.Error()))
+
 		return ctrl.Result{}, fmt.Errorf("failed to obtain infrastructure name: %w", err)
 	}
 
 	cluster, err := r.ensureCoreCluster(ctx, client.ObjectKey{Namespace: r.ManagedNamespace, Name: ocpInfrastructureName}, logger)
 	if err != nil {
+		operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionTrue,
+			operatorstatus.ReasonSyncFailed, fmt.Sprintf("controller is degraded: %s", err.Error()))
+
 		return ctrl.Result{}, fmt.Errorf("failed to ensure core cluster: %w", err)
 	}
 
 	if !cluster.DeletionTimestamp.IsZero() {
-		if err := r.SetStatusAvailable(ctx, ""); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
-		}
-
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, err
 	}
 
 	if err := r.ensureCoreClusterControlPlaneInitializedCondition(ctx, cluster); err != nil {
+		operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionTrue,
+			operatorstatus.ReasonSyncFailed, fmt.Sprintf("controller is degraded: %s", err.Error()))
+
 		return ctrl.Result{}, fmt.Errorf("failed to ensure core cluster has the ControlPlaneInitializedCondition: %w", err)
 	}
 
-	if err := r.SetStatusAvailable(ctx, ""); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
-	}
-
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
 }
 
 // ensureCoreCluster creates a cluster with the given name and returns the cluster object.
@@ -239,4 +266,18 @@ func getOCPInfrastructureName(infra *configv1.Infrastructure) (string, error) {
 	}
 
 	return infra.Status.InfrastructureName, nil
+}
+
+// syncControllerConditions sets the CoreClusterController conditions.
+func (r *CoreClusterController) syncControllerConditions(ctx context.Context, conds *[]configv1.ClusterOperatorStatusCondition) error {
+	co, err := r.GetOrCreateClusterOperator(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster operator: %w", err)
+	}
+
+	if err := r.SyncStatus(ctx, co, *conds); err != nil {
+		return fmt.Errorf("failed to sync cluster operator status: %w", err)
+	}
+
+	return nil
 }
--- a/pkg/controllers/corecluster/corecluster_controller.go
+++ b/pkg/controllers/corecluster/corecluster_controller.go
@@ -46,6 +46,12 @@ const (
 	capiInfraClusterAPIVersionV1Beta1 = "infrastructure.cluster.x-k8s.io/v1beta1"
 	capiInfraClusterAPIVersionV1Beta2 = "infrastructure.cluster.x-k8s.io/v1beta2"
 	clusterOperatorName               = "cluster-api"
+
+	// CoreClusterControllerAvailableCondition is the condition type that indicates the CoreCluster controller is available.
+	CoreClusterControllerAvailableCondition = "CoreClusterControllerAvailable"
+
+	// CoreClusterControllerDegradedCondition is the condition type that indicates the CoreCluster controller is degraded.
+	CoreClusterControllerDegradedCondition = "CoreClusterControllerDegraded"
 )
 
 var (
@@ -53,6 +59,9 @@ var (
 	errUnsupportedPlatformType                     = errors.New("unsupported platform type")
 	errOpenshiftInfraShouldNotBeNil                = errors.New("infrastructure object should not be nil")
 	errOpenshiftInfrastructureNameShouldNotBeEmpty = errors.New("infrastructure object's infrastructureName should not be empty")
+
+	//nolint:gochecknoglobals
+	expectedConditionTypes = []string{CoreClusterControllerAvailableCondition, CoreClusterControllerDegradedCondition}
 )
 
 // CoreClusterController reconciles a Cluster object.
@@ -81,6 +90,8 @@ func (r *CoreClusterController) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // Reconcile reconciles the core cluster object for the openshift-cluster-api namespace.
+//
+//nolint:funlen
 func (r *CoreClusterController) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx).WithName(controllerName)
 	logger.Info("Reconciling core cluster")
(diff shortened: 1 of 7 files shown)


pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -81,36 +81,49 @@ func (r *CoreClusterController) SetupWithManager(mgr ctrl.Manager) error {
...
+		err        error
+		conditions []configv1.ClusterOperatorStatusCondition
+	)
+
+	defer func() {
+		// This function runs before every return, to ensure the controller conditions
+		// for that codepath are correctly set to the ClusterOperator status.
+		if err := r.setControllerConditions(ctx, conditions); err != nil && reterror == nil {
+			reterror = fmt.Errorf("failed to sync cluster operator status: %w", err)
+		}
```
JoelSpeed, line 99: If that status sync fails and there was another error, do we still want to log the error?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -81,36 +81,49 @@ func (r *CoreClusterController) SetupWithManager(mgr ctrl.Manager) error {
...
 	if err := r.ensureCoreClusterControlPlaneInitializedCondition(ctx, cluster); err != nil {
+		conditions = conditionsDegradedAvailable(err)
+
 		return ctrl.Result{}, fmt.Errorf("failed to ensure core cluster has the ControlPlaneInitializedCondition: %w", err)
 	}
 
-	if err := r.SetStatusAvailable(ctx, ""); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
-	}
+	conditions = conditionsNormal()
```
JoelSpeed, line 126: I think helpers here are detrimental to the goal here. I'd rather we just set the conditions we care about
I would rather we set individual conditions as they come up and not tie them together in things like `DegradedAvailable`, how do we know if there's an error in `ensureCoreCluster` that it should be available still?

Do we have any helpers on a `[]configv1.ClusterOperatorStatusCondition` like CAPI does so that we can do things like `SetCondition`?

I think a small wrapper around them that lets us call something like below would make this more flexible
```
conds.SetCondition("Degraded", "CoreClusterCreateFailed", err)
```

We can then also set specific error reasons, or even pass the conditions list into things like `ensureCoreCluster` to allow it to provide more specific condition reasons when it does need to set degraded, or set available false if it detects that's appropriate within the lower level function

pkg/controllers/corecluster/corecluster_controller.go
damdo: This is bubbled up as we set `reterror` as a named return argument, so that would throw a reconciler error and log that in the controller logs.

pkg/controllers/corecluster/corecluster_controller.go
damdo: The wrapper is essentially what:
```
operatorstatus.NewClusterOperatorStatusCondition(operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionFalse, operatorstatus.ReasonAsExpected, "Core Cluster Controller works as expected"),
```

does.
So I don't think we should duplicate that.
I'm ok with setting those individiually rather than with helpers, although I think it hurts readability a bit.
Will push the change.

pkg/controllers/corecluster/corecluster_controller.go
JoelSpeed: The original error gets bubbled up, looking at the logic as it is now, the status sync error would not get bubbled up

pkg/controllers/corecluster/corecluster_controller.go
JoelSpeed: It's not the same, the wrapper that exists in CAPI does a lot more, for example, it checks if the condition exists already in the list, and ensures that we do not duplicate conditions in the list, we don't otherwise have that right now and it would certainly be easier to have a common wrapper for that across all of these than re-implementing the deduplication many times

I believe it also sets the transition timestamps as it sets those values, which could also be useful for us\

In CAPI, their wrapper wraps the equivalent of `operatorstatus.NewClusterOperatorStatusCondition` for us

pkg/controllers/corecluster/corecluster_controller.go
damdo: Yes, I think you are right, the outer wrapped error takes precedence vs the named return argument error.
I logged the innermost one in case both happen.

pkg/controllers/corecluster/corecluster_controller.go
damdo: I get what you say now, wasn't clearer to me with only the above comment.
Ended up defining an `operatorstatus.SetCondition()` wrapper in the `operatorstatus` pkg centrally. That does:
- ensure we don't duplicate but update already existing conditions
- set transition timestamps (this was already in place in the inner NewClusterOperatorStatusCondition())
- I've also had a go at using ServerSideApply for `statusSync` ing.

Let me know WDYT

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -86,33 +86,60 @@ func (r *CoreClusterController) Reconcile(ctx context.Context, req reconcile.Req
...
+		operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionTrue,
+			operatorstatus.ReasonSyncFailed, fmt.Sprintf("controller is degraded: %s", err.Error()))
 
-	if err := r.SetStatusAvailable(ctx, ""); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
+		return ctrl.Result{}, fmt.Errorf("failed to ensure core cluster has the ControlPlaneInitializedCondition: %w", err)
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
```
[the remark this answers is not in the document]
damdo: I think we do, if we want to trigger a reconcile on status sync errors. See the comment I added in the defer function.

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -86,33 +86,60 @@ func (r *CoreClusterController) Reconcile(ctx context.Context, req reconcile.Req
...
+
+	conditions := &[]configv1.ClusterOperatorStatusCondition{}
+
+	// We optimistically set Degraded=false at the beginning.
+	operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerDegradedCondition, configv1.ConditionFalse,
+		operatorstatus.ReasonAsExpected, "controller works as expected")
+
+	// We optimistically set Available=true at the beginning.
+	operatorstatus.SetCondition(conditions, operatorstatus.CoreClusterControllerAvailableCondition, configv1.ConditionTrue,
+		operatorstatus.ReasonAsExpected, "controller is available")
```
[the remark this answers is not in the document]
JoelSpeed: I think it would be better for the controller to set the unknown explicitly rather than have it inferred by higher level objects.

I think it's best practice to "set all known condition types for a controller on every reconcile"

pkg/controllers/corecluster/corecluster_controller.go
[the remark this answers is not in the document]
JoelSpeed: But at this point, err should be nil no? So we are returning a nil variable?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -240,3 +267,17 @@ func getOCPInfrastructureName(infra *configv1.Infrastructure) (string, error) {
...
 }
+
+// syncControllerConditions sets the CoreClusterController conditions.
+func (r *CoreClusterController) syncControllerConditions(ctx context.Context, conds *[]configv1.ClusterOperatorStatusCondition) error {
+	co, err := r.GetOrCreateClusterOperator(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster operator: %w", err)
+	}
+
+	if err := r.SyncStatus(ctx, controllerName, co, *conds); err != nil {
```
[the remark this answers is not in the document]
JoelSpeed: Oh, have you got a link to an example of that? I thought a change of reason would represent a transition too 🤔

pkg/controllers/corecluster/corecluster_controller.go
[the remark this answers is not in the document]
damdo: Got it thanks, the code includes it now.

pkg/controllers/corecluster/corecluster_controller.go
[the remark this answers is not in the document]
damdo: From the godoc of the apps API :
https://github.com/kubernetes/api/blob/4dccc5e86b957cea946a63c4f052ee7dec3946ce/apps/v1/types.go#L746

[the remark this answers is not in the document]
damdo: Also here: https://github.com/kubernetes/kubernetes/blob/c7489b20f209a75ec7a239f6a40c3780c9d6e40d/pkg/controller/podautoscaler/horizontal.go#L1426-L1428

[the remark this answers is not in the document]
JoelSpeed: Fair enough, I'd be interested to see if other folks around OCP agree with that, maybe start a thread somewhere appropriate? Happy to go with that for now though

pkg/controllers/corecluster/corecluster_controller.go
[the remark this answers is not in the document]
damdo: The defer function runs before this return line, and may or may not set `err` to a `syncErr`.

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -81,38 +87,89 @@ func (r *CoreClusterController) SetupWithManager(mgr ctrl.Manager) error {
...
+		if syncErr := r.SyncControllerConditions(ctx, co, controllerName, conditions,
+			[]string{CoreClusterControllerAvailableCondition, CoreClusterControllerDegradedCondition}); syncErr != nil {
+			if err != nil {
+				// The sync of the controller conditions failed but there is also a non nil reconcile error.
+				// Log the sync error and pass down the reconciler one.
+				logger.Error(syncErr, "failed to sync cluster operator status")
+			} else {
+				// If there is no reconcile error, but there is a sync error, then we want to propagate the
+				// sync error to be returned by the Reconcile function, so that it will cause an immediate requeue/retry.
+				err = syncErr
```
JoelSpeed, line 119: Is this the better way to achieve this vs named return? I thought with named return the final return could be `nil`, but, if this works, and it's commented, I don't have a strong opinion, just curious if both paths work

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -81,38 +87,89 @@ func (r *CoreClusterController) SetupWithManager(mgr ctrl.Manager) error {
...
+	}()
+
 	ocpInfrastructureName, err := getOCPInfrastructureName(r.Infra)
 	if err != nil {
+		operatorstatus.SetCondition(conditions, CoreClusterControllerDegradedCondition, configv1.ConditionTrue,
+			operatorstatus.ReasonSyncFailed, fmt.Sprintf("controller is degraded: %s", err.Error()))
+
+		operatorstatus.SetCondition(conditions, CoreClusterControllerAvailableCondition, configv1.ConditionTrue,
+			operatorstatus.ReasonAsExpected, "controller is available")
+
```
JoelSpeed, line 131: We can't say that the controller is available here, or in any of the error cases. What if we never fetched the infra, and so we haven't synced the core cluster at all?

We should only set available true at the end when there have been no errors, no?

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -81,38 +87,89 @@ func (r *CoreClusterController) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
 
 	if !cluster.DeletionTimestamp.IsZero() {
-		if err := r.SetStatusAvailable(ctx, ""); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to set status available: %w", err)
-		}
+		operatorstatus.SetCondition(conditions, CoreClusterControllerDegradedCondition, configv1.ConditionFalse,
+			operatorstatus.ReasonAsExpected, "controller works as expected")
+
+		operatorstatus.SetCondition(conditions, CoreClusterControllerAvailableCondition, configv1.ConditionTrue,
```
JoelSpeed, line 150: If the cluster is being deleted, are we really available?

pkg/operatorstatus/operator_status.go
```diff
@@ -157,14 +158,87 @@ func (r *ClusterOperatorStatusClient) GetOrCreateClusterOperator(ctx context.Con
...
+
+func getMissingConditionTypes(conditions *[]configv1.ClusterOperatorStatusCondition, expectedConditionTypes []string) []string {
+	missing := []string{}
+
+	for _, e := range expectedConditionTypes {
+		found := false
+
+		for _, c := range *conditions {
+			if c.Type == configv1.ClusterStatusConditionType(e) {
+				found = true
```
JoelSpeed, line 183: Break the inner loop here?

pkg/controllers/corecluster/corecluster_controller.go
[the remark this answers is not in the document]
damdo: Yeah might be worth asking out of curiosity. I'll keep this as is for now.

pkg/controllers/corecluster/corecluster_controller.go
damdo: I think this is easier to reason about, as one doesn't need to understand ordering of nil vs named returns.

pkg/controllers/corecluster/corecluster_controller.go
damdo: Again, this boils down to the actual meaning we want to give to Available=false vs Unknown.
Happy to stay unknown in these cases and only go available once we finally reach the bottom of the reconcile logic.

JoelSpeed: Available for the CoreCluster controller means a CoreCluster is in the cluster and doesn't have a deletion timestamp
(which shouldn't happen anyway because admission validations block this), is there anything else it needs to be considered available? A ready condition?

So any path that errors, doesn't know if the core cluster is there right? We only know the core cluster is there and available when we have completed a successful reconcile, so IMO, it should only be available true right at the end of the reconcile, or, once we know that the corecluster exists (if that is earlier)

pkg/controllers/corecluster/corecluster_controller.go
damdo: That's trickier. I think we will end up disallowing core cluster deletion via VAP/WebHook, but the operator will still keep reconciling without error and will recreate this object.

Either could be ok depending on how we look at things.
If we don't explicitly set it here, we are probably inheriting Available from the previous state.

Ok, now I dropped and updated the test, let's see how we roll with this.

pkg/operatorstatus/operator_status.go
damdo: Ah yes, good catch.

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -81,38 +87,89 @@ func (r *CoreClusterController) SetupWithManager(mgr ctrl.Manager) error {
...
 
+	var err error
+
+	co, err := r.GetOrCreateClusterOperator(ctx)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get or create cluster operator: %w", err)
+	}
+
+	// Populate conditions from the existing cluster operator status.
+	conditions := &co.Status.Conditions
```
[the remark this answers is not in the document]
damdo: Added filtering for this

pkg/controllers/corecluster/corecluster_controller.go
JoelSpeed: It seemed to be ending the reconcile once it detected the deletion timestamp? Is that expected then? That seems unusual given your description that it should continue to reconcile

damdo: Keep reconciling I mean coming back to a new reconciliation cycle once the object as been fully deleted an perform a recreation.

Bear in mind I haven't touched this logic at all with the PR, so the behaviour is as it was prior to my changes.

One thing we can do here to improve speed of recovery from object deletion is probably to add a requeueAfter to detect the object is being deleted. So we can catch it quickly and recreate.

damdo:
Blocked by https://github.com/openshift/api/pull/2194 and https://github.com/openshift/cluster-version-operator/pull/1149

pkg/controllers/corecluster/corecluster_controller.go
```diff
@@ -46,13 +46,22 @@ const (
...
+	CoreClusterControllerDegradedCondition = "CoreClusterControllerDegraded"
 )
 
 var (
 	errPlatformStatusShouldNotBeNil                = errors.New("infrastructure platformStatus should not be nil")
 	errUnsupportedPlatformType                     = errors.New("unsupported platform type")
 	errOpenshiftInfraShouldNotBeNil                = errors.New("infrastructure object should not be nil")
 	errOpenshiftInfrastructureNameShouldNotBeEmpty = errors.New("infrastructure object's infrastructureName should not be empty")
+
+	//nolint:gochecknoglobals
```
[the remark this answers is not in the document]
damdo: Yes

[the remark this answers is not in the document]
damdo: Talked about this out of band

pkg/controllers/corecluster/corecluster_controller.go
damdo: Talked about this out of band. It's cool.

damdo:
/assign @JoelSpeed 
For approval

PR #258, opened 2025-02-07
OCPCLOUD-2645: MAPI to CAPI Machine synchronization

damdo:
Follow up on https://github.com/openshift/cluster-capi-operator/pull/237 to perform the machine synchronisation.


pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -244,3 +432,154 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+
+	if err := r.Status().Patch(ctx, mapiMachine, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner(controllerName)); err != nil {
+		return fmt.Errorf("failed to patch MAPI machine status with synchronized condition: %w", err)
+	}
+
+	return nil
+}
+
+// setLastTransitionTime determines if the last transition time should be set or updated for a given condition type.
+func setLastTransitionTime(condType machinev1beta1.ConditionType, conditions []machinev1beta1.Condition, conditionAc *machinev1applyconfigs.ConditionApplyConfiguration) {
```
theobarberbany, line 543: I reckon a bunch of these we'll eventually want in a utilities package / as helpers to be used across controllers

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -91,35 +140,234 @@ var _ = Describe("MachineSync Reconciler", func() {
...
+
+			It("should not make any changes", func() {
+				resourceVersion := mapiMachine.GetResourceVersion()
+				Consistently(k.Object(mapiMachine), timeout).Should(
+					HaveField("ResourceVersion", Equal(resourceVersion)),
+				)
+			})
+		})
+
+		Context("when the MAPI machine set does not exist and the CAPI machine set does", func() {
```
JoelSpeed, line 289: machine set in this description 
```suggestion
		Context("when the MAPI machine does not exist and the CAPI machine does", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -91,35 +140,234 @@ var _ = Describe("MachineSync Reconciler", func() {
...
+			BeforeEach(func() {
+				By("Creating the CAPI machine set")
```
JoelSpeed, line 291:
```suggestion
				By("Creating the CAPI machine")
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -91,35 +140,234 @@ var _ = Describe("MachineSync Reconciler", func() {
...
+				capiMachine = capiMachineBuilder.Build()
+
+				Expect(k8sClient.Create(ctx, capiMachine)).Should(Succeed())
+			})
+
+			It("should not make any changes to the CAPI machine set", func() {
```
JoelSpeed, line 297:
```suggestion
			It("should not make any changes to the CAPI machine", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -91,35 +140,234 @@ var _ = Describe("MachineSync Reconciler", func() {
...
+				resourceVersion := capiMachine.GetResourceVersion()
+				Consistently(k.Object(capiMachine), timeout).Should(
+					HaveField("ResourceVersion", Equal(resourceVersion)),
+				)
+			})
+
+			It("should not create a MAPI machine set", func() {
```
JoelSpeed, line 304:
```suggestion
			It("should not create a MAPI machine", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -91,35 +140,234 @@ var _ = Describe("MachineSync Reconciler", func() {
...
+				Expect(k8sClient.Create(ctx, mapiMachine)).Should(Succeed())
+
+				By("Setting the AuthoritativeAPI to MachineAPI")
+				Eventually(k.UpdateStatus(mapiMachine, func() {
+					mapiMachine.Status.AuthoritativeAPI = machinev1beta1.MachineAuthorityMachineAPI
+				})).Should(Succeed(), "should have succeeded updating the AuthoritativeAPI")
+			})
+
+			Context("when the CAPI machine does not exist", func() {
+				It("should create the CAPI machine and CAPI infra machine", func() {
```
JoelSpeed, line 327: Could these be separated to two `It` blocks?

pkg/util/sync.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright 2024 Red Hat, Inc.
```
JoelSpeed, line 2: 2025?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -39,11 +47,28 @@ import (
...
+	reasonFailedToUpdateCAPIMachine        = "FailedToUpdateCAPIMachine"
+	reasonFailedToCreateCAPIMachine        = "FailedToCreateCAPIMachine"
+
 	capiNamespace  string = "openshift-cluster-api"
 	mapiNamespace  string = "openshift-machine-api"
-	machineSetKind string = "MachineSet"
+	machineKind    string = "Machine"
 	controllerName string = "MachineSyncController"
+
+	messageSuccessfullySynchronized = "Successfully synchronized CAPI Machine to MAPI"
```
JoelSpeed, line 71: Is this the correct way around?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -184,6 +210,168 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
+		return mapi2capi.FromAWSMachineAndInfra(mapiMachine, r.Infra).ToMachineAndInfrastructureMachine() //nolint:wrapcheck
+	case configv1.PowerVSPlatformType:
+		return mapi2capi.FromPowerVSMachineAndInfra(mapiMachine, r.Infra).ToMachineAndInfrastructureMachine() //nolint:wrapcheck
+	default:
+		return nil, nil, nil, fmt.Errorf("%w: %s", errPlatformNotSupported, r.Platform)
+	}
+}
+
+// createOrUpdateCAPIInfraMachine creates a CAPI infra machine from a MAPI machine, or updates if it exists and it is out of date.
+func (r *MachineSyncReconciler) createOrUpdateCAPIInfraMachine(ctx context.Context, mapiMachine *machinev1beta1.Machine, infraMachine client.Object, newCAPIInfraMachine client.Object) (ctrl.Result, error) { //nolint:unparam
```
JoelSpeed, line 274: What params are not being used?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -184,6 +210,168 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
+
+			return ctrl.Result{}, createErr
+		}
+
+		logger.Info("Successfully created CAPI infra machine")
+
+		return ctrl.Result{}, nil
+	}
+
+	isEqualCAPIInfraMachine, err := capiInfraMachineIsEqual(r.Platform, infraMachine, newCAPIInfraMachine)
```
JoelSpeed, line 294: CAPI infra machines are generally immutable right? So what happens when someone changes the providerSpec on MAPI, I guess the conversion logic starts reporting that it's broken, and there's no get out of jail apart from going back?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -184,6 +210,168 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
+	}
+
+	if isEqualCAPIInfraMachine {
+		logger.Info("No changes detected in CAPI infra machine")
+		return ctrl.Result{}, nil
+	}
+
+	logger.Info("Updating CAPI infra machine")
+
+	if err := r.Update(ctx, newCAPIInfraMachine); err != nil {
```
JoelSpeed, line 314: The only updates we really expect are metadata right?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -184,6 +210,168 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
+		return ctrl.Result{}, updateErr
+	}
+
+	logger.Info("Successfully updated CAPI infra machine")
+
+	return ctrl.Result{}, nil
+}
+
+// createOrUpdateCAPIMachine creates a CAPI machine from a MAPI one, or updates if it exists and it is out of date.
+func (r *MachineSyncReconciler) createOrUpdateCAPIMachine(ctx context.Context, mapiMachine *machinev1beta1.Machine, capiMachine *capiv1beta1.Machine, newCAPIMachine *capiv1beta1.Machine) (ctrl.Result, error) { //nolint:unparam
```
JoelSpeed, line 333: Why unparam?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -244,3 +432,109 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+	switch r.Platform {
+	case configv1.AWSPlatformType:
+		infraCluster = &capav1beta2.AWSCluster{}
+		infraMachine = &capav1beta2.AWSMachine{}
+	case configv1.PowerVSPlatformType:
+		infraCluster = &capibmv1.IBMPowerVSCluster{}
+		infraMachine = &capibmv1.IBMPowerVSMachine{}
+	default:
+		return nil, nil, fmt.Errorf("%w: %s", errPlatformNotSupported, r.Platform)
+	}
```
JoelSpeed, line 460: This switch will grow quite large, should we factor it out into a separate function?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -244,3 +432,109 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+		return nil, nil, fmt.Errorf("failed to get CAPI infrastructure machine: %w", err)
+	}
+
+	return infraCluster, infraMachine, nil
+}
+
+// capiInfraMachineIsEqual checks whether the provided CAPI infra machines are equal.
+func capiInfraMachineIsEqual(platform configv1.PlatformType, infraMachine1, infraMachine2 client.Object) (bool, error) {
+	switch platform {
+	case configv1.AWSPlatformType:
```
JoelSpeed, line 476: Perhaps we need to create some sort of function/interface that handles this kind of comparison so that we aren't going to just make one massive switch here with the same thing for each platform?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -244,3 +432,109 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+		return reflect.DeepEqual(typedInfraMachine1.Spec, typedinfraMachine2.Spec) && util.ObjectMetaIsEqual(typedInfraMachine1.ObjectMeta, typedinfraMachine2.ObjectMeta), nil
+	default:
+		return false, fmt.Errorf("%w: %s", errPlatformNotSupported, platform)
+	}
+}
+
+// updateSynchronizedConditionWithPatch updates the synchronized condition
+// using a server side apply patch. We do this to force ownership of the
+// 'Synchronized' condition and 'SynchronizedGeneration'.
+func (r *MachineSyncReconciler) updateSynchronizedConditionWithPatch(ctx context.Context, mapiMachine *machinev1beta1.Machine, status corev1.ConditionStatus, reason, message string, generation *int64) error {
```
JoelSpeed, line 508: Why not `applySynchronizedCondition`?

Does this controller make any other writes to the object? We may want the synchronized condition to use its own field owner otherwise other writes could remove the condition, especially when we start copying status back and forth

pkg/controllers/machinesync/machine_sync_controller.go
damdo: It is for the empty `ctrl.Result{}` that is always returned.
Stylistically I'd prefer to keep that in the function returns, as it is a method over the reconciler.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: > applySynchronizedCondition

Happy to change to this though, I think it makes more sense.

damdo: > Does this controller make any other writes to the object? We may want the synchronized condition to use its own field owner otherwise other writes could remove the condition, especially when we start copying status back and forth

It does resync the object, so yeah, probably nothing else on the status side at the moment, but there will be

pkg/controllers/machinesync/machine_sync_controller.go
damdo: It is not and we had the same issue for the machineset_sync. Now fixed, Thanks.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: It is for the empty ctrl.Result{} that is always returned.
Stylistically I'd prefer to keep that in the function returns, as it is a method over the reconciler.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Here we are in the scenario where the `authoritativeAPI: MachineAPI` (see the switch where this reconciliation func branches out).

Even if the CAPI infra machine is immutable (meaning if you change its spec it doesn't change the underlying instance) my understanding is we don't care at this point of the logic, as the authoritativeness is in MAPIs hands.
We will just update the CAPI infra machine object to reflect MAPI's providerSpec.

It seems to me that what you are talking about is something that is going to be dealt with when we have `authoritativeAPI: ClusterAPI`.

But again, maybe I am reading you wrong/have a different understanding.

damdo: Or are you talking about object immutability?

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Let's talk about that here: https://github.com/openshift/cluster-capi-operator/pull/258#discussion_r1963236079

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Agreed. Repurposed `getInfraMachineFromProvider` for that too.

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Ah, IIRC this was named like this because initially I used a patch, rather than SSA. Must've not updated helper func names :)

JoelSpeed: Lets change this, and I think use a specific name for the field manager for the synchronized condition so that in the future, if we start writing the status otherwise, the synchronized condition can still be treated as a separate transaction for the purposes of SSA

pkg/controllers/machinesync/machine_sync_controller.go
JoelSpeed: I was wondering about object immutability, could we see errors if we try to update the object and get told that something is immutable? In that case, I would guess we want to delete and recreate the object?

Probably can wait until we observe this to come up with a fix, but something to keep in mind

pkg/controllers/machinesync/machine_sync_controller.go
damdo: I had a go at this, tried with a common interface, but I am not sure we have common "behaviour" (methods) stable across all the InfraMachine Provide APIs.

Then I tried with unstructured.unstructured, but it feels pretty unreliable and quite complex.
Not sure it is worth the squeeze.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Renamed func and added a more specific field owner.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Tested object/instance immutability for a CAPA AWSMachine.
The [CAPA webhook](https://github.com/openshift/cluster-api-provider-aws/blob/1be2193f0afa34e5ffd66f0cd04545542629e16f/openshift/manifests/0000_30_cluster-api_04_cm.infrastructure-aws.yaml#L18300-L18309) (which we run) denies the change. The API validation allows it, the controller allows it and doesn't react on it.

Detailed steps:

Steps:
```
# delete the AWS transport configmap (to stop capiinstaller recreating the WH)
oc -n openshift-cluster-api delete cm aws

# delete the ValidatingWebhookConfig for CAPA
oc -n openshift-cluster-api delete validatingwebhookconfiguration capa-validating-webhook-configuration

# edit the AWSMachine, e.g. changing the instance type
oc -n openshift-cluster-api edit awsmachine aws-machineset-d2tjk
```

Result:
The change was allowed.
The CAPA controllers didn't even care about the change.
The underlying cloud instance type stayed the same and nothing errored.
The object instance type

JoelSpeed: I think we want to continue running the webhooks, since over time there is a plan to move those immutability checks into the CRD schema

So in that case, I guess we need to detect the immutability issue, and skip the update somehow? Or mark the synchronisation as degraded for this particular machine, until the user puts things back?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -441,7 +437,7 @@ func (r *MachineSetSyncReconciler) updateSynchronizedConditionWithPatch(ctx cont
 	msAc := machinev1applyconfigs.MachineSet(mapiMachineSet.GetName(), mapiMachineSet.GetNamespace()).
 		WithStatus(statusAc)
 
-	if err := r.Status().Patch(ctx, mapiMachineSet, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner("machineset-sync-controller")); err != nil {
+	if err := r.Status().Patch(ctx, mapiMachineSet, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner(controllerName+"-SyncronizedCondition")); err != nil {
```
JoelSpeed, line 440:
```suggestion
	if err := r.Status().Patch(ctx, mapiMachineSet, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner(controllerName+"-SynchronizedCondition")); err != nil {
```

pkg/controllers/machinesync/machine_sync_controller.go
JoelSpeed: Thinking about this a little more, we could build a `Provider` interface to capture the entirety of the platform specific knowledge

Which would then mean that everything a platform implementer needs to implement is contained in 1/2 files, vs being spread out across the controllers

It would look something like

```
func GetPlatformProvider(platform configv1.Platform) Provider {
  switch platform {
  // One switch to rule them all
  default:
    panic("the controller calling this code should not be running on an unsupported platform")
  }
}

type Provider interface {
  InfraMachineType client.Object
  InfraClusterType client.Object
  ConvertMAPIToCAPIMachine(...) (...)
  ConvertMAPIToCAPIMachineSet(...) (...)
  ConvertCAPIToMAPIMachine(...) (...)
  ConvertCAPIToMAPIMachineSet(...) (...)
  CAPIInfraMachineEqual(...) (...)
  CAPIInfraMachineTemplateEqaul(...) (...)
  ....
}
```

This could all be in a separate package and mean that the core controller logic need not know anything about platforms other than `GetPlatformProvider` will always return it a non-nil version of this interface 🤔

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -244,3 +433,103 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+		WithConditions(conditionAc)
+
+	if status == corev1.ConditionTrue && generation != nil {
+		statusAc = statusAc.WithSynchronizedGeneration(*generation)
+	}
+
+	msAc := machinev1applyconfigs.Machine(mapiMachine.GetName(), mapiMachine.GetNamespace()).
+		WithStatus(statusAc)
+
+	if err := r.Status().Patch(ctx, mapiMachine, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner(controllerName+"-SyncronizedCondition")); err != nil {
```
JoelSpeed, line 530: Please check if there are expectations on the names of the field owners, they may supposed to be lowercase or not include hyphens, I'm not sure but we should check the requirements
```suggestion
	if err := r.Status().Patch(ctx, mapiMachine, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner(controllerName+"-SynchronizedCondition")); err != nil {
```

pkg/controllers/machinesync/machine_sync_controller.go
damdo: We discussed this out of band. And we both think it is a good idea and the right way to go about for all the interaction points with the provider. We won't do this in this PR but rather do it across the board for the migration controllers in a follow up refactor PR.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Yes I checked. It dows work.

```
apiVersion: machine.openshift.io/v1beta1
kind: Machine
metadata:
  creationTimestamp: "2025-02-25T08:48:58Z"
  generateName: foo
  generation: 1
  managedFields:
  - apiVersion: machine.openshift.io/v1beta1
    fieldsType: FieldsV1
    manager: MachineSyncController-SynchronizedCondition
    operation: Apply
    subresource: status
    time: "2025-02-25T08:48:58Z"
...
```

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Done: https://issues.redhat.com/browse/OCPCLOUD-2894

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -91,35 +140,236 @@ var _ = Describe("MachineSync Reconciler", func() {
...
+
+				By("Setting the AuthoritativeAPI to Migrating")
+				Eventually(k.UpdateStatus(mapiMachine, func() {
+					mapiMachine.Status.AuthoritativeAPI = machinev1beta1.MachineAuthorityMigrating
+				})).Should(Succeed())
+			})
+
+			It("should not make any changes to either machine", func() {
+				mapiResourceVersion := mapiMachine.GetResourceVersion()
+				capiResourceVersion := capiMachine.GetResourceVersion()
```
JoelSpeed, line 256: Might want to add a comment explaining that this is the original RV since we haven't fetched the resource since it was created?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -91,35 +140,236 @@ var _ = Describe("MachineSync Reconciler", func() {
...
+								HaveField("Type", Equal(consts.SynchronizedCondition)),
+								HaveField("Status", Equal(corev1.ConditionTrue)),
+								HaveField("Reason", Equal("ResourceSynchronized")),
+								HaveField("Message", Equal("Successfully synchronized MAPI Machine to CAPI")),
+							))),
+					)
+				})
+			})
+
+			Context("when the CAPI machine does exist", func() {
```
JoelSpeed, line 350: What about when the infra machine exists but not the normal machine?

damdo: We do have a case for this right? 🤔 
Line 197 of the that file.

JoelSpeed: Ack yep, I see it now

pkg/controllers/machinesync/machine_sync_controller_test.go
damdo: Done.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: We discussed it here: https://redhat-internal.slack.com/archives/GE2HQ9QP4/p1740482982813219
and decided to delete and recreate the CAPI Infra machine when that's out of date, to follow what's stated in the enhancement.

Marking this as resolved.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -244,3 +450,120 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+		WithSeverity(severity)
+
+	util.SetLastTransitionTime(consts.SynchronizedCondition, mapiMachine.Status.Conditions, conditionAc)
+
+	statusAc := machinev1applyconfigs.MachineStatus().
+		WithConditions(conditionAc)
+
+	if status == corev1.ConditionTrue && generation != nil {
+		statusAc = statusAc.WithSynchronizedGeneration(*generation)
+	}
```
JoelSpeed, line 559: If you send an apply and don't send the synchronized generation, it will clear the previously set synchronized generation, which I don't think is what we want, we want to copy over the old value when the condition is false I think.

Could you add a test for this?

It's probably worth having some unit testing for this function specifically so that we can be certain of it's behaviour

pkg/conversion/mapi2capi/aws.go
```diff
@@ -464,6 +476,12 @@ func convertAWSBlockDeviceMappingSpecToCAPI(fldPath *field.Path, mapiBlockDevice
 		nonRootVolumes = append(nonRootVolumes, volume)
 	}
 
+	if len(nonRootVolumes) == 0 {
+		// In CAPA API this is optional and omitempty,
+		// so omit this if we don't have any.
+		nonRootVolumes = nil
+	}
+
```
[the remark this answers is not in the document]
damdo: Yeah that also works, happy either way. I've updated to follow your approach.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -184,36 +213,213 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
+
+		if condErr := r.applySynchronizedConditionWithPatch(
+			ctx, mapiMachine, corev1.ConditionFalse, reasonFailedToUpdateCAPIInfraMachine, deleteErr.Error(), nil); condErr != nil {
+			return ctrl.Result{}, false, utilerrors.NewAggregate([]error{deleteErr, condErr})
+		}
+
+		return ctrl.Result{}, false, deleteErr
+	}
+
+	logger.Info("Successfully deleted CAPI infra machine")
```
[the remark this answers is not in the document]
damdo: Yes, that's why I added a boolean in the return, so at one level higher we can detect the sync is in progress and set the condition appropriately.

I have added a variable (syncronizationIsProgressing) with a descriptive name and comment to make this clearer.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -184,36 +213,213 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
 }
 
 // shouldMirrorCAPIMachineToMAPIMachine takes a CAPI machine and determines if there should
 // be a MAPI mirror, it returns true only if:
 //
-// 1. The CAPI Machine is owned by a CAPI MachineSet,
-// 2. That owning CAPI MachineSet has a MAPI MachineSet Mirror.
+// 1. The CAPI Machine is owned by a CAPI Machine,
+// 2. That owning CAPI Machine has a MAPI Machine Mirror.
 func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context.Context, logger logr.Logger, machine *capiv1beta1.Machine) (bool, error) {
```
RadekManak, line 416: Looks like every word MachineSet was changed to Machine in this function.

damdo: Fixed, thanks.

JoelSpeed:
Some linter issues but otherwise LGTM
/approve

pkg/controllers/machinesync/machine_sync_controller.go
RadekManak: This is still not resolved. It's not just this comment, but every occurrence of the word MachineSet in this function.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -211,9 +431,9 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
 	logger.WithName("shouldMirrorCAPIMachineToMAPIMachine").
 		Info("checking if CAPI machine should be mirrored", "machine", machine.GetName())
 
-	// Check if owner refs point to CAPI MachineSet
+	// Check if owner refs point to CAPI Machine
 	for _, ref := range machine.ObjectMeta.OwnerReferences {
-		if ref.Kind != machineSetKind || ref.APIVersion != capiv1beta1.GroupVersion.String() {
+		if ref.Kind != machineKind || ref.APIVersion != capiv1beta1.GroupVersion.String() {
```
RadekManak, line 436: This is still not resolved. It's not just the function comment, but every occurrence of the word MachineSet in this function was changed to Machine. This changes what the code does.

RadekManak (changes requested):

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Oops yes, thanks @RadekManak ! Good catch!
We need to add tests for those things anyway, will work on it on separate PRs in concert with @theobarberbany

damdo: Fixed now.

PR #259, opened 2025-02-13
Apply provider manifests with SSA

mdbooth:
This change uses Server-Side Apply to apply provider manifests in place of the logic from library-go, which uses an Update with custom kind-specific client-side merge logic. Using SSA here should be equivalent to how these manifests would be applied by any other client.

The merge logic from library-go should no longer be required as long as the manifests are not specifying any fields which will be overwritten by an existing controller.

In particular, there should be no conflict with service-ca setting `caBundle` in various places, as long as the specified manifests do not include a caBundle. However, note that due to a validation bug in older versions of k8s, some CRDs do still specify an empty caBundle in their CRDs. These would have to be removed for this to work.

The expected flow of reconciles between cluster-capi-operator and service-ca-operator becomes:
* cluster-capi-operator: apply initial manifests
* service-ca-operator: adds caBundle to CRDs and validating webhooks
* cluster-capi-operator: triggered by update to manifests it managed applies manifests again

The final update will produce no change as long as no managed fields have been updated, so there will be no further reconciles.

However, objects with the provider label were being reconciled too often due to insufficient filtering. This PR includes a second commit which addresses that.

### DIFF
--- a/pkg/controllers/capiinstaller/capi_installer_controller.go
+++ b/pkg/controllers/capiinstaller/capi_installer_controller.go
@@ -46,9 +46,6 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-capi-operator/pkg/controllers"
 	"github.com/openshift/cluster-capi-operator/pkg/operatorstatus"
-	"github.com/openshift/library-go/pkg/operator/events"
-	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
-	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 
 	"github.com/klauspost/compress/zstd"
 )
@@ -74,7 +71,6 @@ const (
 
 var (
 	errEmptyProviderConfigMap = errors.New("provider configmap has no components data")
-	errResourceNotFound       = errors.New("resource not found")
 )
 
 // CapiInstallerController reconciles a ClusterOperator object.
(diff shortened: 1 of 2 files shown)


mdbooth:
Interesting observation: with this change kube-apiserver now reports that it's reloading all the provider CRDs every 5 minutes. However, *it was also doing that before*.

I think this might be good, and there's also another issue.

mdbooth:
Looks like everything interesting passed. I'm pushing an update which:
* Fixes lints
* Adds a proper field owner for the SSA operation
* Adds a second commit which should dramatically reduce the number of reconciles

mdbooth:
That last commit reduced the number of reconciles from 67 without this PR to 10.

However:
* That's still 8 more than I'd expect
* With only the first commit there were 1498 reconciles, which I can't explain

Still think this is an improvement.

mdbooth:
That last commit is probably too much for this PR. I'll split it into a separate PR if you want it.

Because we're now iterating directly over objects, it's trivial to add an owner reference to every object that we create. This allows us to be much more explicit in the objects we watch. It's also a more obvious fit for the behaviour of this controller: the controller doesn't really manage objects which have a capi provider label; it manages objects it created, whatever label they have.

It also gives us a way to implement the automatic removal of objects which are no longer referenced in a provider ConfigMap. This actually came up in CAPO when we removed our MutatingWebhook! Clusterctl implements this correctly, but users who were doing it manually had failed installations after upgrade because there was a floating MutatingWebhook referring to an endpoint which no longer existed.

mdbooth:
Replaced by https://github.com/openshift/cluster-capi-operator/pull/292 as I no longer have write access to this fork.

PR #260, opened 2025-02-17
OCPCLOUD-2718: Add default VolumeSize for MAPI to CAPI conversion

racheljpg:
Hello! This is a draft PR to get feedback for the default VolumeSize for the MAPI to CAPI conversion for AWS. Thanks!

### DIFF
--- a/pkg/conversion/mapi2capi/aws.go
+++ b/pkg/conversion/mapi2capi/aws.go
@@ -475,12 +475,6 @@ func blockDeviceMappingSpecToVolume(fldPath *field.Path, bdm mapiv1.BlockDeviceM
 
 	capiKMSKey := convertKMSKeyToCAPI(bdm.EBS.KMSKey)
 
-	if bdm.EBS.VolumeSize == nil {
-		// The volume size is required in CAPA, we will have to return an error, until we can come up with an appropriate way to handle this.
-		// TODO(OCPCLOUD-2718): Either find a way to handle the required value, or, force users to set the value.
-		errs = append(errs, field.Required(fldPath.Child("ebs", "volumeSize"), "volumeSize is required, but is missing"))
-	}
-
 	if rootVolume && !ptr.Deref(bdm.EBS.DeleteOnTermination, true) {
 		warnings = append(warnings, field.Invalid(fldPath.Child("ebs", "deleteOnTermination"), bdm.EBS.DeleteOnTermination, "root volume must be deleted on termination, ignoring invalid value false").Error())
 	} else if !rootVolume && !ptr.Deref(bdm.EBS.DeleteOnTermination, true) {
@@ -494,7 +488,7 @@ func blockDeviceMappingSpecToVolume(fldPath *field.Path, bdm mapiv1.BlockDeviceM
 
 	return capav1.Volume{
 		DeviceName:    ptr.Deref(bdm.DeviceName, ""),
-		Size:          *bdm.EBS.VolumeSize,
+		Size:          ptr.Deref(bdm.EBS.VolumeSize, 120), // The installer uses 120GiB by default as of 4.19.
 		Type:          capav1.VolumeType(ptr.Deref(bdm.EBS.VolumeType, "")),
 		IOPS:          ptr.Deref(bdm.EBS.Iops, 0),
 		Encrypted:     bdm.EBS.Encrypted,
--- a/pkg/conversion/mapi2capi/aws_fuzz_test.go
+++ b/pkg/conversion/mapi2capi/aws_fuzz_test.go
@@ -145,6 +145,7 @@ func awsProviderSpecFuzzerFuncs(codecs runtimeserializer.CodecFactory) []interfa
 			c.FuzzNoCustom(ebs)
 
 			// Fuzz required fields so that they are not empty.
+			// Setting volumeSize to a random int64 value.
 			if ebs.VolumeSize == nil {
 				ebs.VolumeSize = ptr.To(c.Int63())
 			}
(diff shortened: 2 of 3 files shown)


JoelSpeed (commented):
Please can you check to see if there are any specific things we are doing in the fuzz testing that will be affected by this change? I think this won't round trip so I expect we have to make sure this field is always set?

Is it possible also to add any unit testing to the test suite we have?

pkg/conversion/mapi2capi/aws.go
```diff
@@ -514,6 +510,18 @@ func convertKMSKeyToCAPI(kmsKey mapiv1.AWSResourceReference) string {
...
 }
 
+func convertEBSVolumeSizeToCAPI(bdm *mapiv1.BlockDeviceMappingSpec) int64 {
+	// If bdm.EBS.VolumeSize is not nil, return the VolumeSize
+	if bdm.EBS.VolumeSize != nil {
+		return *bdm.EBS.VolumeSize
+	}
+
+	// Otherwise, return the default VolumeSize (in GB)
+	defaultVolumeSize := int64(120)
```
JoelSpeed, line 520: Would appreciate if we could explain where this 120 value came from? The installer?

racheljpg, line 520: I'm not 100% on this being a reasonable default value right now - need to do some more research on this

racheljpg: Sorry, I commented on being unsure if this is the right value for the default before refreshing the page so I had not seen this comment :) yes, i'm taking this from the installer, but I didn't go terribly deep into that research yet (like checking if it's ever changed there, etc) so let me do some further research to craft a more thoughtful response

JoelSpeed: Ok cool, if you can link/explain where this came from that would be helpful, and understanding how it's changed over time is also pretty useful. The installer folks may be able to help answer that question

racheljpg: I've done some digging and seen that a conversation was had about it in 2021 [here](https://github.com/openshift/openshift-docs/issues/35793#issuecomment-915504739), it seems like 100GB was the default, but 120GB was set to make sure the E2E tests pass reliably. Happy to ask the installer folks about this still if you think it could be necessary? Perhaps I'm being naive, but Is there a huge difference in setting it at 120GB instead of 100GB? :)

pkg/conversion/mapi2capi/aws_test.go
```diff
@@ -207,10 +207,8 @@ var _ = Describe("mapi2capi AWS conversion", func() {
 					EBS: &mapiv1.EBSBlockDeviceSpec{},
 				}}),
 			),
-			infra: infra,
-			expectedErrors: []string{
-				"spec.providerSpec.value.blockDevices[0].ebs.volumeSize: Required value: volumeSize is required, but is missing",
-			},
+			infra:            infra,
+			expectedErrors:   []string{}, // No error is expected anymore, because volumeSize is now set to a default amount if nil.
```
racheljpg, line 211: If I'm thinking about this the right way, it seems like this error shouldn't be expected anymore, because if the volumeSize is missing it should now be set (and this test is no longer spitting back this expected error). I don't really know if it's helpful to keep this test case here in this state? Is there a more useful way we could update this?

pkg/conversion/mapi2capi/aws_fuzz_test.go
```diff
@@ -145,6 +145,7 @@ func awsProviderSpecFuzzerFuncs(codecs runtimeserializer.CodecFactory) []interfa
 			c.FuzzNoCustom(ebs)
 
 			// Fuzz required fields so that they are not empty.
+			// Setting volumeSize to a random int64 value.
```
racheljpg, line 148: It seems the fuzz test is already setting the volumeSize if it's empty (to a random int64 value). Is having it constantly set at 120GiB what we want here, or is there any value in having it set to a random value?

racheljpg:
Hey @JoelSpeed sorry for the delay getting back to you but I've let you some more questions. Apologies if they're simple, but it seems to me like it was already somewhat accounted for in the tests as they are - although I think there are some cases that are probably not coming into my mind that would be useful for us?

pkg/conversion/mapi2capi/aws.go
JoelSpeed: I'm fairly happy to keep it at 120 if that's what we currently recommend for our CI

pkg/conversion/mapi2capi/aws.go
```diff
@@ -494,7 +490,7 @@ func blockDeviceMappingSpecToVolume(fldPath *field.Path, bdm mapiv1.BlockDeviceM
 
 	return capav1.Volume{
 		DeviceName:    ptr.Deref(bdm.DeviceName, ""),
-		Size:          *bdm.EBS.VolumeSize,
+		Size:          ebsVolumeSize,
```
JoelSpeed, line 493: You could `ptr.Deref` here, and add a comment inline
```suggestion
		Size:          ptr.Deref(bdm.EBS.VolumeSize, 120), // The installer uses 120GiB by default as of 4.19.
```

JoelSpeed: When you say invalidate? Do you mean, make it redundant? I guess that's fine no? We can delete the code?

AFAICT this would achieve the same thing that function is doing

racheljpg: Ah, yes I was mistaken - it does do what the function does already. Thanks!

PR #264, opened 2025-02-20
NO-JIRA: Add lifecyclehook script that updates manifests-gen

RadekManak:
Adds lifecyclehook script that will run make update-manifests-gen in openshift folder in the rebased repository.

This is meant to run before running generate-ocp-manifests.sh hook to ensure manifests-gen is present and up-to-date.

I considered these two scripts being combined, but having the option to quickly disable one is preferred.

### DIFF
--- a/hack/rebasebot-hook-scripts/update-ocp-manifests-gen.sh
+++ b/hack/rebasebot-hook-scripts/update-ocp-manifests-gen.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # This script is run by rebasebot during the rebase of the CAPI repositories.
 # This script must be added to rebasebot as a post-rebase hook, which will download it from this repository and execute it on the repository being rebased.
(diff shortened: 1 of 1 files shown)


hack/rebasebot-hook-scripts/update-ocp-manifests-gen.sh
```diff
@@ -0,0 +1,32 @@
...
+        git add -A
+        git commit "${author_flag[@]}" -q -m "$commit_message"
+    fi
+}
+
+main(){
+    pushd openshift
+    make update-manifests-gen
+    popd
+    stage_and_commit "UPSTREAM: <carry>: Update manifests generator"
```
JoelSpeed, line 29: Over time we will stack up lots of these, is there any reason we shouldn't drop them and just create a new fresh commit each time?

damdo (commented):
/approve

leaving LGTM to @JoelSpeed

PR #265, opened 2025-02-26
OCPCLOUD-2647: CAPI to MAPI Machine Synchronisation

theobarberbany:


pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -210,6 +221,50 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 
 // reconcileCAPIMachinetoMAPIMachine reconciles a CAPI Machine to a MAPI Machine.
 func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Context, capiMachine *capiv1beta1.Machine, mapiMachine *machinev1beta1.Machine) (ctrl.Result, error) {
```
damdo, line 223: This function is invoked in two places.
When it is invoked at like [199](https://github.com/openshift/cluster-capi-operator/pull/265/files#diff-a00680d231c0783b5333cc3e2c6f15e625b31572f57b7fb4b598dcb18d2418f3R199)

The `mapiMachine` argument is always an empty machine `mapiMachine := &machinev1beta1.Machine{}` as `mapiMachineNotFound` is `true`.

As such you can't use mapiMachine.Bla() in subsequent calls (line 255 to 265) as they'll all yield empty strings.

damdo: I'd also probably set the MAPI machine to nil [here](https://github.com/openshift/cluster-capi-operator/pull/265/files#diff-a00680d231c0783b5333cc3e2c6f15e625b31572f57b7fb4b598dcb18d2418f3R168-R169), like we do for the CAPI one [here](https://github.com/openshift/cluster-capi-operator/pull/265/files#diff-a00680d231c0783b5333cc3e2c6f15e625b31572f57b7fb4b598dcb18d2418f3R185-R186).

As later on ([here](https://github.com/openshift/cluster-capi-operator/pull/265/files#diff-a00680d231c0783b5333cc3e2c6f15e625b31572f57b7fb4b598dcb18d2418f3R486)) we check on it to decide whether to update it or to create it.

theobarberbany: > The mapiMachine argument is always an empty machine mapiMachine := &machinev1beta1.Machine{} as mapiMachineNotFound is true.

>As such you can't use mapiMachine.Bla() in subsequent calls (line 255 to 265) as they'll all yield empty strings.

Good catch, this makes sense! This is for the scenario where the MAPI Machine does not exist, and CAPI is authoritative and we get a failure either converting or fetching CAPI resources.

In this instance, we'll error trying to set the synchronized condition  and add that to the aggregate that'll get logged. I think it likely makes sense to not bother trying to set the condition and just emitting an event?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -334,18 +388,90 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+						)
+					})
+
+					It("should not create a MAPI machine", func() {
+						Consistently(k.ObjectList(&machinev1beta1.MachineList{}), timeout).ShouldNot(HaveField("Items",
+							ContainElement(HaveField("ObjectMeta.Name", Equal(capiMachine.GetName()))),
+						))
+					})
+				})
+
```
theobarberbany, line 444: Once we have owner reference conversion in place, this test case won't fail and can be un-commented / built out.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -200,17 +220,95 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+	if capiMachine == nil {
+		logger.Error(errCAPIMachineNotFound, "machine", mapiMachine.Name)
+
+		if condErr := r.applySynchronizedConditionWithPatch(
+			ctx, mapiMachine, corev1.ConditionFalse, reasonCAPIMachineNotFound, errCAPIMachineNotFound.Error(), nil); condErr != nil {
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{errCAPIMachineNotFound, condErr})
+		}
+
+		return ctrl.Result{}, errCAPIMachineNotFound
+	}
```
damdo, line 244: That's good that we check for this.
Although I think this needs to be changed up a bit.

We need to account for several things here, like what happens with when the CAPI Machine gets deleted?
What other sub-cases could there be?

While testing the migration controller I came up with this logic for it, which takes into account deletion but also things in between (phantom machines). But don't take it as truth :) it might need extra changes, tests & such.

```go
	phantomMAPIMachine := capiMachine == nil && mapiMachine != nil // happens when the CAPI machine gets deleted quicker than the MAPI one.
	capiMachineBeingDeleted := capiMachine != nil && !capiMachine.DeletionTimestamp.IsZero() && mapiMachine != nil

	if capiMachineBeingDeleted || phantomMAPIMachine {
		// CAPI Machine is (being) deleted.
		// Issue a deletion for the MAPI Machine mirror as well.
		logger.Info("Cluster API Machine is being deleted, issue deletion for the Machine API Machine mirror")
		if mapiMachine.DeletionTimestamp.IsZero() {
			if err := r.Delete(ctx, mapiMachine); err != nil {
				return ctrl.Result{}, fmt.Errorf("failed to delete MAPI Machine")
			}
		}

		// Requeue.
		return ctrl.Result{}, nil
	}
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -200,17 +220,95 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+
+	newMapiMachine, warns, err := r.convertCAPIToMAPIMachine(capiMachine, infraMachine, infraCluster)
+	if err != nil {
+		conversionErr := fmt.Errorf("failed to convert CAPI machine to MAPI machine: %w", err)
+
+		if mapiMachine == nil {
+			r.Recorder.Event(capiMachine, corev1.EventTypeWarning, "SynchronizationWarning", conversionErr.Error())
+			return ctrl.Result{}, conversionErr
+		}
```
damdo, line 270: Nice! So we at least give the user some warning there!

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -200,17 +220,95 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+		if condErr := r.applySynchronizedConditionWithPatch(
+			ctx, mapiMachine, corev1.ConditionFalse, reasonFailedToConvertCAPIMachineToMAPI, createUpdateErr.Error(), nil); condErr != nil {
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{createUpdateErr, condErr})
+		}
+
+		return result, createUpdateErr
+	}
+
+	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachine, corev1.ConditionTrue,
+		consts.ReasonResourceSynchronized, messageSuccessfullySynchronizedCAPItoMAPI, &newMapiMachine.Generation)
```
damdo, line 311: My understanding is that on `syncronizedGeneration` we should set the `metatada.generation` of the currently  authoritativeAPI resource (in this case CAPI Machine), we just finished syncing the non-authoritative resource (MAPI Machine) with.
```suggestion
	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachine, corev1.ConditionTrue,
		consts.ReasonResourceSynchronized, messageSuccessfullySynchronizedCAPItoMAPI, &capiMachine.Generation)
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -407,6 +524,44 @@ func (r *MachineSyncReconciler) createOrUpdateCAPIMachine(ctx context.Context, m
 	return ctrl.Result{}, nil
 }
 
+// createOrUpdateMAPIMachine creates a MAPI machine from a CAPI one, or updates
+// if it exists and it is out of date.
+func (r *MachineSyncReconciler) createOrUpdateMAPIMachine(ctx context.Context, mapiMachine *machinev1beta1.Machine, newMAPIMachine *machinev1beta1.Machine) (ctrl.Result, error) { //nolint:unparam
```
damdo, line 529: Let's change all the MAPI and CAPI contractions in the strings of this function to `Cluster API` and `Machine API`

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -428,16 +583,16 @@ func initInfraMachineAndInfraClusterFromProvider(platform configv1.PlatformType)
 // 1. The CAPI machine is owned by a CAPI machineset,
 // 2. That owning CAPI machineset has a MAPI machineset Mirror.
 func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context.Context, logger logr.Logger, machine *capiv1beta1.Machine) (bool, error) {
-	logger.WithName("shouldMirrorCAPIMachineToMAPIMachine").
-		Info("checking if CAPI machine should be mirrored", "machine", machine.GetName())
+	logger.V(4).WithName("shouldMirrorCAPIMachineToMAPIMachine").
+		Info("Checking if CAPI machine should be mirrored", "machine", machine.GetName())
```
damdo, line 587: contactions in this func too

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -544,3 +720,18 @@ var _ = Describe("applySynchronizedConditionWithPatch", func() {
...
+		return machinev1beta1.AWSMachineProviderConfig{}, nil
+	}
+
+	spec := machinev1beta1.AWSMachineProviderConfig{}
+	if err := yaml.Unmarshal(rawExtension.Raw, &spec); err != nil {
+		return machinev1beta1.AWSMachineProviderConfig{}, fmt.Errorf("error unmarshalling providerSpec: %w", err)
+	}
+
+	return spec, nil
+}
```
damdo, line 737: Can we wrap/use `mapi2capi.AWSProviderSpecFromRawExtension` then?

damdo (commented):
Mostly looks good, there are a couple of things we need to address, for which I left some comments.

Thanks!

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Ah nice! Cheers for the head start on 2648 :D 

I've not really considered deletion at all with this PR yet. The scenario I had in mind here was someone having a MAPI machine with AuthorititaveAPI set to CAPI, without a CAPI Machine existing. 

I'm unsure if we could actually get to this state once everything is put together, but was seeing segfaults in test setup where the CAPI machine was created after the MAPI one.

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: Yeah, that's the thinking. 

Do we need to update RBAC to allow events to be created on the CAPI machine? 🤔

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -272,6 +275,57 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
 			})
 		})
 
+		Context("when the MAPI machine has MachineAuthority set to Cluster API", func() {
```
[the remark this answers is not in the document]
theobarberbany: Do we want these tests in this PR, or a deletion specific one?

pkg/controllers/machinesync/machine_sync_controller_test.go
theobarberbany: I'm not actually sure here. 

I'm using it as follows:  

```go
It("should update the MAPI provider spec", func() {
  Eventually(k.Object(mapiMachine), timeout).Should(
    WithTransform(awsProviderSpecFromMachine,
      HaveField("InstanceType", Equal("m7i.4xlarge")),
      ))
})
```

I'm not sure how to pass  `mapiMachine.Spec.ProviderSpec.Value` to `withTransform` in this instance, keeping the `Eventually` over the `k.Object`.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -200,17 +220,95 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 	case machinev1beta1.MachineAuthorityClusterAPI:
 		return r.reconcileCAPIMachinetoMAPIMachine(ctx, capiMachine, mapiMachine)
 	case machinev1beta1.MachineAuthorityMigrating:
-		logger.Info("machine currently migrating", "machine", mapiMachine.GetName())
+		logger.Info("Machine currently migrating", "machine", mapiMachine.GetName())
 		return ctrl.Result{}, nil
 	default:
-		logger.Info("machine AuthoritativeAPI has unexpected value", "AuthoritativeAPI", mapiMachine.Status.AuthoritativeAPI)
+		logger.Info("Machine AuthoritativeAPI has unexpected value", "AuthoritativeAPI", mapiMachine.Status.AuthoritativeAPI)
```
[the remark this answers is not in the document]
theobarberbany: I think this is a good idea. Will update. ****

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: 🤦 good catch, this is a copypasta fubar :P

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -200,17 +220,95 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+	for _, warning := range warns {
+		logger.Info("Warning during conversion", "warning", warning)
+		r.Recorder.Event(mapiMachine, corev1.EventTypeWarning, "ConversionWarning", warning)
+	}
+
+	if mapiMachine != nil {
+		newMapiMachine.SetResourceVersion(util.GetResourceVersion(mapiMachine))
+	}
+
+	newMapiMachine.SetNamespace(mapiMachine.GetNamespace())
```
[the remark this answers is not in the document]
theobarberbany: Ah - I just need to set the namespace set on the reconciler regardless!

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Probably not, it is already quite wide. We can try and see though.

pkg/controllers/machinesync/machine_sync_controller_test.go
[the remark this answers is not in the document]
damdo: Probably in the follow up at this point

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Ok cool, we can add this in the follow up PR then 👍

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -200,17 +222,98 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+	}
+
+	if mapiMachine != nil {
+		newMapiMachine.SetResourceVersion(util.GetResourceVersion(mapiMachine))
+	}
+
+	newMapiMachine.SetNamespace(r.MAPINamespace)
+	newMapiMachine.Spec.AuthoritativeAPI = machinev1beta1.MachineAuthorityClusterAPI
+	newMapiMachine.Spec.Labels = util.MergeMaps(mapiMachine.Spec.Labels, newMapiMachine.Spec.Labels)
+	newMapiMachine.Labels = util.MergeMaps(mapiMachine.Labels, newMapiMachine.Labels)
```
damdo, line 297: mapiMachine.Labels will panic if mapiMachine is nil

theobarberbany: Ah I haven't pushed yet 😅

pkg/controllers/machinesync/machine_sync_controller_test.go
theobarberbany: Ohh - I see. Sure, makes sense.

damdo (commented):
Thanks @theobarberbany 
Looks good to me!

/approve
/lgtm

/hold

damdo:
This is behind the CustomNoUpgrade and MachineAPIMigration FeatureSet.
As such we are ok adding the acknowledge-critical-fixes-only label

/unhold

damdo:
As mentioned [above](https://github.com/openshift/cluster-capi-operator/pull/265#issuecomment-2757690288) all these tests are not running this PR's code, and they are all affected by the `namespace/openshift-marketplace` Image pull backoff issue: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/265/pull-ci-openshift-cluster-capi-operator-main-e2e-azure-ovn-techpreview/1905189333669777408 https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/265/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-ovn/1905221129690157056 https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/265/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-ovn-serial/1905242771908726784 https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/265/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-ovn-techpreview/1905221129736294400


/override ci/prow/e2e-aws-ovn-serial ci/prow/e2e-aws-ovn-techpreview ci/prow/e2e-aws-ovn

damdo:
See: 

/override ci/prow/e2e-aws-ovn-techpreview

pkg/controllers/machinesync/machine_sync_controller.go
theobarberbany: We also hit this if we create a MAPI machine with authorititve API as `ClusterAPI` when no CAPI counterpart exists. 

This means, e.g scaling a MAPI machineset with CAPI authority will fail as we don't find the CAPI machine. I think this previously worked, so we just need to explicitly handle this case and add a test.

damdo: Context: https://issues.redhat.com/browse/OCPBUGS-55349 and https://issues.redhat.com/browse/OCPBUGS-54970

PR #268, opened 2025-03-07
OCPCLOUD-2564: add migration controllers

damdo:
This PR adds the MAPI-CAPI migration controllers, resposible for handling the authoritativeness switching between MAPI and CAPI.

### DIFF
--- a/pkg/controllers/machinemigration/machine_migration_controller.go
+++ b/pkg/controllers/machinemigration/machine_migration_controller.go
@@ -1,0 +1,397 @@
+/*
+Copyright 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package machinemigration
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	capiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/annotations"
+	conditionsv1beta2 "sigs.k8s.io/cluster-api/util/conditions/v1beta2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	machinev1applyconfigs "github.com/openshift/client-go/machine/applyconfigurations/machine/v1beta1"
+	consts "github.com/openshift/cluster-capi-operator/pkg/controllers"
+	"github.com/openshift/cluster-capi-operator/pkg/util"
+)
+
+const controllerName = "MachineMigrationController"
+
+// MachineMigrationReconciler reconciles Machine resources for migration.
+type MachineMigrationReconciler struct {
+	client.Client
+	Scheme        *runtime.Scheme
+	Recorder      record.EventRecorder
+	CAPINamespace string
+	MAPINamespace string
+}
+
+// SetupWithManager sets up the MachineMigration controller.
+func (r *MachineMigrationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Allow the namespaces to be set externally for test purposes, when not set,
+	// default to the production namespaces.
+	if r.CAPINamespace == "" {
+		r.CAPINamespace = consts.DefaultManagedNamespace
+	}
+
+	if r.MAPINamespace == "" {
+		r.MAPINamespace = consts.DefaultMAPIManagedNamespace
+	}
+
+	if err := ctrl.NewControllerManagedBy(mgr).
+		Named(controllerName).
+		For(&machinev1beta1.Machine{}, builder.WithPredicates(util.FilterNamespace(r.MAPINamespace))).
+		Complete(r); err != nil {
+		return fmt.Errorf("failed to create controller: %w", err)
+	}
+
+	r.Client = mgr.GetClient()
+	r.Scheme = mgr.GetScheme()
+	r.Recorder = mgr.GetEventRecorderFor(controllerName)
+
+	return nil
+}
+
+// Reconcile performs the reconciliation for a Machine.
+//
+//nolint:funlen
+func (r *MachineMigrationReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithValues("namespace", req.Namespace, "name", req.Name)
+	ctx = logr.NewContext(ctx, logger)
+
+	logger.V(1).Info("Reconciling machine")
+	defer logger.V(1).Info("Finished reconciling machine")
+
+	mapiMachine := &machinev1beta1.Machine{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get MAPI machine: %w", err)
+	}
+
+	if mapiMachine.Spec.AuthoritativeAPI == mapiMachine.Status.AuthoritativeAPI {
+		// No migration is being requested for this resource, nothing to do.
+		return ctrl.Result{}, nil
+	}
+
+	// If authoritativeAPI status is empty, it means it is the first time we see this resource.
+	// Set the status.authoritativeAPI to match the spec.authoritativeAPI.
+	if mapiMachine.Status.AuthoritativeAPI == "" {
+		if err := r.applyStatusAuthoritativeAPIWithPatch(ctx, mapiMachine, mapiMachine.Spec.AuthoritativeAPI); err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to apply authoritativeAPI to status with patch: %w", err)
+		}
+
+		// Wait for the patching to take effect.
+		return ctrl.Result{}, nil
+	}
+
+	// Check if the Synchronized condition is set to True.
+	// If it is not, this indicates an unmigratable resource and therefore should take no action.
+	if cond, err := util.GetConditionStatus(mapiMachine, string(consts.SynchronizedCondition)); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to get synchronizedCondition for %s/%s: %w", mapiMachine.Namespace, mapiMachine.Name, err)
+	} else if cond != corev1.ConditionTrue {
+		logger.Info("Machine not synchronized with latest authoritative generation, will retry later")
+
+		return ctrl.Result{}, nil
+	}
+
+	// Make sure the authoritativeAPI resource status is set to migrating.
+	if mapiMachine.Status.AuthoritativeAPI != machinev1beta1.MachineAuthorityMigrating {
+		logger.Info("Detected migration request for machine")
+
+		if err := r.applyStatusAuthoritativeAPIWithPatch(ctx, mapiMachine, machinev1beta1.MachineAuthorityMigrating); err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to set authoritativeAPI %q to status: %w", machinev1beta1.MachineAuthorityMigrating, err)
+		}
+
+		logger.Info("Acknowledged migration request for machine")
+
+		// Wait for the change to propagate.
+		return ctrl.Result{}, nil
+	}
+
+	// Request pausing on the authoritative resource.
+	if updated, err := r.requestOldAuthoritativeResourcePaused(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to request pause on authoritative machine: %w", err)
+	} else if updated {
+		logger.Info("Requested pausing for authoritative machine")
+
+		// Wait for the change to propagate.
+		// Since there is not a watch for CAPI infra machines here
+		// we will not get an event for the CAPI infra machine's paused condition change,
+		// as such, manually requeue.
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
+
+	// Check that the authoritative resource is paused.
+	if paused, err := r.isOldAuthoritativeResourcePaused(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to check paused on authoritative machine: %w", err)
+	} else if !paused {
+		// The Authoritative API resource is not paused yet, requeue to check later.
+		logger.Info("Authoritative machine is not paused yet, will retry later")
+
+		return ctrl.Result{}, nil
+	}
+
+	// Check if the synchronizedGeneration matches the old authoritativeAPI's resource current generation.
+	if isSynchronizedGenMatchingOldAuthority, err := r.isSynchronizedGenerationMatchingOldAuthoritativeAPIGeneration(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to check synchronizedGeneration matches old authority: %w", err)
+	} else if !isSynchronizedGenMatchingOldAuthority {
+		// The to-be Authoritative API resource is not fully synced up yet, requeue to check later.
+		logger.Info("Authoritative machine and its copy are not synchronized yet, will retry later")
+
+		return ctrl.Result{}, nil
+	}
+
+	// Set the actual AuthoritativeAPI to the desired one, reset the synchronized generation and condition.
+	if err := r.setNewAuthoritativeAPIAndResetSynchronized(ctx, mapiMachine, mapiMachine.Spec.AuthoritativeAPI); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to apply authoritativeAPI to status with patch: %w", err)
+	}
+
+	// Make sure the new authoritative resource has been requested to unpause.
+	if err := r.ensureUnpauseRequestedOnNewAuthoritativeResource(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to ensure the new AuthoritativeAPI has been un-paused: %w", err)
+	}
+
+	logger.Info("Machine authority switch has now been completed and the resource unpaused")
+	logger.Info("Machine migrated successfully")
+
+	return ctrl.Result{}, nil
+}
+
+// isOldAuthoritativeResourcePaused checks whether the old authoritative resource is paused.
+func (r *MachineMigrationReconciler) isOldAuthoritativeResourcePaused(ctx context.Context, m *machinev1beta1.Machine) (bool, error) {
+	if m.Spec.AuthoritativeAPI == machinev1beta1.MachineAuthorityClusterAPI {
+		cond, err := util.GetConditionStatus(m, "Paused")
+		if err != nil {
+			return false, fmt.Errorf("unable to get paused condition for %s/%s: %w", m.Namespace, m.Name, err)
+		}
+
+		return cond == corev1.ConditionTrue, nil
+	}
+
+	// For MachineAuthorityMachineAPI, check the corresponding CAPI resource.
+	capiMachine := &capiv1beta1.Machine{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: r.CAPINamespace, Name: m.Name}, capiMachine); err != nil {
+		return false, fmt.Errorf("failed to get Cluster API machine: %w", err)
+	}
+
+	machinePausedCondition := conditionsv1beta2.Get(capiMachine, capiv1beta1.PausedV1Beta2Condition)
+	if machinePausedCondition == nil {
+		return false, nil
+	}
+
+	infraMachineRef := capiMachine.Spec.InfrastructureRef
+
+	infraMachine, err := util.GetReferencedObject(ctx, r.Client, r.Scheme, infraMachineRef)
+	if err != nil {
+		return false, fmt.Errorf("failed to get Cluster API infra machine: %w", err)
+	}
+
+	infraMachinePausedConditionStatus, err := util.GetConditionStatus(infraMachine, capiv1beta1.PausedV1Beta2Condition)
+	if err != nil {
+		return false, fmt.Errorf("unable to get paused condition for %s/%s: %w", infraMachine.GetNamespace(), infraMachine.GetName(), err)
+	}
+
+	return (machinePausedCondition.Status == metav1.ConditionTrue) && (infraMachinePausedConditionStatus == corev1.ConditionTrue), nil
+}
+
+func (r *MachineMigrationReconciler) ensureUnpauseRequestedOnNewAuthoritativeResource(ctx context.Context, mapiMachine *machinev1beta1.Machine) error {
+	// Request that the new authoritative resource reconciliation is un-paused.
+	//nolint:wsl
+	switch mapiMachine.Spec.AuthoritativeAPI {
+	case machinev1beta1.MachineAuthorityClusterAPI:
+		// For requesting unpausing of a CAPI resource, remove the paused annotation on it.
+		// So check if the ClusterAPI resource has the paused annotation and if so remove it.
+		capiMachine := &capiv1beta1.Machine{}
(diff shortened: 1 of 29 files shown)


pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,391 @@
...
+	if err := r.ensurePausedRequestedOnOldAuthoritativeAPI(ctx, mapiMachineSet); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to ensure old AuthoritativeAPI has been requested to pause: %w", err)
+	}
+
+	// Check if the old authoritativeAPI resource is actually paused.
+	if hasPausedConditionTrue, err := r.isOldAuthoritativeAPIStatusConditionPaused(ctx, mapiMachineSet); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to check old AuthoritativeAPI has paused=true condition: %w", err)
+	} else if !hasPausedConditionTrue {
+		// The old Authoritative API resource is not paused yet, requeue to check later.
+		return ctrl.Result{Requeue: true}, nil
```
JoelSpeed, line 132: To consider, should this requeueAfter? How quickly do we expect this to loop?

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,391 @@
...
+		// The to-be Authoritative API resource is not fully synced up yet, requeue to check later.
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	// Check if the Synchronized condition is set to True.
+	// if it is not, this indicates an unmigratable resource and therefore we should error.
+	if getConditionStatus(mapiMachineSet, consts.SynchronizedCondition) != corev1.ConditionTrue {
+		// The Resource is NOT Synchronized, error.
+		return ctrl.Result{}, errors.New("unable to proceed migrating, resources are not synchronized")
+	}
```
JoelSpeed, line 148: We check this twice?

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,391 @@
...
+
+	// Remove finalizer from the old authoritative API.
+	// This will ensure no status changes on the same reconcile.
+	// The finalizer must be removed from the object before we take any actions.
+	if removedFinalizer, err := r.ensureFinalizerRemovedOnOldAuthoritativeAPI(ctx, mapiMachineSet); err != nil {
+		return ctrl.Result{}, fmt.Errorf("error removing finalizer: %w", err)
+	} else if removedFinalizer {
+		return ctrl.Result{Requeue: true}, nil
+	}
```
JoelSpeed, line 166: I wasn't expecting finalizer cleanup to be handled here, but in the sync controller deletion logic. I think having the finalizer on the resource here shouldn't be an issue AFAICT

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,391 @@
...
+		// So Check if the ClusterAPI resource has the paused annotation, otherwise set it.
+		capiMachineSet := &capiv1beta1.MachineSet{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: consts.DefaultManagedNamespace, Name: mapiMachineSet.Name}, capiMachineSet); err != nil {
+			return fmt.Errorf("failed to get CAPI machine set: %w", err)
+		}
+
+		if !annotations.HasPaused(capiMachineSet) {
+			capiMachineSetCopy := capiMachineSet.DeepCopy()
+			annotations.AddAnnotations(capiMachineSet, map[string]string{capiv1beta1.PausedAnnotation: ""})
+			if err := r.Patch(ctx, capiMachineSet, client.MergeFrom(capiMachineSetCopy)); err != nil {
```
JoelSpeed, line 267: Do we need to requeue after making a change like this?

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,391 @@
...
+		if c.Type == condition {
+			return c.Status
+		}
+	}
+
+	return corev1.ConditionUnknown
+}
+
+func (r *MachineSetMigrationReconciler) isSynchronizedGenerationMatchingAuthoritativeAPIGeneration(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet) (bool, error) {
+	// Request that the old authoritative resource reconciliation is paused.
```
JoelSpeed, line 290: Might want to explain why the checks appear reversed initially, took me a moment to realise

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	// Check if we need to add the finalizer.
+	for _, finalizer := range newAuthoritativeResource.GetFinalizers() {
+		if finalizer == migrationControllerFinalizer {
+			return false, nil
+		}
+	}
+
+	newAuthoritativeResource.SetFinalizers(append(newAuthoritativeResource.GetFinalizers(), migrationControllerFinalizer))
```
JoelSpeed, line 353: I think in terms of finalizers, the migration controller is supposed to ensure that if the MAPI resource doesn't have the MAPI finalizer, it adds it, and if the CAPI resource doesn't have the expected CAPI finalizer, it adds that.

I think the sync controller has its own finalizer that it is supposed to add to resources to enable deletion synchronisation, but I don't think a migration specific finalizer is needed

JoelSpeed (commented):
Looked at this only from a high level at the moment, broad strokes this is what I'm expecting to be implemented, need to check the finalizers behaviour though as I don't think it's currently what I had in mind for this controller re finalizer involvement

pkg/controllers/machinesetmigration/machineset_migration_controller.go
damdo: We discussed this offline, there are two kinds of finalizers, the CAPI/MAPI own ones and the new sync finalizer.
This is to ensure the first kind is set or removed accordingly, as described in https://issues.redhat.com/browse/OCPCLOUD-2564 migration controller behaviours
```
- Add appropriate Finalizer to new authoritative API, requeue here until we observe this in cache
- Remove Finalizer from old authoritative API, requeue here until we observe this in cache
```

pkg/controllers/machinesetmigration/machineset_migration_controller.go
damdo: Yes, that'd be better, added.

pkg/controllers/machinesetmigration/machineset_migration_controller.go
damdo: Yes, done.

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+	if err != nil {
+		return false, fmt.Errorf("unable to get paused condition for %s/%s: %w", infraMachine.GetNamespace(), infraMachine.GetName(), err)
+	}
+
+	return (machinePausedCondition.Status == metav1.ConditionTrue) && (infraMachinePausedConditionStatus == corev1.ConditionTrue), nil
+}
+
+func (r *MachineMigrationReconciler) ensureUnpauseRequestedOnNewAuthoritativeResource(ctx context.Context, mapiMachine *machinev1beta1.Machine) error {
+	// Request that the new authoritative resource reconciliation is un-paused.
+	//nolint:wsl,exhaustive
```
JoelSpeed, line 242: Why do these not matter?

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+				return fmt.Errorf("failed to patch Cluster API machine: %w", err)
+			}
+		}
+
+		if annotations.HasPaused(infraMachine) {
+			infraMachineCopy, ok := infraMachine.DeepCopyObject().(client.Object)
+			if !ok {
+				return fmt.Errorf("unable to assert Cluster API infra machine as client.Object: %w", err)
+			}
+			util.RemoveAnnotation(infraMachine, capiv1beta1.PausedAnnotation)
```
JoelSpeed, line 273: Why util here when delete is used on 261? What's the difference?

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+// ensureFinalizerOnNewAuthoritativeResource adds a finalizer to the resource if required.
+// If the finalizer already exists, this function should be a no-op.
+// If the finalizer is added, the function will return true so that the reconciler can requeue the object.
+// Adding the finalizer in a separate reconcile ensures that spec updates are separate from status updates.
+func (r *MachineMigrationReconciler) ensureFinalizerOnNewAuthoritativeResource(ctx context.Context, m *machinev1beta1.Machine) (bool, error) {
+	if m.Spec.AuthoritativeAPI != machinev1beta1.MachineAuthorityClusterAPI {
+		return false, nil
+	}
```
JoelSpeed, line 424: Why not ensure the MAPI finalizer when going the other way?

We probably just want to make sure the finalizers are on both sides at all times

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+		return false, fmt.Errorf("failed to ensure finalizer on Cluster API machine: %w", err)
+	}
+
+	return updated || infraMachineUpdated, nil
+}
+
+// ensureNoFinalizerOnOldAuthoritativeResource removes the finalizer from the resource if required.
+// If the finalizer doesn't exists, this function should be a no-op.
+// If the finalizer gets removed, the function will return true so that the reconciler can requeue the object.
+func (r *MachineMigrationReconciler) ensureNoFinalizerOnOldAuthoritativeResource(ctx context.Context, m *machinev1beta1.Machine) (bool, error) {
```
JoelSpeed, line 457: I think this is to be removed right?

pkg/controllers/machinemigration/suite_test.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var testScheme *runtime.Scheme
+var testRESTMapper meta.RESTMapper
+var ctx = context.Background()
+
+func TestAPIs(t *testing.T) {
```
JoelSpeed, line 55: APIs?

pkg/util/annotations.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright 2024 Red Hat, Inc.
```
JoelSpeed, line 2: 2025

pkg/controllers/machinesetmigration/suite_test.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var testScheme *runtime.Scheme
+var testRESTMapper meta.RESTMapper
+var ctx = context.Background()
+
+func TestAPIs(t *testing.T) {
```
JoelSpeed, line 55: APIs?

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
+
+	// Remove finalizer from the old authoritative API, this ensures no status changes on the same reconcile.
+	if removed, err := r.ensureNoFinalizerOnOldAuthoritativeResource(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from old resource: %w", err)
+	} else if removed {
+		// Wait for it to take effect.
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
```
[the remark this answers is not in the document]
damdo: Yes this predates that conversation, will work on a change in this PR.

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+		// Check again later.
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
+
+	// Check if the Synchronized condition is set to True.
+	// If it is not, this indicates an unmigratable resource and therefore should take no action.
+	if cond, err := util.GetConditionStatus(mapiMachine, string(consts.SynchronizedCondition)); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to get synchronizedCondition for %s/%s: %w", mapiMachine.Namespace, mapiMachine.Name, err)
+	} else if cond != corev1.ConditionTrue {
+		logger.Info("New machine not yet in sync with the old authoritative one, will retry later")
```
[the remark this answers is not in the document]
damdo: Here old authoritative refers to the currently authoritative one and "new" means to-be authoritative.
The latter needs to be in-sync with the former for us to continue migrating.

I know the wording is confusing, I am improving it.

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+	if isSynchronizedGenMatchingOldAuthority, err := r.isSynchronizedGenerationMatchingOldAuthoritativeAPIGeneration(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to check synchronizedGeneration matches old authority: %w", err)
+	} else if !isSynchronizedGenMatchingOldAuthority {
+		// The to-be Authoritative API resource is not fully synced up yet, requeue to check later.
+		logger.Info("To-be authoritative and old machine are not synced yet, will retry later")
+
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
+
+	logger.Info("New machine is now in sync with the old authoritative one")
```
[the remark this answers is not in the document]
damdo: Yes, when I wrote them I meant them to be temporary so it could help the development, I'll make them silent by default.

pkg/controllers/machinemigration/machine_migration_controller.go
damdo: `exhaustive`

The only other entry for the switch on `mapiMachine.Spec.AuthoritativeAPI` is Migrating, which is disallowed to be set at the spec level, only at the status level (I covered it with a comment in `default`)

https://github.com/openshift/api/blob/8fcc4e71758aefadc7bf7da355ca8619bcc1d34d/machine/v1beta1/types_machine.go#L269


`wsl`
is complaining because some `case` are empty, but I want to explicitly keep them so we can add the comment there.

pkg/controllers/machinemigration/machine_migration_controller.go
damdo: `delete` only works on maps and their keys.

the `infraMachine` is a client.Object and we don't have direct access to its map, so the util gets it, does the delete and sets back.

pkg/controllers/machinemigration/machine_migration_controller.go
damdo: Yes

damdo:
/hold

I'll need to clean up git history once the PR has been approved.

pkg/controllers/machinemigration/machine_migration_controller.go
JoelSpeed: `client.Object` has `GetAnnotations()` doesn't it? So `delete(infraMachine.GetAnnotations(), capiv1beta1.PausedAnnotation)` should work here

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,356 @@
...
+
+	logger.V(1).Info("Reconciling machine set")
+	defer logger.V(1).Info("Finished reconciling machine set")
+
+	mapiMachineSet := &machinev1beta1.MachineSet{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, mapiMachineSet); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get MAPI machine set: %w", err)
+	}
+
+	logger.WithValues("machine set", mapiMachineSet.Name)
```
JoelSpeed, line 93: Duplicates `name` on L82?

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -0,0 +1,356 @@
...
+		// Check ClusterAPI generation against the synchronized one.
+		capiMachineSet := &capiv1beta1.MachineSet{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: r.CAPINamespace, Name: mapiMachineSet.Name}, capiMachineSet); err != nil {
+			return false, fmt.Errorf("failed to get Cluster API machine set: %w", err)
+		}
+
+		// Given the CAPI infra machine template is immutable
+		// we do not check for its generation to be synced up with the generation of the MAPI machine set.
+		return (mapiMachineSet.Status.SynchronizedGeneration == capiMachineSet.Generation), nil
+	default:
```
JoelSpeed, line 304: Is this possible? It would allow you to remove exhaustive and also help if we ever did add new values, since exhaustive would point out we haven't made a choice for the new value
```suggestion
	case machinev1beta1.Migrating, default:
```

pkg/controllers/machinemigration/machine_migration_controller_test.go
```diff
@@ -0,0 +1,693 @@
...
+		}
+
+		k = komega.New(k8sClient)
+	})
+
+	AfterEach(func() {
+		By("Cleaning up MAPI test resources")
+		testutils.CleanupResources(Default, ctx, cfg, k8sClient, mapiNamespace.GetName(),
+			&machinev1beta1.Machine{},
+			&machinev1beta1.Machine{},
```
JoelSpeed, line 130: Machine twice?

pkg/controllers/machinemigration/machine_migration_controller_test.go
```diff
@@ -0,0 +1,693 @@
...
+			&configv1.Infrastructure{},
+		)
+		testutils.CleanupResources(Default, ctx, cfg, k8sClient, capiNamespace.GetName(),
+			&capiv1beta1.Machine{},
+			&capiv1beta1.Machine{},
```
JoelSpeed, line 135: Machine twice?

pkg/controllers/machinemigration/machine_migration_controller_test.go
```diff
@@ -0,0 +1,693 @@
...
+							Status:             corev1.ConditionFalse}}).
+						Build()
+					mapiMachine.Status.AuthoritativeAPI = updatedMAPIMachine.Status.AuthoritativeAPI
+					mapiMachine.Status.Conditions = updatedMAPIMachine.Status.Conditions
+				})).Should(Succeed())
+
+				req = reconcile.Request{NamespacedName: client.ObjectKeyFromObject(mapiMachine)}
+			})
+
+			It("should requeue for later processing", func() {
```
JoelSpeed, line 216: Should we check the machine hasn't been written to? Resource version hasn't changed?

pkg/controllers/machinemigration/machine_migration_controller_test.go
```diff
@@ -0,0 +1,693 @@
...
+				_, err := reconciler.Reconcile(ctx, req)
+				Expect(err).NotTo(HaveOccurred(), "reconciler should not have errored")
+			})
+		})
+
+		Context("when a migration is request is first detected", func() {
```
JoelSpeed, line 222:
```suggestion
		Context("when a migration request is first detected", func() {
```

pkg/controllers/machinemigration/machine_migration_controller_test.go
```diff
@@ -0,0 +1,693 @@
...
+
+				It("should request pausing for the old authoritative resource (CAPI) and stay in Migrating status", func() {
+					_, err := reconciler.Reconcile(ctx, req)
+					Expect(err).NotTo(HaveOccurred())
+
+					updatedM := &machinev1beta1.Machine{}
+					Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(mapiMachine), updatedM)).To(Succeed())
+					Expect(updatedM.Status.AuthoritativeAPI).To(Equal(machinev1beta1.MachineAuthorityMigrating))
+
+					updatedCAPIMS := &capiv1beta1.Machine{}
```
JoelSpeed, line 330: MS?

Should we check the infra machine too?

pkg/controllers/machinesetmigration/machineset_migration_controller_test.go
```diff
@@ -0,0 +1,666 @@
...
+				req = reconcile.Request{NamespacedName: client.ObjectKeyFromObject(mapiMachineSet)}
+			})
+
+			It("should requeue for later processing", func() {
+				_, err := reconciler.Reconcile(ctx, req)
+				Expect(err).NotTo(HaveOccurred(), "reconciler should not have errored")
+			})
+		})
+
+		Context("when a migration is request is first detected", func() {
```
JoelSpeed, line 222:
```suggestion
		Context("when a migration request is first detected", func() {
```

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,394 @@
...
+
+	logger.V(1).Info("Reconciling machine")
+	defer logger.V(1).Info("Finished reconciling machine")
+
+	mapiMachine := &machinev1beta1.Machine{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get MAPI machine: %w", err)
+	}
+
+	logger.WithValues("machine", mapiMachine.Name)
```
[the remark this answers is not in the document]
damdo: Ah nice catch, removed

pkg/controllers/machinesetmigration/machineset_migration_controller.go
damdo: Dropped the `default` and added that one.

pkg/controllers/machinemigration/machine_migration_controller_test.go
damdo: Search/Replace gone wrong, one should be MachineSet

pkg/controllers/machinemigration/machine_migration_controller_test.go
damdo: Search/Replace gone wrong, one should be MachineSet

pkg/controllers/machinemigration/machine_migration_controller_test.go
damdo: Yes we should, added.

pkg/controllers/machinemigration/machine_migration_controller_test.go
damdo: In the migration controllers we don't have logic that changes the spec, only the metadata (annotations) and the status. So we would never update the ResourceVersion right? In any case

JoelSpeed: Resource version updates any time a resource is written, doesn't matter whether it's spec/status/metadata

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -0,0 +1,492 @@
...
+	}
+
+	logger.Info("Acknowledged migration request for machine")
+
+	// Request pausing on the old authoritative resource.
+	if updated, err := r.requestOldAuthoritativeResourcePaused(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to request pause on old authoritative machine: %w", err)
+	} else if updated {
+		// Wait for it to take effect.
+		return ctrl.Result{RequeueAfter: time.Second}, nil
```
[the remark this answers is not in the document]
damdo: Addressed this with an out of band chat, kept this and added a comment on why

pkg/controllers/machinemigration/machine_migration_controller_test.go
damdo: I misremembered!
Added Expects to cover these where appropriate.

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -274,8 +274,8 @@ func (r *MachineSetMigrationReconciler) requestOldAuthoritativeResourcePaused(ct
 		// The only provider we are interested in, that reconciles the inframachinetemplate, is ibmcloud/powervs
 		// which only updates its status
 		// see: https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/blob/main/controllers/ibmpowervsmachinetemplate_controller.go
-	default:
-		// Any other value is disallowed by the openAPI schema validation.
+	case machinev1beta1.MachineAuthorityMigrating:
```
JoelSpeed, line 277: Please make sure the machine side of this change gets the same update

damdo: Ofc, thanks for catching it. Updated.

damdo:
This code is not running in techpreview or stable at the moment, only in custom no upgrade, also I've seen things passing already in the previous test runs, I'll override them in the interest of time.

/override ci/prow/e2e-aws-capi-techpreview ci/prow/e2e-aws-ovn ci/prow/e2e-azure-capi-techpreview ci/prow/e2e-gcp-capi-techpreview ci/prow/e2e-gcp-ovn-techpreview ci/prow/e2e-openstack-capi-techpreview ci/prow/e2e-vsphere-capi-techpreview

PR #271, opened 2025-03-14
OCPCLOUD-2716: Handle Machine owner references translation between MAPI and CAPI

RadekManak:
This PR adds support to MachineSync and MachineSetSync controller for converting OwnerReferences in mirrored MachineSets, Machines and InfraMachines have correct ownerReferences.

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -36,10 +36,14 @@ import (
 
 	"github.com/go-test/deep"
 	machinev1applyconfigs "github.com/openshift/client-go/machine/applyconfigurations/machine/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
 	awscapiv1beta1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	capibmv1 "sigs.k8s.io/cluster-api-provider-ibmcloud/api/v1beta2"
 	capiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -66,6 +70,12 @@ var (
 
 	// errUnrecognizedConditionStatus is returned when the condition status is not recognized.
 	errUnrecognizedConditionStatus = errors.New("error unrecognized condition status")
+
+	// errUnsuportedOwnerKindForConversion is returned when the owner kind is not supported for conversion.
+	errUnsuportedOwnerKindForConversion = errors.New("unsupported owner kind for conversion")
+
+	// errMachineAPIMachineSetOwnerReferenceConversionUnsupported.
+	errMachineAPIMachineSetOwnerReferenceConversionUnsupported = errors.New("could not convert Machine API machine set owner references to Cluster API")
 )
 
 const (
@@ -251,6 +261,20 @@ func (r *MachineSetSyncReconciler) syncMachineSets(ctx context.Context, mapiMach
 func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
+	if err := r.validateMAPIMachineSetOwnerReferences(mapiMachineSet); err != nil {
+		if condErr := r.applySynchronizedConditionWithPatch(
+			ctx, mapiMachineSet, corev1.ConditionFalse, reasonFailedToConvertCAPIMachineSetToMAPI, err.Error(), nil); condErr != nil {
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{err, condErr})
+		}
+
+		if errors.Is(err, errMachineAPIMachineSetOwnerReferenceConversionUnsupported) {
+			logger.Error(err, "unable to convert Machine API machine set to Cluster API, owner references conversion is not supported for machine set")
+			return ctrl.Result{}, nil
+		} else {
+			return ctrl.Result{}, fmt.Errorf("failed to convert Machine API machine set owner references to Cluster API: %w", err)
+		}
+	}
+
 	newCAPIMachineSet, newCAPIInfraMachineTemplate, warns, err := r.convertMAPIToCAPIMachineSet(mapiMachineSet)
 	if err != nil {
 		conversionErr := fmt.Errorf("failed to convert MAPI machine set to CAPI machine set: %w", err)
@@ -270,6 +294,27 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 	newCAPIMachineSet.SetNamespace(r.CAPINamespace)
 	newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace = r.CAPINamespace
 
+	clusterOwnerRefence, err := r.fetchCAPIClusterOwnerReference(ctx)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get Cluster API cluster owner reference: %w", err)
+	}
+
+	newCAPIMachineSet.OwnerReferences = []metav1.OwnerReference{clusterOwnerRefence}
+
+	if result, err := r.ensureCAPIInfraMachineTemplate(ctx, mapiMachineSet, newCAPIMachineSet, newCAPIInfraMachineTemplate, clusterOwnerRefence); err != nil {
+		return result, fmt.Errorf("unable to ensure CAPI infra machine template: %w", err)
+	}
+
+	if result, err := r.createOrUpdateCAPIMachineSet(ctx, mapiMachineSet, capiMachineSet, newCAPIMachineSet); err != nil {
+		return result, fmt.Errorf("unable to ensure CAPI machine set: %w", err)
+	}
+
+	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachineSet, corev1.ConditionTrue,
+		consts.ReasonResourceSynchronized, messageSuccessfullySynchronizedMAPItoCAPI, &mapiMachineSet.Generation)
+}
+
+// ensureCAPIInfraMachineTemplate ensures the CAPI InfraMachineTemplate is created or updated from the MAPI MachineSet.
+func (r *MachineSetSyncReconciler) ensureCAPIInfraMachineTemplate(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, newCAPIMachineSet *capiv1beta1.MachineSet, newCAPIInfraMachineTemplate client.Object, clusterOwnerRefence metav1.OwnerReference) (ctrl.Result, error) {
 	_, infraMachineTemplate, err := r.fetchCAPIInfraResources(ctx, newCAPIMachineSet)
 	if err != nil && !apierrors.IsNotFound(err) {
 		fetchErr := fmt.Errorf("failed to fetch CAPI infra resources: %w", err)
@@ -284,23 +329,31 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 
 	newCAPIInfraMachineTemplate.SetResourceVersion(util.GetResourceVersion(infraMachineTemplate))
 	newCAPIInfraMachineTemplate.SetNamespace(r.CAPINamespace)
+	newCAPIInfraMachineTemplate.SetOwnerReferences([]metav1.OwnerReference{clusterOwnerRefence})
 
 	if result, err := r.createOrUpdateCAPIInfraMachineTemplate(ctx, mapiMachineSet, infraMachineTemplate, newCAPIInfraMachineTemplate); err != nil {
 		return result, fmt.Errorf("unable to ensure CAPI infra machine template: %w", err)
 	}
 
-	if result, err := r.createOrUpdateCAPIMachineSet(ctx, mapiMachineSet, capiMachineSet, newCAPIMachineSet); err != nil {
-		return result, fmt.Errorf("unable to ensure CAPI machine set: %w", err)
-	}
-
-	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachineSet, corev1.ConditionTrue,
-		consts.ReasonResourceSynchronized, messageSuccessfullySynchronizedMAPItoCAPI, &mapiMachineSet.Generation)
+	return ctrl.Result{}, nil
 }
 
 // reconcileCAPIMachineSetToMAPIMachineSet reconciles a CAPI MachineSet to a
 // MAPI MachineSet.
+//
+//nolint:funlen
 func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx context.Context, capiMachineSet *capiv1beta1.MachineSet, mapiMachineSet *machinev1beta1.MachineSet) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
+
+	if err := r.validateCAPIMachineSetOwnerReferences(capiMachineSet); err != nil {
+		logger.Error(err, "unable to convert Cluster API machine set to Machine API. Cluster API machine set has non-convertible owner references")
+
+		if condErr := r.applySynchronizedConditionWithPatch(ctx, mapiMachineSet, corev1.ConditionFalse, reasonFailedToConvertCAPIMachineSetToMAPI, err.Error(), nil); condErr != nil {
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{err, condErr})
+		}
+
+		return ctrl.Result{}, nil
+	}
 
 	infraCluster, infraMachineTemplate, err := r.fetchCAPIInfraResources(ctx, capiMachineSet)
 	if err != nil {
@@ -335,6 +388,7 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
 	newMapiMachineSet.SetNamespace(mapiMachineSet.GetNamespace())
 	// The conversion does not set a resource version, so we must copy it over
 	newMapiMachineSet.SetResourceVersion(util.GetResourceVersion(mapiMachineSet))
+	newMapiMachineSet.OwnerReferences = []metav1.OwnerReference{} // No CAPI machine set owner references are converted to MAPI machine set.
 
 	mapiMachineSetsDiff := compareMAPIMachineSets(mapiMachineSet, newMapiMachineSet)
 	if len(mapiMachineSetsDiff) > 0 {
@@ -360,6 +414,48 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
 
 	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachineSet, corev1.ConditionTrue,
 		consts.ReasonResourceSynchronized, messageSuccessfullySynchronizedCAPItoMAPI, &capiMachineSet.Generation)
+}
+
+// fetchCAPIClusterOwnerReference fetches the OpenShift cluster object instance and returns owner reference to it.
+// The OwnerReference has Controller set to false and BlockOwnerDeletion set to true.
+func (r *MachineSetSyncReconciler) fetchCAPIClusterOwnerReference(ctx context.Context) (metav1.OwnerReference, error) {
+	cluster := &capiv1beta1.Cluster{}
+	if err := r.Get(ctx, client.ObjectKey{Namespace: r.CAPINamespace, Name: r.Infra.Status.InfrastructureName}, cluster); err != nil {
+		return metav1.OwnerReference{}, fmt.Errorf("failed to get CAPI cluster: %w", err)
+	}
+
+	return metav1.OwnerReference{
+		APIVersion:         cluster.APIVersion,
+		Kind:               cluster.Kind,
+		Name:               cluster.Name,
+		UID:                cluster.UID,
+		Controller:         ptr.To(false),
+		BlockOwnerDeletion: ptr.To(true),
+	}, nil
+}
+
+// validateMAPIMachineSetOwnerReferences validates the owner references are allowed for conversion.
+func (r *MachineSetSyncReconciler) validateMAPIMachineSetOwnerReferences(mapiMachineSet *machinev1beta1.MachineSet) error {
+	if len(mapiMachineSet.OwnerReferences) > 0 {
+		return field.Invalid(field.NewPath("metadata", "ownerReferences"), mapiMachineSet.OwnerReferences, errMachineAPIMachineSetOwnerReferenceConversionUnsupported.Error())
+	}
+
+	return nil
+}
+
+// validateCAPIMachineSetOwnerReferences validates the owner references are allowed for conversion.
+func (r *MachineSetSyncReconciler) validateCAPIMachineSetOwnerReferences(capiMachineSet *capiv1beta1.MachineSet) error {
+	if len(capiMachineSet.OwnerReferences) > 1 {
+		return field.TooMany(field.NewPath("metadata", "ownerReferences"), len(capiMachineSet.OwnerReferences), 1)
+	} else if len(capiMachineSet.OwnerReferences) == 1 {
+		// Only reference to the Cluster is allowed.
+		ownerRef := capiMachineSet.OwnerReferences[0]
+		if ownerRef.Kind != capiv1beta1.ClusterKind || ownerRef.APIVersion != capiv1beta1.GroupVersion.String() {
+			return field.Invalid(field.NewPath("metadata", "ownerReferences"), capiMachineSet.OwnerReferences, errUnsuportedOwnerKindForConversion.Error())
+		}
+	}
+
+	return nil
 }
 
 // convertCAPIToMAPIMachineSet converts a CAPI MachineSet to a MAPI MachineSet, selecting the correct converter based on the platform.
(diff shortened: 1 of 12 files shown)


pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -270,6 +282,23 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 	newCAPIMachineSet.SetNamespace(r.CAPINamespace)
 	newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace = r.CAPINamespace
 
+	convertedCAPIMachineSetOwnerReferences, err := r.convertMAPIMachineSetOwnerReferencesToCAPI(ctx, mapiMachineSet)
```
damdo, line 285: Could we return something that is also usable later for newCAPIInfraMachineTemplate instead of only for newCAPIMachineSet?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -285,6 +314,20 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
...
+
+	newCAPIInfraMachineTemplate.SetOwnerReferences([]metav1.OwnerReference{{
+		APIVersion:         capiv1beta1.GroupVersion.String(),
+		Kind:               capiv1beta1.ClusterKind,
+		Name:               cluster.Name,
+		UID:                cluster.UID,
+		Controller:         ptr.To(false),
+		BlockOwnerDeletion: ptr.To(true),
+	}})
+
```
damdo, line 330: See above comment:

> Could we return something that is also usable later for newCAPIInfraMachineTemplate instead of only for newCAPIMachineSet?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -336,6 +381,17 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
 	// The conversion does not set a resource version, so we must copy it over
 	newMapiMachineSet.SetResourceVersion(util.GetResourceVersion(mapiMachineSet))
 
+	newMapiMachineSet.OwnerReferences, err = r.convertCAPIMachineSetOwnerReferencesToMAPI(capiMachineSet)
+	if err != nil {
+		logger.Error(err, "Failed to convert Cluster API machineSet owner references to Machine API")
```
damdo, line 386:
```suggestion
		logger.Error(err, "unable to convert Cluster API machine set to Machine API. Cluster API machine set has non-convertible owner references")
```

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -208,6 +226,16 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 					)).Should(Succeed())
 				})
 
+				It("should create MachineSet and MachineTemplate with CAPI Cluster OwnerReference", func() {
```
damdo, line 229:
```suggestion
				It("should create MachineSet and InfraMachineTemplate with CAPI Cluster OwnerReference", func() {
```

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -227,6 +255,19 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 					Expect(k8sClient.Create(ctx, capiMachineSet)).Should(Succeed())
 				})
 
+				It("should update MachineSet and MachineTemplate with CAPI Cluster OwnerReference", func() {
```
damdo, line 258:
```suggestion
				It("should update MachineSet and InfraMachineTemplate with CAPI Cluster OwnerReference", func() {
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -236,6 +247,21 @@ func (r *MachineSyncReconciler) reconcileMAPIMachinetoCAPIMachine(ctx context.Co
...
 
+	newCAPIMachine.OwnerReferences, err = r.convertMAPIMachineOwnerReferencesToCAPI(ctx, mapiMachine)
+	if err != nil {
+		if condErr := r.applySynchronizedConditionWithPatch(ctx, mapiMachine, corev1.ConditionFalse, reasonFailedToConvertMAPIMachineToCAPI, err.Error(), nil); condErr != nil {
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{err, condErr})
+		}
+
+		var fe *field.Error
+		if errors.As(err, &fe) || errors.Is(err, errUnsuportedOwnerKindForConversion) {
+			logger.Error(err, "Failed to convert Machine API machine owner references to Cluster API")
```
damdo, line 258: Let's change this to the same msg change as above.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -464,6 +501,106 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+		key := types.NamespacedName{
+			Namespace: r.CAPINamespace,
+			Name:      ownerReference.Name,
+		}
+
+		capiMachineSet := capiv1beta1.MachineSet{}
+		// Get the CAPI machineSet with same name as MAPI machineSet
+		if err := r.Get(ctx, key, &capiMachineSet); err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil, fmt.Errorf("cluster API MachineSet not found: %w", err)
```
damdo, line 533:
```suggestion
				return nil, fmt.Errorf("Cluster API machine set not found: %w", err)
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -464,6 +501,106 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+			} else {
+				return nil, fmt.Errorf("error getting MachineSet: %w", err)
```
damdo, line 535:
```suggestion
				return nil, fmt.Errorf("error getting Cluster API machine set: %w", err)
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -464,6 +501,106 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+		key := types.NamespacedName{
+			Namespace: r.MAPINamespace,
+			Name:      ownerReference.Name,
+		}
+
+		mapiMachineSet := machinev1beta1.MachineSet{}
+		// Get the MAPI machineSet with same name as CAPI machineSet
+		if err := r.Get(ctx, key, &mapiMachineSet); err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil, fmt.Errorf("machine API MachineSet not found: %w", err)
```
damdo, line 583:
```suggestion
				return nil, fmt.Errorf("Machine API machine set not found: %w", err)
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -464,6 +501,106 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+			} else {
+				return nil, fmt.Errorf("error getting MachineSet: %w", err)
```
damdo, line 585:
```suggestion
				return nil, fmt.Errorf("error getting Machine API machine set: %w", err)
```

damdo (commented):
Thanks for working on this!
Left a bunch of nitpicks but the main logic LGTM

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -497,9 +634,15 @@ func (r *MachineSyncReconciler) fetchCAPIInfraResources(ctx context.Context, cap
 
 // compareCAPIMachines compares CAPI machines a and b, and returns a list of differences, or none if there are none.
 func compareCAPIMachines(capiMachine1, capiMachine2 *capiv1beta1.Machine) []string {
+	capiMachine1Copy := capiMachine1.DeepCopy()
+	capiMachine2Copy := capiMachine2.DeepCopy()
+
+	capiMachine1Copy.OwnerReferences = nil
+	capiMachine2Copy.OwnerReferences = nil
+
```
damdo, line 642: So if these are always nil, how do we pick up the diff there?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -369,11 +441,35 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+					Eventually(k.Get(capiMachine)).Should(Succeed(), "should have succeeded getting a CAPI Machine")
+					Expect(capiMachine.OwnerReferences).To(Equal(ownerReferencesToCapiMachineSet))
+				})
+
 				It("should create the CAPI infra machine", func() {
 					capiInfraMachine := capav1builder.AWSMachine().WithName(mapiMachine.Name).WithNamespace(capiNamespace.Name).Build()
 					Eventually(k.Get(capiInfraMachine)).Should(Succeed(), "should have succeeded getting a CAPI Infra Machine")
 				})
 
+				It("should have OwnerReference from InfraMachine to Machine", func() {
```
damdo, line 455:
```suggestion
				It("should have InfraMachine's OwnerReference on the Machine", func() {
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -362,6 +418,48 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
...
+	} else if len(capiMachineSet.OwnerReferences) == 1 {
+		// Only reference to the Cluster is allowed.
+		ownerRef := capiMachineSet.OwnerReferences[0]
+		if ownerRef.Kind != capiv1beta1.ClusterKind || ownerRef.APIVersion != capiv1beta1.GroupVersion.String() {
+			return nil, field.Invalid(field.NewPath("metadata", "ownerReferences"), capiMachineSet.OwnerReferences, errUnsuportedOwnerKindForConversion.Error())
+		}
+	}
+
+	// MAPI MachineSet does not have owner references.
+	return []metav1.OwnerReference{}, nil
```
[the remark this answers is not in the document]
damdo: Personally I think the function name correctly captures what we are doing as a whole, but yes, I agree we should add a comment explaining what's happening inside the fuction.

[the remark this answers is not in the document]
theobarberbany: Yea, I think my naming gripe can be a non-blocking nit:  We're effectively unsetting OwnerReferences, right? Not converting? 

I think _convert_ might be a bit overzealous given we're only functional in one particularly specific scenario...

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -79,6 +85,9 @@ var (
 	errPlatformNotSupported = errors.New("error determining InfraMachine type, platform not supported")
 	// errUnrecognizedConditionStatus is returned when the condition status is not recognized.
 	errUnrecognizedConditionStatus = errors.New("error unrecognized condition status")
+
+	// errUnsuportedOwnerKindForConversion is returned when attempting to convert unsupported ownerReference.
+	errUnsuportedOwnerKindForConversion = errors.New("unsupported owner kind for ownerReference conversion")
```
JoelSpeed, line 90:
```suggestion
	errUnsuportedOwnerKindForConversion = errors.New("unsupported owner kind for owner reference conversion")
```

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -464,6 +501,106 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+	capiOwnerReferences := []metav1.OwnerReference{}
+
+	if len(mapiMachine.OwnerReferences) > 1 {
+		return nil, field.TooMany(field.NewPath("metadata", "ownerReferences"), len(mapiMachine.OwnerReferences), 1)
+	} else if len(mapiMachine.OwnerReferences) == 1 {
+		ownerRef := mapiMachine.OwnerReferences[0]
+		if ownerRef.Kind != machineSetKind || ownerRef.APIVersion != machinev1beta1.GroupVersion.String() {
+			return nil, field.Invalid(field.NewPath("metadata", "ownerReferences"), mapiMachine.OwnerReferences, errUnsuportedOwnerKindForConversion.Error())
+		}
+	}
```
JoelSpeed, line 517: This seems redundant when we have line 520 no?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -464,6 +501,106 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+	mapiOwnerReferences := []metav1.OwnerReference{}
+
+	if len(capiMachine.OwnerReferences) > 1 {
+		return nil, field.TooMany(field.NewPath("metadata", "ownerReferences"), len(capiMachine.OwnerReferences), 1)
+	} else if len(capiMachine.OwnerReferences) == 1 {
+		ownerRef := capiMachine.OwnerReferences[0]
+		if ownerRef.Kind != machineSetKind || ownerRef.APIVersion != capiv1beta1.GroupVersion.String() {
+			return nil, field.Invalid(field.NewPath("metadata", "ownerReferences"), capiMachine.OwnerReferences, errUnsuportedOwnerKindForConversion.Error())
+		}
+	}
```
JoelSpeed, line 567: Is this redundant considering 570?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -99,23 +108,73 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+		Expect(k8sClient.Create(ctx, mapiMachineSet)).Should(Succeed())
+		ownerReferencesToMapiMachineSet = []metav1.OwnerReference{{
+			APIVersion:         machinev1beta1.GroupVersion.String(),
+			Kind:               machineSetKind,
+			Name:               mapiMachineSet.Name,
+			UID:                mapiMachineSet.UID,
+			Controller:         ptr.To(true),
+			BlockOwnerDeletion: ptr.To(true),
+		},
+		}
```
JoelSpeed, line 143: Maybe this would be neater?
```suggestion
		ownerReferencesToMapiMachineSet = []metav1.OwnerReference{
			{
				APIVersion:         machinev1beta1.GroupVersion.String(),
				Kind:               machineSetKind,
				Name:               mapiMachineSet.Name,
				UID:                mapiMachineSet.UID,
				Controller:         ptr.To(true),
				BlockOwnerDeletion: ptr.To(true),
			},
		}
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -99,23 +108,73 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				},
+			},
+		}
+
+		// Create the CAPI MachineSet to have valid owner reference to it
+		capiMachineSet = capiMachineSetBuilder.
+			WithNamespace(capiNamespace.GetName()).WithName("foo-machineset").WithTemplate(capiMachineTemplate).WithClusterName(infrastructureName).
+			Build()
+		Expect(k8sClient.Create(ctx, capiMachineSet)).Should(Succeed())
+		ownerReferencesToCapiMachineSet = []metav1.OwnerReference{{
```
JoelSpeed, line 168: Indentation weird here too

pkg/controllers/machinesync/machine_sync_controller_test.go
JoelSpeed: Other way around now?

JoelSpeed (commented):
Does this also cover https://issues.redhat.com/browse/OCPCLOUD-2748? Do we need discussion about the use cases for MachineSet owner references?

damdo:
@RadekManak we had a chat with @JoelSpeed about whether this fully covers https://issues.redhat.com/browse/OCPCLOUD-2748 or not.

We think it doesn't fully, but that's fine. We'll come back to https://issues.redhat.com/browse/OCPCLOUD-2748 once we are in 4.20 land and add in/modify what's left to be done.

pkg/controllers/machinesync/machine_sync_controller.go
RadekManak: This was incorrect.  I think I needed to compare the owner references differently if there could be extra ones that we would not be converting. This was not the way to do It and is not required as the current iteration only allows machine set ownerreference conversion.

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
RadekManak: I renamed the function as it only handles validation now to reflect that.

RadekManak:
I think this is ready for re-review. 

[OCPCLOUD-2748](https://issues.redhat.com/browse/OCPCLOUD-2748) Is not fully covered in this PR. 
This PR only handles MAPI/CAPI ownerReferences that are known. Any usecases with third-party owner references should still be explored in that card.

damdo:
I just ran this locally and it looks like the happy path is working.
One thing, I noticed we get this log line:
```
E0326 16:24:36.046388   62135 machine_sync_controller.go:360] "unable to convert Machine API machine to Cluster API, unsupported owner reference in conversion" err="metadata.ownerReferences: Invalid value: []v1.OwnerReference{v1.OwnerReference{APIVersion:\"machine.openshift.io/v1\", Kind:\"ControlPlaneMachineSet\", Name:\"cluster\", UID:\"3927d542-7509-46ee-8d98-6367cf6fcac4\", Controller:(*bool)(0x140016cb0ea), BlockOwnerDeletion:(*bool)(0x140016cb0eb)}}: unsupported owner kind for owner reference conversion" controller="MachineSyncController" controllerGroup="machine.openshift.io" controllerKind="Machine" Machine="openshift-machine-api/ddonati-test916-j8fnc-master-2" namespace="openshift-machine-api" name="ddonati-test916-j8fnc-master-2" reconcileID="56d01f34-ce3e-426b-8abd-1c43291abd35"
```

Should we create a special case to not error/log for CPMS Machines, to avoid spamming the logs?
We can add a comment with a TODO once we add handling the conversion for those.

JoelSpeed:
> Should we create a special case to not error/log for CPMS Machines, to avoid spamming the logs?
We can add a comment with a TODO once we add handling the conversion for those.

Or, once we've determined that a Machine cannot be converted, we don't requeue? That means not returning the error to controller runtime so we will have to recognise when an error is a conversion error that isn't going to be fixed by requeueing

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -464,6 +499,102 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
...
+
+	capiOwnerReferences = append(capiOwnerReferences, capiOwnerReference)
+
+	return capiOwnerReferences, nil
+}
+
+// convertCAPIMachineOwnerReferencesToMAPI converts CAPI machine ownerReferences to MAPI ownerReferences.
+//
+//nolint:dupl
+func (r *MachineSyncReconciler) convertCAPIMachineOwnerReferencesToMAPI(ctx context.Context, capiMachine *capiv1beta1.Machine) ([]metav1.OwnerReference, error) { //nolint:unused
```
damdo, line 553: Is this used somewhere?

JoelSpeed:
I think it's valid to log when there's a conversion error. I don't think it's valid to requeue in that case. If we log the error, it should only happen when there's an update to the Machine right?

damdo:
> I think it's valid to log when there's a conversion error. I don't think it's valid to requeue in that case. If we log the error, it should only happen when there's an update to the Machine right?

Ok so let's only log.Error instead of erroring, in case of owner ref conversion errors.

For the CPMS I don't think there is value to us/the user to log.error when converting the CPMS Machines, those are out of scope for this migration so IMO we shouldn't log.error without a real reason.

JoelSpeed:
In the future they will be in scope though 🤔

damdo:
I don't see the need to log errors just because this might be possible in the future. But I am happy to yield if you think it is important.

damdo:
> Ok so let's only log.Error instead of erroring in any case of owner ref conversion errors.

@RadekManak as a matter of fact, I think we could do the owner reference check / conversion (`convertMAPIMachineOwnerReferencesToCAPI()`) before doing the whole `convertMAPIToCAPIMachine()`, so we bail early if we shouldn't/aren't convert the owner refs. Same for the MachineSet. WDYT?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -270,6 +280,42 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 	newCAPIMachineSet.SetNamespace(r.CAPINamespace)
 	newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace = r.CAPINamespace
 
+	err = r.validateMAPIMachineSetOwnerReferences(mapiMachineSet)
+	if err != nil {
```
damdo, line 284: nit, would be cleaner to do inline `if err:= func(); err !=nil {` here

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -336,6 +380,19 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
 	// The conversion does not set a resource version, so we must copy it over
 	newMapiMachineSet.SetResourceVersion(util.GetResourceVersion(mapiMachineSet))
 
+	err = r.validateCAPIMachineSetOwnerReferences(capiMachineSet)
+	if err != nil {
```
damdo, line 384: Let's inline the error here too

pkg/controllers/machinesync/machine_sync_controller.go
RadekManak: This will be called from reconcileCAPIMachinetoMAPIMachine whe implmeneted.

RadekManak:
Moved the owner references to be checked first.

I added new error for control plane machines. Instead of printing the generic error, It now info logs that the conversion is currently not supported. The condition will have the new error message. 

This should prevent any confusion as to why the conversion is failing and still keep the signal as to why the machine is not being mirrored into CAPI.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -214,9 +227,32 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
 
+	newCAPIOwnerReferences, err := r.convertMAPIMachineOwnerReferencesToCAPI(ctx, mapiMachine)
+	if err != nil {
+		if condErr := r.applySynchronizedConditionWithPatch(ctx, mapiMachine, corev1.ConditionFalse, reasonFailedToConvertMAPIMachineToCAPI, err.Error(), nil); condErr != nil {
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{err, condErr})
+		}
+
+		var fe *field.Error
+		if errors.As(err, &fe) || errors.Is(err, errUnsuportedOwnerKindForConversion) {
+			if fe.Detail == errUnsuportedOwnerKindForConversion.Error() {
```
damdo, line 243: Should this be `errUnsupportedCPMSOwnedMachineConversion`?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -214,9 +227,32 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
+//
+//nolint:funlen
 func (r *MachineSyncReconciler) reconcileMAPIMachinetoCAPIMachine(ctx context.Context, mapiMachine *machinev1beta1.Machine, capiMachine *capiv1beta1.Machine) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
+	newCAPIOwnerReferences, err := r.convertMAPIMachineOwnerReferencesToCAPI(ctx, mapiMachine)
+	if err != nil {
+		if condErr := r.applySynchronizedConditionWithPatch(ctx, mapiMachine, corev1.ConditionFalse, reasonFailedToConvertMAPIMachineToCAPI, err.Error(), nil); condErr != nil {
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{err, condErr})
+		}
```
damdo, line 239: This here causes a loop as it triggers re-reconciles also in cases where we don't want to, like in the `errUnsuportedOwnerKindForConversion` cases below.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -214,9 +227,32 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
+				logger.Info("Not converting %v Machine API machine. Conversion for machines owned by control plane machine set is currently not supported", mapiMachine.Name)
+				return ctrl.Result{}, nil
+			}
+
+			logger.Error(err, "unable to convert Machine API machine to Cluster API, unsupported owner reference in conversion")
+
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, fmt.Errorf("failed to convert Machine API machine owner references to Cluster API: %w", err)
```
damdo, line 253: I don't think we should ever error in case of failed owner reference conversion.
As we decided here.
https://github.com/openshift/cluster-capi-operator/pull/271#issuecomment-2754897624

damdo: This will otherwise cause infinite reconcile loops when, for example, MAPI MachineSets are owned by  more than one owner reference.

See for exampl case 
```
	if len(mapiMachine.OwnerReferences) > 1 {
```
in `convertMAPIMachineOwnerReferencesToCAPI()`.

RadekManak: This should now only retry when
	if err := r.Get(ctx, key, &capiMachineSet); err != nil {
fails.

The terminal field errors are handled above.

damdo (commented):
Thanks @RadekManak for addressing the feedback.
From what I can see all the feedback that @theobarberbany @JoelSpeed and I had was addressed

I also tested the changes locally and things were flowing correctly for the happy cases I rehearsed.

As I was mentioning to @theobarberbany on #265 we could probably to with a bit more in-depth testing especially on the various error paths, but we may want to coordinate and discuss that effort in combination with some coverage data, to help us better understand how to tackle that.

/approve
/lgtm

PR #272, opened 2025-03-14
NO-JIRA: Expand CAPI to Cluster API and MAPI to Machine API

RadekManak:
Expand CAPI to Cluster API and MAPI to Machine api in user facing strings. 

This should cover all occurrences, as I have generated this using sed a manually filtered test strings that don't have to follow this convention.

### DIFF
--- a/pkg/controllers/capiinstaller/capi_installer_controller.go
+++ b/pkg/controllers/capiinstaller/capi_installer_controller.go
@@ -123,7 +123,7 @@ func (r *CapiInstallerController) reconcile(ctx context.Context, log logr.Logger
 
 	// Process each one of the desired providers.
 	for providerConfigMapLabelTypeVal, providerConfigMapLabelNameVal := range providerConfigMapLabels {
-		log.Info("reconciling CAPI provider", "name", providerConfigMapLabelNameVal)
+		log.Info("reconciling Cluster API provider", "name", providerConfigMapLabelNameVal)
 
 		// Get a List all the ConfigMaps matching the desired provider labels.
 		configMapList := &corev1.ConfigMapList{}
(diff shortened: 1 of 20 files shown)


jeana-redhat (approved):
The expansion LGTM. There are other docs convention things that I would suggest changing, but I am guessing you'd prefer to keep commits sort of targeted to one topic. 

I guess I can collect other things I notice and maybe open a PR myself later suggestions? In any case, this gives me a very good idea at what attempting something like that would look like.

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -277,14 +277,14 @@ func (r *CapiInstallerController) setAvailableCondition(ctx context.Context, log
 
 	conds := []configv1.ClusterOperatorStatusCondition{
 		operatorstatus.NewClusterOperatorStatusCondition(capiInstallerControllerAvailableCondition, configv1.ConditionTrue, operatorstatus.ReasonAsExpected,
-			"CAPI Installer Controller works as expected"),
+			"Cluster API Installer Controller works as expected"),
```
damdo, line 280: All these `CAPI Installer Controller` should stay as is, as that's the actual controller name.

pkg/operatorstatus/operator_status.go
```diff
@@ -70,7 +70,7 @@ func (r *ClusterOperatorStatusClient) SetStatusAvailable(ctx context.Context, av
 	}
 
 	if availableConditionMsg == "" {
-		availableConditionMsg = fmt.Sprintf("Cluster CAPI Operator is available at %s", r.ReleaseVersion)
+		availableConditionMsg = fmt.Sprintf("Cluster Cluster API Operator is available at %s", r.ReleaseVersion)
```
damdo, line 73: This should not change

damdo (changes requested):
A couple of places should not be changed.
The rest looks good.

/unlgtm

JoelSpeed:
@RadekManak Could you please get this rebased? I think we should prio getting this through

PR #274, opened 2025-03-24
[WIP] OCPQE-28828: Created Cluster-api-tests-ext command for origin e2e

sunzhaohua2:


Dockerfile.rhel
```diff
@@ -1,11 +1,18 @@
...
 FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
 COPY --from=builder /go/src/github.com/openshift/cluster-capi-operator/bin/cluster-capi-operator .
 COPY --from=builder /go/src/github.com/openshift/cluster-capi-operator/bin/machine-api-migration .
 COPY --from=builder /go/src/github.com/openshift/cluster-capi-operator/manifests /manifests
+COPY --from=builder /tmp/build/cluster-api-tests-ext.gz .
 
 LABEL io.openshift.release.operator true
+LABEL io.k8s.display-name="OpenShift Cluster CAPI Operator" \
+    io.openshift.release.operator=true \
+    io.openshift.tags="openshift,tests,e2e,e2e-extension"
```
JoelSpeed, line 18: Please make sure files are finished with an empty new line

.golangci.yml
```diff
@@ -134,3 +134,16 @@ issues:
         - gochecknoglobals
         - err113
         - wrapcheck
+
+    # Exclude some linters from running on tests files.
+    - path: e2e/.*|e2e/framework/.*
```
theobarberbany, line 139: Hmm - do we not want to enforce the same coding standards across our e2es? I can understand excluding gosec.. not as sure about the others. 
cc @damdo

vr4manta:
Is there any desire for these existing tests to be included in the already defined suites provided by origin?  If we label the e2e tests using similar formats as origin (such as using [Suite:openshift/conformance/parallel] or [Suite:openshift/conformance/serial]) they will get included in those suites.  We can also add to all these tests any FeatureGate markers so they only run when certain feature gates are enabled.

sunzhaohua2:
> Is there any desire for these existing tests to be included in the already defined suites provided by origin? If we label the e2e tests using similar formats as origin (such as using [Suite:openshift/conformance/parallel] or [Suite:openshift/conformance/serial]) they will get included in those suites. We can also add to all these tests any FeatureGate markers so they only run when certain feature gates are enabled.

I think yes, right? @shellyyang1989 @damdo wdyt?

sunzhaohua2:
> there are many e2e changes that sound not relevant to OTE. Are they intended?

@shellyyang1989 yes, changing the files name from _test.go to without _test.go resulted in many new linter warnings. so to resolve these warnings, need these changes.

damdo:
> > > Is there any desire for these existing tests to be included in the already defined suites provided by origin? If we label the e2e tests using similar formats as origin (such as using [Suite:openshift/conformance/parallel] or [Suite:openshift/conformance/serial]) they will get included in those suites. We can also add to all these tests any FeatureGate markers so they only run when certain feature gates are enabled.
> > 
> > 
> > I think yes, right? @shellyyang1989 @damdo wdyt?
> 
> We need to be cautious about it I think. Per https://docs.google.com/document/d/1cFZj9QdzW8hbHc3H0Nce-2xrJMtpDJrwAse9H7hLiWk/edit?tab=t.0 the openshift/conformance contain highly reliable, fast running tests that run on a wide variety of configurations. Tests should be as fast as possible; and less than 5 minutes in duration and new tests must pass at >= 99%. Do we really meet the criteria?

I agree with @shellyyang1989 here, nice to have them compatible with OTE, but let's be cautious on where we add them for now :)

damdo:
/assign @mdbooth 

As you are already reviewing the other OTE PR

theobarberbany:
@coderabbitai review

PR #276, opened 2025-03-24
NO-JIRA: fix: e2e: azure conditionally set identity

damdo:
https://github.com/openshift/installer/pull/9538 introduced some changes in how MAPZ machines get generated.
So we need to account for it when we generate the CAPZ machine from those.

Discussion context here: https://redhat-internal.slack.com/archives/C05KZA3NVU6/p1742824263449689

### DIFF
--- a/e2e/azure_test.go
+++ b/e2e/azure_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


e2e/azure_test.go
```diff
@@ -101,13 +101,24 @@ func createAzureMachineTemplate(cl client.Client, mapiProviderSpec *mapiv1.Azure
 	Expect(err).To(BeNil(), "capz-manager-bootstrap-credentials secret should exist")
 	subscriptionID := azure_credentials_secret.Data["azure_subscription_id"]
 	azureImageID := fmt.Sprintf("/subscriptions/%s%s", subscriptionID, mapiProviderSpec.Image.ResourceID)
+
+	var (
+		identity               azurev1.VMIdentity
```
patrickdillon, line 106: I think we should default this to `VMIdentityNone`

PR #281, opened 2025-04-04
OCPCLOUD-2880: label selector and labels conversion, fixes to sync controllers and conversions

damdo:
This PR contains:
- assorted fixes to the sync controllers and the conversion libraries that I picked up while running the migration controllers doing full, real, back and forth MAPI->CAPI->MAPI migrations of machines and machine sets.
- full labels and annotations syncing between MAPI and CAPI, while also adding node labels and annotations propagation from CAPI machines down to their nodes
- machine sets selector bidirectional conversion
- improved diffing and diffs readability for machines and machine sets
- temporary hardcoded credentials secret conversion support for the default case to allow for migrations to momentarily work
- adds several `TODO(docs)` for lossy conversions that we might need to document (cc. @jeana-redhat)

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -43,7 +43,7 @@ import (
 	awscapiv1beta1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	capibmv1 "sigs.k8s.io/cluster-api-provider-ibmcloud/api/v1beta2"
 	capiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
-
+	"sigs.k8s.io/cluster-api/util/annotations"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -126,6 +126,7 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 
 	if err := ctrl.NewControllerManagedBy(mgr).
+		Named(controllerName).
 		For(&machinev1beta1.MachineSet{}, builder.WithPredicates(util.FilterNamespace(r.MAPINamespace))).
 		Watches(
 			&capiv1beta1.MachineSet{},
@@ -247,11 +248,11 @@ func (r *MachineSetSyncReconciler) syncMachineSets(ctx context.Context, mapiMach
 	case authoritativeAPI == machinev1beta1.MachineAuthorityClusterAPI && capiMachineSet != nil:
 		return r.reconcileCAPIMachineSetToMAPIMachineSet(ctx, capiMachineSet, mapiMachineSet)
 	case authoritativeAPI == machinev1beta1.MachineAuthorityMigrating:
-		logger.Info("machine set is currently being migrated")
+		logger.Info("Machine set is currently being migrated")
 		return ctrl.Result{}, nil
 
 	default:
-		logger.Info("unexpected value for authoritativeAPI", "AuthoritativeAPI", mapiMachineSet.Status.AuthoritativeAPI)
+		logger.Info("Unexpected value for authoritativeAPI", "AuthoritativeAPI", mapiMachineSet.Status.AuthoritativeAPI)
 
 		return ctrl.Result{}, nil
 	}
@@ -275,6 +276,11 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 		}
 	}
 
+	clusterOwnerRefence, err := r.fetchCAPIClusterOwnerReference(ctx)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get Cluster API cluster owner reference: %w", err)
+	}
+
 	newCAPIMachineSet, newCAPIInfraMachineTemplate, warns, err := r.convertMAPIToCAPIMachineSet(mapiMachineSet)
 	if err != nil {
 		conversionErr := fmt.Errorf("failed to convert MAPI machine set to CAPI machine set: %w", err)
@@ -290,16 +296,19 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 		r.Recorder.Event(mapiMachineSet, corev1.EventTypeWarning, "ConversionWarning", warning)
 	}
 
-	newCAPIMachineSet.SetResourceVersion(util.GetResourceVersion(client.Object(capiMachineSet)))
-	newCAPIMachineSet.SetNamespace(r.CAPINamespace)
-	newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace = r.CAPINamespace
-
-	clusterOwnerRefence, err := r.fetchCAPIClusterOwnerReference(ctx)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to get Cluster API cluster owner reference: %w", err)
+	if capiMachineSet != nil {
+		newCAPIMachineSet.SetGeneration(capiMachineSet.GetGeneration())
+		newCAPIMachineSet.SetUID(capiMachineSet.GetUID())
+		newCAPIMachineSet.SetCreationTimestamp(capiMachineSet.GetCreationTimestamp())
+		newCAPIMachineSet.SetManagedFields(capiMachineSet.GetManagedFields())
+		newCAPIMachineSet.SetResourceVersion(util.GetResourceVersion(client.Object(capiMachineSet)))
 	}
 
+	newCAPIMachineSet.SetNamespace(r.CAPINamespace)
+	newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace = r.CAPINamespace
 	newCAPIMachineSet.OwnerReferences = []metav1.OwnerReference{clusterOwnerRefence}
+	// Set the paused annotation on the new CAPI MachineSet, as we want to create it paused.
+	annotations.AddAnnotations(newCAPIMachineSet, map[string]string{capiv1beta1.PausedAnnotation: ""})
 
 	if result, err := r.ensureCAPIInfraMachineTemplate(ctx, mapiMachineSet, newCAPIMachineSet, newCAPIInfraMachineTemplate, clusterOwnerRefence); err != nil {
 		return result, fmt.Errorf("unable to ensure CAPI infra machine template: %w", err)
@@ -327,9 +336,15 @@ func (r *MachineSetSyncReconciler) ensureCAPIInfraMachineTemplate(ctx context.Co
 		return ctrl.Result{}, fetchErr
 	}
 
-	newCAPIInfraMachineTemplate.SetResourceVersion(util.GetResourceVersion(infraMachineTemplate))
+	if !util.IsNilObject(infraMachineTemplate) {
+		newCAPIInfraMachineTemplate.SetResourceVersion(util.GetResourceVersion(infraMachineTemplate))
+		newCAPIInfraMachineTemplate.SetFinalizers(infraMachineTemplate.GetFinalizers())
+	}
+
 	newCAPIInfraMachineTemplate.SetNamespace(r.CAPINamespace)
 	newCAPIInfraMachineTemplate.SetOwnerReferences([]metav1.OwnerReference{clusterOwnerRefence})
+	// Set the paused annotation on the new CAPI InfraMachineTemplate, as we want to create it paused.
+	annotations.AddAnnotations(newCAPIInfraMachineTemplate, map[string]string{capiv1beta1.PausedAnnotation: ""})
 
 	if result, err := r.createOrUpdateCAPIInfraMachineTemplate(ctx, mapiMachineSet, infraMachineTemplate, newCAPIInfraMachineTemplate); err != nil {
 		return result, fmt.Errorf("unable to ensure CAPI infra machine template: %w", err)
@@ -384,15 +399,28 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
 		r.Recorder.Event(mapiMachineSet, corev1.EventTypeWarning, "ConversionWarning", warning)
 	}
 
-	newMapiMachineSet.Spec.Template.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.Labels, newMapiMachineSet.Spec.Template.Labels)
+	newMapiMachineSet.SetGeneration(mapiMachineSet.GetGeneration())
+	newMapiMachineSet.SetUID(mapiMachineSet.GetUID())
+	newMapiMachineSet.SetCreationTimestamp(mapiMachineSet.GetCreationTimestamp())
+	newMapiMachineSet.SetManagedFields(mapiMachineSet.GetManagedFields())
+	newMapiMachineSet.SetResourceVersion(util.GetResourceVersion(client.Object(mapiMachineSet)))
 	newMapiMachineSet.SetNamespace(mapiMachineSet.GetNamespace())
-	// The conversion does not set a resource version, so we must copy it over
-	newMapiMachineSet.SetResourceVersion(util.GetResourceVersion(mapiMachineSet))
-	newMapiMachineSet.OwnerReferences = []metav1.OwnerReference{} // No CAPI machine set owner references are converted to MAPI machine set.
+	newMapiMachineSet.Spec.Template.ObjectMeta.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.ObjectMeta.Labels, newMapiMachineSet.Spec.Template.ObjectMeta.Labels)
+	newMapiMachineSet.Spec.Template.Spec.ObjectMeta.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.Spec.ObjectMeta.Labels, newMapiMachineSet.Spec.Template.Spec.ObjectMeta.Labels)
+	// Restore API authoritativeness, as it gets lost in MAPI->CAPI->MAPI translation.
+	newMapiMachineSet.Spec.AuthoritativeAPI = mapiMachineSet.Spec.AuthoritativeAPI
+	newMapiMachineSet.Spec.Template.Spec.AuthoritativeAPI = mapiMachineSet.Spec.Template.Spec.AuthoritativeAPI
+	// Restore the original MAPI selector as it is immutable.
+	newMapiMachineSet.Spec.Selector = mapiMachineSet.Spec.Selector
+	newMapiMachineSet.OwnerReferences = nil // No CAPI machine set owner references are converted to MAPI machine set.
 
-	mapiMachineSetsDiff := compareMAPIMachineSets(mapiMachineSet, newMapiMachineSet)
+	mapiMachineSetsDiff, err := compareMAPIMachineSets(mapiMachineSet, newMapiMachineSet)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to compare MAPI machine sets: %w", err)
+	}
+
 	if len(mapiMachineSetsDiff) > 0 {
-		logger.Info("Changes detected, updating MAPI machine set", "diff", mapiMachineSetsDiff)
+		logger.Info("Changes detected for MAPI machine set. Updating it", "diff", fmt.Sprintf("%+v", mapiMachineSetsDiff))
 
 		if err := r.Update(ctx, newMapiMachineSet); err != nil {
 			logger.Error(err, "Failed to update MAPI machine set")
@@ -409,7 +437,7 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
 
 		logger.Info("Successfully updated MAPI machine set")
 	} else {
-		logger.Info("No changes detected in MAPI machine set")
+		logger.Info("No changes detected for MAPI machine set")
 	}
 
 	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachineSet, corev1.ConditionTrue,
@@ -593,11 +621,11 @@ func (r *MachineSetSyncReconciler) createOrUpdateCAPIInfraMachineTemplate(ctx co
 	}
 
 	if len(capiInfraMachineTemplatesDiff) == 0 {
-		logger.Info("No changes detected in CAPI infra machine template")
+		logger.Info("No changes detected for CAPI infra machine template")
 		return ctrl.Result{}, nil
 	}
 
-	logger.Info("Changes detected, updating CAPI infra machine template", "diff", capiInfraMachineTemplatesDiff)
+	logger.Info("Changes detected for CAPI infra machine template. Updating it", "diff", fmt.Sprintf("%+v", capiInfraMachineTemplatesDiff))
 
 	if err := r.Update(ctx, newCAPIInfraMachineTemplate); err != nil {
 		logger.Error(err, "Failed to update CAPI infra machine template")
@@ -642,11 +670,11 @@ func (r *MachineSetSyncReconciler) createOrUpdateCAPIMachineSet(ctx context.Cont
 	capiMachineSetsDiff := compareCAPIMachineSets(capiMachineSet, newCAPIMachineSet)
 
 	if len(capiMachineSetsDiff) == 0 {
-		logger.Info("No changes detected in CAPI machine set")
+		logger.Info("No changes detected for CAPI machine set")
 		return ctrl.Result{}, nil
 	}
 
-	logger.Info("Changes detected, updating CAPI machine set", "diff", capiMachineSetsDiff)
+	logger.Info("Changes detected for CAPI machine set. Updating it", "diff", fmt.Sprintf("%+v", capiMachineSetsDiff))
 
 	if err := r.Update(ctx, newCAPIMachineSet); err != nil {
 		logger.Error(err, "Failed to update CAPI machine set")
@@ -681,7 +709,7 @@ func initInfraMachineTemplateAndInfraClusterFromProvider(platform configv1.Platf
 }
 
 // compareCAPIInfraMachineTemplates compares CAPI infra machine templates a and b, and returns a list of differences, or none if there are none.
-func compareCAPIInfraMachineTemplates(platform configv1.PlatformType, infraMachineTemplate1, infraMachineTemplate2 client.Object) ([]string, error) {
+func compareCAPIInfraMachineTemplates(platform configv1.PlatformType, infraMachineTemplate1, infraMachineTemplate2 client.Object) (map[string]any, error) {
 	switch platform {
 	case configv1.AWSPlatformType:
 		typedInfraMachineTemplate1, ok := infraMachineTemplate1.(*awscapiv1beta1.AWSMachineTemplate)
@@ -694,9 +722,15 @@ func compareCAPIInfraMachineTemplates(platform configv1.PlatformType, infraMachi
 			return nil, errAssertingCAPIAWSMachineTemplate
 		}
 
-		var diff []string
-		diff = append(diff, deep.Equal(typedInfraMachineTemplate1.Spec, typedinfraMachineTemplate2.Spec)...)
-		diff = append(diff, util.ObjectMetaEqual(typedInfraMachineTemplate1.ObjectMeta, typedinfraMachineTemplate2.ObjectMeta)...)
+		diff := make(map[string]any)
+
+		if diffSpec := deep.Equal(typedInfraMachineTemplate1.Spec, typedinfraMachineTemplate2.Spec); len(diffSpec) > 0 {
+			diff[".spec"] = diffSpec
+		}
+
+		if diffObjectMeta := util.ObjectMetaEqual(typedInfraMachineTemplate1.ObjectMeta, typedinfraMachineTemplate2.ObjectMeta); len(diffObjectMeta) > 0 {
+			diff[".metadata"] = diffObjectMeta
+		}
 
 		return diff, nil
 	case configv1.PowerVSPlatformType:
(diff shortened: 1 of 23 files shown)


pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -384,15 +399,28 @@ func (r *MachineSetSyncReconciler) reconcileCAPIMachineSetToMAPIMachineSet(ctx c
...
+	newMapiMachineSet.SetGeneration(mapiMachineSet.GetGeneration())
+	newMapiMachineSet.SetUID(mapiMachineSet.GetUID())
+	newMapiMachineSet.SetCreationTimestamp(mapiMachineSet.GetCreationTimestamp())
+	newMapiMachineSet.SetManagedFields(mapiMachineSet.GetManagedFields())
+	newMapiMachineSet.SetResourceVersion(util.GetResourceVersion(client.Object(mapiMachineSet)))
 	newMapiMachineSet.SetNamespace(mapiMachineSet.GetNamespace())
-	// The conversion does not set a resource version, so we must copy it over
-	newMapiMachineSet.SetResourceVersion(util.GetResourceVersion(mapiMachineSet))
-	newMapiMachineSet.OwnerReferences = []metav1.OwnerReference{} // No CAPI machine set owner references are converted to MAPI machine set.
+	newMapiMachineSet.Spec.Template.ObjectMeta.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.ObjectMeta.Labels, newMapiMachineSet.Spec.Template.ObjectMeta.Labels)
```
theobarberbany, line 408: I can't remember, did we decide that we want CAPI labels to be propagated to the MAPI counterpart?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -710,9 +744,15 @@ func compareCAPIInfraMachineTemplates(platform configv1.PlatformType, infraMachi
...
-		diff = append(diff, deep.Equal(typedInfraMachineTemplate1.Spec, typedinfraMachineTemplate2.Spec)...)
-		diff = append(diff, util.ObjectMetaEqual(typedInfraMachineTemplate1.ObjectMeta, typedinfraMachineTemplate2.ObjectMeta)...)
+		diff := make(map[string]any)
+
+		if diffSpec := deep.Equal(typedInfraMachineTemplate1.Spec, typedinfraMachineTemplate2.Spec); len(diffSpec) > 0 {
+			diff[".spec"] = diffSpec
+		}
+
+		if diffObjectMeta := deep.Equal(typedInfraMachineTemplate1.ObjectMeta, typedinfraMachineTemplate2.ObjectMeta); len(diffObjectMeta) > 0 {
+			diff[".metadata"] = diffObjectMeta
```
theobarberbany, line 754: how does this print? do we `%+v` the keys? I found the diff confusing when I was trying to debug some tests.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -315,22 +318,28 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
 	}
 
+	newMapiMachine.SetNamespace(r.MAPINamespace)
+	newMapiMachine.SetOwnerReferences(newMAPIOwnerReferences)
+
 	if mapiMachine != nil {
 		newMapiMachine.SetResourceVersion(util.GetResourceVersion(mapiMachine))
-		newMapiMachine.Spec.Labels = util.MergeMaps(mapiMachine.Spec.Labels, newMapiMachine.Spec.Labels)
-		newMapiMachine.Labels = util.MergeMaps(mapiMachine.Labels, newMapiMachine.Labels)
+		// newMapiMachine.Spec.Labels = util.MergeMaps(mapiMachine.Spec.Labels, newMapiMachine.Spec.Labels)
```
theobarberbany, line 331: Do we want a TODO to work this out later? Why have we stopped this with labels? :)

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -402,12 +407,35 @@ func (r *MachineSyncReconciler) reconcileMAPIMachinetoCAPIMachine(ctx context.Co
...
+		// Restore finalizers.
+		newCAPIMachine.SetFinalizers(capiMachine.GetFinalizers())
+	}
+
 	newCAPIMachine.SetNamespace(r.CAPINamespace)
 	newCAPIMachine.Spec.InfrastructureRef.Namespace = r.CAPINamespace
-
 	newCAPIMachine.OwnerReferences = newCAPIOwnerReferences
 
+	if len(newCAPIMachine.OwnerReferences) > 0 {
```
theobarberbany, line 429: 🎉nice catch :)

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -673,6 +714,9 @@ func (r *MachineSyncReconciler) shouldMirrorCAPIMachineToMAPIMachine(ctx context
 	logger.V(4).WithName("shouldMirrorCAPIMachineToMAPIMachine").
 		Info("Checking if Cluster API machine should be mirrored", "machine", machine.GetName())
 
+	// TODO: should we check if the CAPI Machine is being deleted and do not bother creating a MAPI Machine mirror?
```
theobarberbany, line 717: I think it's a good idea :)

pkg/conversion/capi2mapi/common.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func convertCAPIMachineSetSelectorToMAPI(mapiSelector metav1.LabelSelector) metav1.LabelSelector {
+	capiSelector := mapiSelector
+	capiSelector.MatchLabels = convertCAPILabelsToMAPILabels(mapiSelector.MatchLabels)
+
+	return capiSelector
+}
+
+func convertCAPILabelsToMAPILabels(capiLabels map[string]string) map[string]string {
```
theobarberbany, line 51: Do we want unit tests for these? (forgive me if I've missed them!)

pkg/conversion/mapi2capi/machine.go
```diff
@@ -17,113 +17,66 @@ limitations under the License.
...
-	if len(spec.Taints) > 0 {
-		errs = append(errs, field.Invalid(fldPath.Child("taints"), spec.Taints, "taints are not currently supported"))
+	capiMachine := &capiv1.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            mapiMachine.Name,
+			Namespace:       capiNamespace,
+			Labels:          convertMAPILabelsToCAPI(mapiMachine.Labels),
+			Annotations:     convertMAPIAnnotationsToCAPI(mapiMachine.Annotations),
+			OwnerReferences: nil, // OwnerReferences not populated here. They are added later by the machineSync controller.
+			// Finalizers: // Finalizers not populated here. They are added later by the machine controllers.
```
JoelSpeed, line 45: Why are these different? Why is one set to nil and the other just commented?

pkg/conversion/mapi2capi/util.go
```diff
@@ -17,72 +17,163 @@ limitations under the License.
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-// fromMAPIMachineToCAPIMachine translates a MAPI Machine to its Core CAPI Machine correspondent.
-func fromMAPIMachineToCAPIMachine(mapiMachine *mapiv1beta1.Machine, apiVersion, kind string) (*capiv1.Machine, field.ErrorList) {
-	var errs field.ErrorList
+func convertMAPIMachineSetSelectorToCAPI(mapiSelector metav1.LabelSelector) metav1.LabelSelector {
+	capiSelector := mapiSelector
```
JoelSpeed, line 29: Unsafe copy, use deepcopy

pkg/util/sync.go
```diff
@@ -55,12 +55,26 @@ func hasSameState(i *machinev1beta1.Condition, j *machinev1applyconfigs.Conditio
...
+
+	if diffLabels := deep.Equal(a.Labels, b.Labels); len(diffLabels) > 0 {
+		objectMetaDiff[".labels"] = diffLabels
+	}
+
+	if diffAnnotations := deep.Equal(a.Annotations, b.Annotations); len(diffAnnotations) > 0 {
+		objectMetaDiff[".annotations"] = diffAnnotations
+	}
+
+	// TODO: uncomment once the issue where CAPI controllers add the finalizer even on paused resources is fixed.
```
RadekManak, line 69: Is there a card we can link here?

pkg/conversion/mapi2capi/util.go
```diff
@@ -17,72 +17,163 @@ limitations under the License.
...
+	capiLabels := make(map[string]string)
 
-	capiMachine := &capiv1.Machine{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:            mapiMachine.Name,
-			Namespace:       capiNamespace,
-			Labels:          mapiMachine.Labels,
-			Annotations:     mapiMachine.Annotations,
-			OwnerReferences: nil, // OwnerReferences not populated here. They are added later by the machineSync controller.
+	toConvertLabels := map[string]string{
```
JoelSpeed, line 38: What's the difference between convert and transform?

pkg/conversion/mapi2capi/util.go
```diff
@@ -17,72 +17,163 @@ limitations under the License.
...
+			capiLabelKey, capiLabelVal := transformFunc(mapiLabelVal)
+			capiLabels[capiLabelKey] = capiLabelVal
+
+			continue
+		}
+
+		// Step 3: ignore MAPI-specific labels that are not explicitly handled.
+		// if strings.HasPrefix(mapiLabelKey, "machine.openshift.io/") {
+		// 	continue
+		// }
```
JoelSpeed, line 70: Why is this one commented out currently?

pkg/conversion/mapi2capi/util.go
```diff
@@ -17,72 +17,163 @@ limitations under the License.
...
+
+		// Step 4: default case - copy over the label as-is to CAPI.
+		capiLabels[mapiLabelKey] = mapiLabelVal
+	}
+
+	return capiLabels
+}
+
+func convertMAPIAnnotationsToCAPI(mapiAnnotations map[string]string) map[string]string {
+	if len(mapiAnnotations) == 0 || mapiAnnotations == nil {
```
JoelSpeed, line 80: The second case cannot be reached

pkg/conversion/mapi2capi/util.go
```diff
@@ -17,72 +17,163 @@ limitations under the License.
...
+
+	for k, v := range mapiAnnotations {
+		capiAnnotations[k] = v
+	}
+
+	return capiAnnotations
+}
+
+func setMAPINodeAnnotationsToCAPINodeAnnotations(mapiNodeAnnotations map[string]string, capiMachine *capiv1.Machine) {
+	if len(mapiNodeAnnotations) == 0 || mapiNodeAnnotations == nil {
```
JoelSpeed, line 94: Second case cannot be reached

pkg/conversion/mapi2capi/util.go
```diff
@@ -17,72 +17,163 @@ limitations under the License.
...
+		capiMachine.Annotations = map[string]string{}
+	}
+
+	for k, v := range mapiNodeAnnotations {
+		capiMachine.Annotations[k] = v
+	}
+}
+
+func setMAPINodeLabelsToCAPINodeLabels(mapiNodeLabels map[string]string, capiMachine *capiv1.Machine) {
+	if len(mapiNodeLabels) == 0 || mapiNodeLabels == nil {
```
JoelSpeed, line 108: Second case cannot be reached

pkg/conversion/test/fuzz/fuzz.go
```diff
@@ -263,10 +270,28 @@ func MAPI2CAPIMachineRoundTripFuzzTest(scheme *runtime.Scheme, infra *configv1.I
...
-		Expect(mapiMachine.Spec).To(WithTransform(ignoreMachineProviderSpec, testutils.MatchViaJSON(ignoreMachineProviderSpec(in.machine.Spec))))
-		Expect(mapiMachine.Spec.ProviderSpec.Value.Raw).To(MatchJSON(in.machine.Spec.ProviderSpec.Value.Raw))
+		// Annotations propagation happens differently in MAPI vs CAPI, so the roundtrip is lossy.
+		// We document this in the migration docs.
+		Expect(mapiMachine.ObjectMeta.Annotations).To(equalMap(util.MergeMaps(in.machine.ObjectMeta.Annotations, in.machine.Spec.ObjectMeta.Annotations)), "converted MAPI machine should have matching .metadata.annotations")
+		Expect(mapiMachine.Spec.ObjectMeta.Annotations).To(equalMap(util.MergeMaps(in.machine.ObjectMeta.Annotations, in.machine.Spec.ObjectMeta.Annotations)), "converted MAPI machine should have matching .spec.annotations")
+		mapiMachine.ObjectMeta.Annotations = nil
+		mapiMachine.Spec.ObjectMeta.Annotations = nil
+		in.machine.ObjectMeta.Annotations = nil
+		in.machine.Spec.ObjectMeta.Annotations = nil
```
JoelSpeed, line 280: Where we have done this before, we have fixed the issue at the fuzzer level. So where the two maps need to be the same on both sides, we could handle this in the custom fuzzer, and we don't need special logic changes here. That may be better as it'll be consistent with how we are handling other cases (in the fuzzer vs in the test checks)

pkg/controllers/machinesetsync/machineset_sync_controller.go
damdo: Kinda

```
%v	the value in a default format
	when printing structs, the plus flag (%+v) adds field names
```, we print field names. also with the new map[string] approach we can prefix the field name when we break down the diff checks like in this case with .metadata.

File: pkg/controllers/machinesync/machine_sync_controller.go (hunk 5 above)
Comment (damdo) [reply pkg/controllers/machinesync/machine_sync_controller.go:717]: Yeah we can deal with this in the deletion logic efforts PR

File: pkg/conversion/capi2mapi/common.go (hunk 6 above)
Comment (damdo) [reply pkg/conversion/capi2mapi/common.go:51]: The fuzzy tests should fully cover the conversion expectations

File: pkg/controllers/machinesetsync/machineset_sync_controller.go [hunk 16]
```diff
@@ -327,9 +336,15 @@ func (r *MachineSetSyncReconciler) ensureCAPIInfraMachineTemplate(ctx context.Co
 		return ctrl.Result{}, fetchErr
 	}
 
-	newCAPIInfraMachineTemplate.SetResourceVersion(util.GetResourceVersion(infraMachineTemplate))
+	if !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
damdo: Moved to a check for the client.Object to be nil/empty.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -619,7 +653,7 @@ func (r *MachineSyncReconciler) createOrUpdateMAPIMachine(ctx context.Context, m
 	if mapiMachine == nil {
 		if err := r.Create(ctx, newMAPIMachine); err != nil {
 			logger.Error(err, "Failed to create Machine API machine")
-			return ctrl.Result{}, fmt.Errorf("failed to create Machine API machine: %w", err)
+			return ctrl.Result{}, fmt.Errorf("failed to create MAPI machine: %w", err)
```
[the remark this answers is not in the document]
damdo: Typo

pkg/util/sync.go
damdo: Not really, with @JoelSpeed we decided this is not an issue as we can work around that. I'll remove it

pkg/util/sync.go
```diff
@@ -55,12 +55,26 @@ func hasSameState(i *machinev1beta1.Condition, j *machinev1applyconfigs.Conditio
...
+	// https://github.com/kubernetes-sigs/cluster-api/blob/c70dca0fc387b44457c69b71a719132a0d9bed58/internal/controllers/machine/machine_controller.go#L207-L210
```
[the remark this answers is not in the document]
damdo: I left this commented code in intentionally so we would talk again about this during the review, like we are now and figure out the best strategy.

Yes I think it is fine for us to ignore the differences in finalizers

> Should the sync controller just convert the MAPI machine finalizer to CAPI and vice versa so they are always on both sides?

Not sure about this, we can still do it in the migration controllers but ignore setting/removing the CAPI one.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -247,8 +250,8 @@ func (r *MachineSyncReconciler) Reconcile(ctx context.Context, req reconcile.Req
 
 // reconcileCAPIMachinetoMAPIMachine reconciles a CAPI Machine to a MAPI Machine.
 //
-//nolint:gocognit
-func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Context, capiMachine *capiv1beta1.Machine, mapiMachine *machinev1beta1.Machine) (ctrl.Result, error) { //nolint:funlen
+//nolint:gocognit,funlen
```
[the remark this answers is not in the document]
damdo: I'll look at doing this, I didn't want to have a too big of a diff at first.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -402,12 +407,35 @@ func (r *MachineSyncReconciler) reconcileMAPIMachinetoCAPIMachine(ctx context.Co
...
 	newCAPIMachine.SetNamespace(r.CAPINamespace)
 	newCAPIMachine.Spec.InfrastructureRef.Namespace = r.CAPINamespace
-
 	newCAPIMachine.OwnerReferences = newCAPIOwnerReferences
 
+	if len(newCAPIMachine.OwnerReferences) > 0 {
+		// For CAPI Machine that is owned by a CAPI MachineSet we must set the capiv1beta1.MachineSetNameLabel
+		// as this is what the CAPI machineset controller sets on the CAPI Machine when it creates it, an it is then later used
+		// by other CAPI tooling for filtering purposes.
+		newCAPIMachine.Labels[capiv1beta1.MachineSetNameLabel] = format.MustFormatValue(newCAPIMachine.OwnerReferences[0].Name)
```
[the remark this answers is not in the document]
damdo: Pretty confident, as this is enforced and converted from MAPI just above via: `convertMAPIMachineOwnerReferencesToCAPI()`.

pkg/conversion/capi2mapi/common.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling providerSpec: %w", err)
+	}
+
+	return &runtime.RawExtension{
+		Raw: rawBytes,
+	}, nil
+}
+
+func convertCAPIMachineSetSelectorToMAPI(mapiSelector metav1.LabelSelector) metav1.LabelSelector {
```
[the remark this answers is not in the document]
damdo: Good catch

pkg/conversion/capi2mapi/common.go
```diff
@@ -0,0 +1,150 @@
...
+	capiSelector := mapiSelector
+	capiSelector.MatchLabels = convertCAPILabelsToMAPILabels(mapiSelector.MatchLabels)
+
+	return capiSelector
+}
+
+func convertCAPILabelsToMAPILabels(capiLabels map[string]string) map[string]string {
+	if len(capiLabels) == 0 || capiLabels == nil {
```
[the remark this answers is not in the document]
damdo: Yes I have removed this already in various places. Forgot about this.

pkg/conversion/capi2mapi/common.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, fmt.Errorf("error marshalling providerSpec: %w", err)
+	}
+
+	return &runtime.RawExtension{
+		Raw: rawBytes,
+	}, nil
+}
+
+func convertCAPIMachineSetSelectorToMAPI(mapiSelector metav1.LabelSelector) metav1.LabelSelector {
+	capiSelector := mapiSelector
```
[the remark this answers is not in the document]
damdo: Ah yes, that'd be best!

pkg/conversion/mapi2capi/util.go
damdo: Not much at the moment, I'll drop the toConvert.

pkg/conversion/mapi2capi/machine.go
damdo: Added in different iterations by different people, I'll switch to nil if it doesn't cause issues.

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -402,12 +407,35 @@ func (r *MachineSyncReconciler) reconcileMAPIMachinetoCAPIMachine(ctx context.Co
...
+	}
+
+	// Set the paused annotation on the new CAPI Machine, as we want to create it paused.
+	annotations.AddAnnotations(newCAPIMachine, map[string]string{capiv1beta1.PausedAnnotation: ""})
```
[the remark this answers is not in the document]
damdo: This line is in `reconcileMAPIMachinetoCAPIMachine()` which is only going to get called if the machine is MAPI authoritative:
```
	switch mapiMachine.Status.AuthoritativeAPI {
	case machinev1beta1.MachineAuthorityMachineAPI
```

pkg/util/sync.go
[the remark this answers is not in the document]
JoelSpeed: If CAPI is always setting the finalizer, even when paused, then I think it makes sense for the sync controller to worry about finalizers, and then the migration controller doesn't need to. We will just always have the finalizers and deletion logic in the sync controller will handle their tear down

pkg/controllers/machinesync/machine_sync_controller.go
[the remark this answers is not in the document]
JoelSpeed: Might be worth explaining why we are confident in a comment. Or making this less fragile somehow in case someone changes this in the future

pkg/util/sync.go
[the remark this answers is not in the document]
damdo: SGTM

pkg/conversion/mapi2capi/machine_test.go
```diff
@@ -86,17 +86,6 @@ var _ = Describe("mapi2capi Machine conversion", func() {
 			expectedWarnings: []string{},
 		}),
 
-		Entry("With unsupported non-CAPI managed labels", mapi2CAPIMachineConversionInput{
```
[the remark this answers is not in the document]
damdo: This is a negative test that covered making sure we error on unsupported node labels.
Now that we support them in conversion, the fuzzy test fully covers checking we copy them over correctly.
That said we could add more tests as a follow up PR, I'd like @theobarberbany and I to add some more testing on the sync controllers too.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: This wasn't required to achieve the updating the mirror machine, as we do this copying in the conversion. I'll remove the comment.

pkg/controllers/machinesetsync/machineset_sync_controller.go
damdo: But now we might even be able to drop this. I'll check.

pkg/util/sync.go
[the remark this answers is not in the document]
damdo: Added a TODO note on #268 
I'll mark this as resolved here.

pkg/conversion/mapi2capi/util.go
damdo: In my tests I've seen that we might need to copy these over to CAPI.
We are using them as labelSelectors in some cases (see snippet below)
and if we don't copy them over to CAPI, when we create the new MAPI machine set mirror from the CAPI (authoritative) one, we can't restore them at the machine template level, losing the ability to select them with the ones defined in the label selector.
Unless we always
```
newMapiMachine.Labels = util.MergeMaps(mapiMachine.Labels, newMapiMachine.Labels)
```
and
```
newMapiMachineSet.Spec.Template.ObjectMeta.Labels = util.MergeMaps(mapiMachineSet.Spec.Template.ObjectMeta.Labels, newMapiMachineSet.Spec.Template.ObjectMeta.Labels)
```
but if these gets deleted things might get weird.

MAPI MachineSet snippet:
```
spec:
  authoritativeAPI: MachineAPI
  replicas: 1
  selector:
    matchLabels:
      machine.openshift.io/cluster-api-cluster: ddonati-test928-t544j
      machine.openshift.io/cluster-api-machineset: ddonati-test928-t544j-worker-eu-south-1a
  template:
    metadata:
      labels:
        machine.openshift.io/cluster-api-cluster: ddonati-test928-t544j
        machine.openshift.io/cluster-api-machine-role: worker
        machine.openshift.io/cluster-api-machine-type: worker
        machine.openshift.io/cluster-api-machineset: ddonati-test928-t544j-worker-eu-south-1a
    spec:
      authoritativeAPI: MachineAPI
      lifecycleHooks: {}
      metadata: {}
```

JoelSpeed: Right, so we have to make sure that the selector matchlabels are always present in the labels being created on the machine. And so right now, basically we are copying them all over?

Should we translate them? What do we expect them to be?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -721,19 +761,52 @@ func compareCAPIInfraMachineTemplates(platform configv1.PlatformType, infraMachi
...
+		return nil, fmt.Errorf("unable to parse first MAPI machine set providerSpec: %w", err)
+	}
+
+	ps2, err := mapi2capi.AWSProviderSpecFromRawExtension(a.Spec.Template.Spec.ProviderSpec.Value)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse second MAPI machine set providerSpec: %w", err)
+	}
+
+	if diffProviderSpec := deep.Equal(ps1, ps2); len(diffProviderSpec) > 0 {
+		diff[".providerSpec"] = diffProviderSpec
```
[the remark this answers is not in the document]
damdo: Yes this is not perfect, still is better than the existing diff as with that there are no keys and it is not possible to understand where the diff is coming from.

For now it makes things understandable.
We can make this better for sure, maybe a TODO for the follow up?

pkg/conversion/test/fuzz/fuzz.go
JoelSpeed: I don't think fuzzing is the right place to be testing something so nuanced, we should have specific tests.

The purpose of the fuzz tests predominantly is to pick up when new fields are added as we vendor changes, or where we have gaps in conversion as we introduce new platforms.

I think it's totally acceptable to say "we must have the same labels in two places in MAPI as the source of a fuzz" so that the fuzzing can focus on it's role, and have separate tests that handle the complexity of the labels conversion elsewhere

Do we have those separate tests right now?

damdo: If we are going to add separate testing for labels then yeah, I think what you are saying makes sense.
If not, then we probably shouldn't.

But yes, having specific testing for this is probably going to be wiser.

pkg/conversion/mapi2capi/machine_test.go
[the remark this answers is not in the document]
damdo: This falls under the same conversation as https://github.com/openshift/cluster-capi-operator/pull/281#discussion_r2033306356

Closing this in favour of that one.

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
damdo: Yeah I agree it is not ideal, but we probably should not get it perfect now.
Let's make a note to iterate over it in a separate PR the stabilisation period. WDYT?

pkg/conversion/test/fuzz/fuzz.go
damdo: If you agree, I can update the fuzzer like you said in this PR, and then parallelise the work on conversion wide label testing in a follow up PR. WDYT?

JoelSpeed: Yep, lets update the fuzzer and follow up with specific label testing in units please

Please make a card to track

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
JoelSpeed: Yeah that's fine by me for now, lets not lose track of this though

JoelSpeed:
Changes in https://github.com/openshift/cluster-capi-operator/pull/281/commits/427c56f90535d05b23704800b000e3259523fd26 LGTM, so it's just fixing up the fuzzing to go?

damdo:
@JoelSpeed Ok fuzzing fixed.
PTAL

pkg/conversion/capi2mapi/aws.go
```diff
@@ -133,6 +133,13 @@ func (m machineAndAWSMachineAndAWSCluster) toProviderSpec() (*mapiv1.AWSMachineP
 		},
 		// UserDataSecret - Populated below.
 		// CredentialsSecret - TODO(OCPCLOUD-2713)
+		// TODO(hack): for CredentialsSecret. Remove when the conversion is in place.
+		// this is the name of the default credentials and is in place to get the base case to work
+		// while we develop for the migration.
+		CredentialsSecret: &corev1.LocalObjectReference{
+			Name: "aws-cloud-credentials",
+		},
```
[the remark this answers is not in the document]
nrb: https://github.com/openshift/cluster-capi-operator/pull/282

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -315,22 +318,28 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
 
+	newMapiMachine.SetNamespace(r.MAPINamespace)
+	newMapiMachine.SetOwnerReferences(newMAPIOwnerReferences)
+
 	if mapiMachine != nil {
 		newMapiMachine.SetResourceVersion(util.GetResourceVersion(mapiMachine))
-		newMapiMachine.Spec.Labels = util.MergeMaps(mapiMachine.Spec.Labels, newMapiMachine.Spec.Labels)
-		newMapiMachine.Labels = util.MergeMaps(mapiMachine.Labels, newMapiMachine.Labels)
+		// newMapiMachine.Spec.Labels = util.MergeMaps(mapiMachine.Spec.Labels, newMapiMachine.Spec.Labels)
+		// newMapiMachine.Labels = util.MergeMaps(mapiMachine.Labels, newMapiMachine.Labels)
```
JoelSpeed, line 332: These bits that are commented out, still needed?

pkg/controllers/machinesync/machine_sync_controller.go
JoelSpeed: Do we need to remove the comment here then?

pkg/conversion/test/fuzz/fuzz.go
```diff
@@ -440,6 +442,9 @@ func CAPIMachineFuzzerFuncs(providerIDFuzz StringFuzzer, infraKind, infraAPIVers
 			func(m *capiv1.Machine, c fuzz.Continue) {
 				c.FuzzNoCustom(m)
 
+				m.Annotations = sharedAnnotations()
+				m.Labels = sharedCAPILabels(clusterName)
+
```
JoelSpeed, line 447: You could fuzz these normally and then set the MachineTemplateSpec lower. The MachineTemplateSpec fuzzer has already run by this point. Then you'd still be fuzzing labels and would only test static labels/annotations

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Oops, removed.

pkg/controllers/machinesync/machine_sync_controller.go
damdo: Missed this, removed.

pkg/conversion/test/fuzz/fuzz.go
damdo: Yeah you are right, updated.

PR #285, opened 2025-04-23
NO-JIRA: Prevent caching CAPI objects across namespaces

JoelSpeed:
We have seen lots of caching (Watches) being established on namespaces we don't actually care about. By default, controller runtime will cache all objects we fetch for all of the default namespaces.

For our controllers, we typically care about the managed namespace, and only want to fetch certain objects from other namespaces.

This PR splits out the specific objects that we watch from other namespaces so that we only cache those objects in the other namespaces and not every object we care about

### DIFF
--- a/cmd/cluster-capi-operator/main.go
+++ b/cmd/cluster-capi-operator/main.go
@@ -22,7 +22,9 @@ import (
 
 	metal3v1 "github.com/metal3-io/cluster-api-provider-metal3/api/v1beta1"
 	"github.com/spf13/pflag"
+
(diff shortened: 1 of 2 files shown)


JoelSpeed:
The CAPI techpreview tests here are not looking so happy, something must be wrong with the new watch configs

JoelSpeed:
Operator looks like it's working as expected
/retest

JoelSpeed:
/retest

I'm not sure, are these tests reliable?

nrb:
/override ci/prow/okd-scos-e2e-aws-ovn 

This job is perma-failing, and not really relevant to this PR.

PR #286, opened 2025-04-24
 OCPCLOUD-2943: MachineSet deletion logic 

theobarberbany:
This PR adds deletion logic to the MachineSetSync controller. 

59bde20875989f6aecd8f610324f02652dd92823 This change sets the deletion finalizer on both MAPI and CAPI machinesets when reconciling both directions.

236c7ee8fc8d447d3b3056bcbaa9fd0e3e874d7f handles deletion when the MAPI MS is authoritative 

b4ffb7a73bbc0f6d83575a7bf067614c7787010f handles deletion when a CAPI MS is authoritative

b1d1a1407c9d6e723c35c7d1756a856a939a34dc  Allows removal of non-authoritative MAPI machinesets

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -736,10 +736,47 @@ func initInfraMachineTemplateAndInfraClusterFromProvider(platform configv1.Platf
 
 func (r *MachineSetSyncReconciler) reconcileMAPItoCAPIMachineSetDeletion(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet) (bool, error) {
 	if mapiMachineSet.DeletionTimestamp.IsZero() {
-		return false, nil
+		if capiMachineSet == nil || capiMachineSet.DeletionTimestamp.IsZero() {
+			// Neither MAPI authoritative machine set nor its CAPI non-authoritative mirror
+			// are being deleted, nothing to reconcile for deletion.
+			return false, nil
+		}
+
+		// The MAPI authoritative machine set is not being deleted, but the CAPI non-authoritative one is.
+		// Issue a deletion also to the MAPI authoritative machine set.
+		if err := r.Client.Delete(ctx, mapiMachineSet); err != nil {
+			return false, fmt.Errorf("failed to delete Machine API machine set: %w", err)
+		}
+
+		// Return true to force a requeue, to allow the deletion propagation.
+		return true, nil
 	}
 
-	return false, nil
+	logger := log.FromContext(ctx)
+
+	if capiMachineSet == nil {
+		logger.Info("Cluster API machine set does not exist, removing corresponding Machine API machine set sync finalizer")
+		// We don't have  a capi machine set to clean up. Just let the MAPI operators
+		// function as normal, and remove the MAPI sync finalizer.
+		_, err := util.RemoveFinalizer(ctx, r.Client, mapiMachineSet, machinesync.SyncFinalizer)
+
+		return true, fmt.Errorf("failed to remove finalizer from Machine API machine set: %w", err)
+	}
+
+	if capiMachineSet.DeletionTimestamp.IsZero() {
+		logger.Info("Machine API machine set is being deleted, issuing deletion to corresponding Cluster API machine set")
+
+		if err := r.Client.Delete(ctx, capiMachineSet); err != nil {
+			return true, fmt.Errorf("failed delete Cluster API machine set: %w", err)
+		}
+	}
+
+	// We'll re-reconcile and remove the MAPI machineset once the CAPI one is not present
+	if _, err := util.RemoveFinalizer(ctx, r.Client, capiMachineSet, machinesync.SyncFinalizer); err != nil {
+		return true, fmt.Errorf("failed to remove finalizer from Cluster API machine set: %w", err)
+	}
+
+	return true, nil
 }
 
 // compareCAPIInfraMachineTemplates compares CAPI infra machine templates a and b, and returns a list of differences, or none if there are none.
@@ -852,6 +889,8 @@ func copyCapiObjectMeta(capiMachineSet, newCAPIMachineSet *capiv1beta1.MachineSe
 		newCAPIMachineSet.SetCreationTimestamp(capiMachineSet.GetCreationTimestamp())
 		newCAPIMachineSet.SetManagedFields(capiMachineSet.GetManagedFields())
 		newCAPIMachineSet.SetResourceVersion(util.GetResourceVersion(client.Object(capiMachineSet)))
+		// Restore finalizers.
+		newCAPIMachineSet.SetFinalizers(capiMachineSet.GetFinalizers())
 	}
 
 	newCAPIMachineSet.SetNamespace(capiNamespace)
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -800,26 +800,27 @@ func (r *MachineSetSyncReconciler) reconcileCAPItoMAPIMachineSetDeletion(ctx con
 
 	if capiMachineSet.DeletionTimestamp.IsZero() {
 		if mapiMachineSet == nil || mapiMachineSet.DeletionTimestamp.IsZero() {
-			// Neither CAPI authoritative machine set nor its MAPI non-authoritative mirror
-			// are being deleted, nothing to reconcile for deletion.
+			// Neither CAPI authoritative machine set nor its MAPI non-authoritative mirror are being deleted, nothing to reconcile for deletion.
 			return false, nil
 		}
-		// The CAPI authoritative machine set is not being deleted, but the MAPI non-authoritative one is
-		// Issue a deletion also to the CAPI authoritative machine set.
-		logger.Info("The non-authoritative Machine API machine set is being deleted, issuing deletion to the corresponding Cluster API machine set")
+		// The CAPI authoritative machine set is not being deleted, but the MAPI non-authoritative one is remove our sync finalizer
+		// on the cluster api resources, and allow deletion of the MAPI machineset
+		logger.Info("The non-authoritative Machine API machine set is being deleted, removing our sync finalizer from the corresponding Cluster API machine set")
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -779,6 +790,70 @@ func (r *MachineSetSyncReconciler) reconcileMAPItoCAPIMachineSetDeletion(ctx con
...
+		logger.Info("Cluster API machine set is being deleted, issuing deletion to corresponding Machine API machine set")
+
+		if err := r.Client.Delete(ctx, mapiMachineSet); err != nil {
+			return true, fmt.Errorf("failed delete Machine API machine set: %w", err)
+		}
+
+		return true, nil
+	}
+
+	// TODO: I don't think we run CAPI  controllers, so this won't be present in unit testing.
```
theobarberbany, line 835: Is it ok that we don't test this in unit tests? 

We could run a goroutine somewhere that spots CAPI MachineSets, adds the finalizer and once it sees a deletion timestamp removes it some seconds later?

This doesn't test how our finalizer may interact with the CAPI one though.. (if at all)

JoelSpeed: The test suite should be able to handle the other finalizers yes, I'm sure we've mocked things like this before

I don't think you even need a controller to do it, you can just structure it as part of a specific test case "when the CAPI finalizer is removed" as perhaps a BeforeEach for some further expectations?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -279,6 +307,50 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+							HaveField("ObjectMeta.Finalizers", ContainElement(machinesync.SyncFinalizer)),
+						)
+						Expect(k8sClient.Delete(ctx, mapiMachineSet)).To(Succeed())
+					})
+					// Expect to see the finalizers, so they're in place before
+					//  we Expect logic that relies on them to work
+					It("should delete the CAPI machine set", func() {
+						// Does this need to be more specific, e.g explicitly be a 404?
+						// TODO: 404
+						Eventually(k.Get(capiMachineSet), timeout).ShouldNot(Succeed())
```
JoelSpeed, line 326: You should be able to check the error perhaps with the `apierrors.IsNotFound` check, as a transform? Have a look at how `MatchError` works in gomega as an alternative

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -370,7 +454,7 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 
 					It("should not populate the field", func() {
 						Eventually(k.Object(mapiMachineSet), timeout).Should(
-							HaveField("Finalizers", BeEmpty()),
+							HaveField("Finalizers", Not(ContainElements("foo", "bar"))),
```
JoelSpeed, line 457: Foo bar?

theobarberbany: Further up we populate `foo`:`bar`, and then expect no finalizers to be propagated. However, once we add deletion logic we rely on sync finalizers being in place.

 This keeps the meaning of the existing test :)

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -693,6 +694,38 @@ func (r *MachineSetSyncReconciler) createOrUpdateCAPIMachineSet(ctx context.Cont
...
 
+// ensureSyncFinalizer ensures the sync finalizer is present across mapi and capi machine sets.
+// It attempts to set both in one call, aggregating errors.
+func (r *MachineSetSyncReconciler) ensureSyncFinalizer(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, capiMachineSet *capiv1beta1.MachineSet) (bool, error) {
+	var shouldRequeue bool
+
+	var errors []error
+
+	if mapiMachineSet != nil {
+		if mapiMachineSet.DeletionTimestamp.IsZero() {
```
nrb, line 705: I haven't paid close attention to this work, but it looks like we're assuming that the deletion timestamp will always be set for both.

Is this field synced across, so that when the authoritative side is marked, the mirror is too?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -708,6 +741,123 @@ func initInfraMachineTemplateAndInfraClusterFromProvider(platform configv1.Platf
...
+	logger := log.FromContext(ctx)
+
+	if mapiMachineSet.DeletionTimestamp.IsZero() {
+		if capiMachineSet == nil || capiMachineSet.DeletionTimestamp.IsZero() {
+			// Neither MAPI authoritative machine set nor its CAPI non-authoritative mirror
+			// are being deleted, nothing to reconcile for deletion.
+			return false, nil
+		}
+
+		// The MAPI authoritative machine set is not being deleted, but the CAPI non-authoritative one is.
```
nrb, line 754: This is a little surprising to me. Is this a new edge case that was found during implementation?

I double checked the enhancement, I didn't see mention of deleting a mirror resource causing the authoritative one to be deleted.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -708,6 +741,123 @@ func initInfraMachineTemplateAndInfraClusterFromProvider(platform configv1.Platf
...
+		// Issue a deletion also to the MAPI authoritative machine set.
+		logger.Info("The non-authoritative Cluster API machine set is being deleted, issuing deletion to the corresponding Machine API machine set")
```
nrb, line 756: Does the logger we're making here have the machine set's name in a structured field?

nrb (commented):
Looks reasonable to me. Did have a few questions.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -279,6 +309,49 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+								HaveField("ObjectMeta.Finalizers", ContainElement(machinesync.SyncFinalizer)),
+								HaveField("ObjectMeta.Finalizers", ContainElement(capiv1beta1.MachineSetFinalizer)),
+							),
+						)
+						Expect(k8sClient.Delete(ctx, mapiMachineSet)).To(Succeed())
+					})
+					// Expect to see the finalizers, so they're in place before
+					//  we Expect logic that relies on them to work
+					It("should delete the CAPI machine set", func() {
+						Eventually(k.Get(capiMachineSet), timeout).Should(WithTransform(apierrors.IsNotFound, BeTrue()))
```
JoelSpeed, line 329: On anything that uses `BeTrue()` it's doubly important to add context, so that when the test fails, we don't just get `expected true to be false`
```suggestion
						Eventually(k.Get(capiMachineSet), timeout).Should(WithTransform(apierrors.IsNotFound, BeTrue()), "eventually capiMachineSet should not be found")
```

Can you do something like ^ on all the similar cases you've got here?

JoelSpeed (commented):
Re-reviewing the EP section on deletion this morning. When a MAPI non-authoritative MachineSet is deleted, it should not sync the deletion across to the CAPI machineset.

I believe the same is also expected for Machines, if the MAPI machine is not authoritative, and has no owner, its deletion should not be sync'd across.

Can you re-read https://github.com/openshift/enhancements/blob/master/enhancements/machine-api/converting-machine-api-to-cluster-api.md#deletion-of-machines and check you agree?

I don't think that's covered here, but what's here looks good, so maybe this should be a separate card?

pkg/controllers/machinesetsync/machineset_sync_controller.go
nrb: ```Machine API 	- 	Delete 	Yes 	No 	Cluster API equivalent Machine remains in cluster```

says that the CAPI machine should stay.

```Cluster API 	- 	Delete 	Yes 	No 	Machine API synchronisation controller will recreate Machine after it is removed```

says that the MAPI machine should be re-created.

I see you've mentioned this in another comment, so I think my concern's being address in any case :)

**Comment by nrb** [comment]
I read through the Slack thread that Damiano linked, and that makes sense in terms of allowing a MachineSet not to care about how or which Machine is deleted. I do think mixing authoritative APIs at different levels is going to be confusing for users, though; it means "authoritative" isn't quite, well, authoritative, if the non-authoritative resource can trigger a deletion of the other.

All that said, I don' think that we can really redesign this at this point. Rather, we'll need to be very explicit about how these cases work in docs and logs so that we can quickly reason about things when users inevitably run into the issue.

**Comment by JoelSpeed** [comment]
> For Machines being able to delete the non-authoritative resource and have the delete propagate is important. It allows integrations (such as MachineSet) to be able to not care about which machine is authoritative, but still be able to scale up and down

I believe this is covered in what's described in the enhancement, by the fact that we need to take a look at whether the machine has an owner to decide if the deletion is propagated.

For Machines, we will only allow the removal of the non-authoritative MAPI machine when the MAPI machine has no owner. So things like MachineSet will continue to work, and MHC wouldn't remove a machine without an owner either, so it's behaviour is also unaffected

File: pkg/controllers/machinesetsync/machineset_sync_controller.go [hunk 8]
```diff
@@ -810,3 +959,40 @@ func compareMAPIMachineSets(a, b *machinev1beta1.MachineSet) (map[string]any, er
...
+		newCAPIMachineSet.SetResourceVersion(util.GetResourceVersion(client.Object(capiMachineSet)))
+		// Restore finalizers.
+		newCAPIMachineSet.SetFinalizers(capiMachineSet.GetFinalizers())
+	}
+
+	newCAPIMachineSet.SetNamespace(capiNamespace)
+	newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace = capiNamespace
+	newCAPIMachineSet.OwnerReferences = []metav1.OwnerReference{clusterOwnerRefence}
+	// Set the paused annotation on the new CAPI MachineSet, as we want to create it paused.
+	annotations.AddAnnotations(newCAPIMachineSet, map[string]string{capiv1beta1.PausedAnnotation: ""})
```
damdo, line 978: This will conflict with my PR likely

damdo:
This will need a bug to be attached and cherry-picking to release-4.19

PR #287, opened 2025-04-29
OCPCLOUD-2992: machineset migration e2e

sunzhaohua2:


theobarberbany:
Heya @sunzhaohua2 :) 

What we have here is useful, but if we look at the [CI run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/287/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1925064843459563520) we see it's passing in 0-1s. 

I think a lot of this is already covered by unit testing, so we might want to focus our efforts a little differently. 

IMO, here for e2es we want to test the behaviour of the MAPI controllers (machine, machineset, etc), CAPI controllers (machine, machineset, etc), and our migration controllers all interacting together - which we can't do with unit tests :) This is because in our unit tests (e.g for the MachineSync controllers) we're not running any of the CAPI or MAPI controllers. 

So the tests we write here want to check things like deletion logic, showing machines templates and so on end up running  / created without status errors (they're healthy - not that they merely exist and are accepted by the api server) and so on.

[OCPBUGS-56897](https://issues.redhat.com/browse/OCPBUGS-56897) this is a good example of a bug that could nicely be caught by an e2e test :) We would want to test the case where  the MAPI machine set goes away, and we assert the CAPI machines do not change (e.g the UUID remains the same consistently, for example).

I think we want to work out a base set of e2e tests that give us broad coverage, and start with adding only those, and then increment :) I'll try and give this some thought!

sunzhaohua2:
> Heya @sunzhaohua2 :)
> 
> What we have here is useful, but if we look at the [CI run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/287/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1925064843459563520) we see it's passing in 0-1s.
> 
> I think a lot of this is already covered by unit testing, so we might want to focus our efforts a little differently.
> 
> IMO, here for e2es we want to test the behaviour of the MAPI controllers (machine, machineset, etc), CAPI controllers (machine, machineset, etc), and our migration controllers all interacting together - which we can't do with unit tests :) This is because in our unit tests (e.g for the MachineSync controllers) we're not running any of the CAPI or MAPI controllers.
> 
> So the tests we write here want to check things like deletion logic, showing machines templates and so on end up running / created without status errors (they're healthy - not that they merely exist and are accepted by the api server) and so on.
> 
> [OCPBUGS-56897](https://issues.redhat.com/browse/OCPBUGS-56897) this is a good example of a bug that could nicely be caught by an e2e test :) We would want to test the case where the MAPI machine set goes away, and we assert the CAPI machines do not change (e.g the UUID remains the same consistently, for example).
> 
> I think we want to work out a base set of e2e tests that give us broad coverage, and start with adding only those, and then increment :) I'll try and give this some thought!
> 
> one example may be something like:
> 
> mapi auth , create machineset, observe capi machine + machineset mirror come up healthy, switch auth see other controllers take over, scale up / down see machines work and delete

Thanks for your suggestion, good advice!  I saw Joel said there will be an e2e doc, I will refer to the doc to make adjustments.

e2e/framework/util.go
```diff
@@ -32,3 +34,13 @@ func GetControlPlaneHostAndPort(cl client.Client) (string, int32, error) {
 
 	return apiUrl.Hostname(), int32(port), nil
 }
+
+// IsTechPreviewNoUpgrade checks if a cluster is a TechPreviewNoUpgrade cluster
```
JoelSpeed, line 38: You should check for the presence of a particular feature gate being enabled in the feature gate status rather than checking tech preview directly.

If we can use the openshift tests extension (I think we got that?), then naming tests based on the feature gate name will be sufficient and the test extension will skip any tests that aren't enabled for us

I realise we don't have a great feature gate for this feature right now, I'll take an action item to create that so that we can start moving on with this

JoelSpeed:
> In the E2E plan Dam wrote Create MAPI MachineSet With specAPI: CAPI (and no existing CAPI MSet with that name) This is possible to do at the moment, but shouldn’t, this means we don't need to add automation for creating MAPI MachineSet With specAPI: CAPI? If so, we need to auto all specAPI: CAPI cases after changing machineAPI to clusterAPI?

I'm not sure if that's true. 

I'd agree that creating a MAPI MachineSet with specAPI MAPI shouldn't be allowed when there is already a CAPI MachineSet. But in the case outlined here, I would expect the MachineSet to be created, and then a CAPI MachineSet to be created as a mirror of the MAPI MachineSet.

The mirror CAPI MachineSet would do the work, and the MAPI MachineSet would always be paused.

In a lot of the actions/scenarios, we care specifically about creating MachineSets, and less about the updates.
Once both MAPI and CAPI MachineSets exist, the behaviours are easier to reason about, writes to non-authoritative resources should be rejected in the most part, and changes to the authoritative resource should be reflected to the non-authoritative resource

JoelSpeed:
I suspect that creating a suite of tests that specifically check the creation mechanics and what happens, and then a separate suite of tests that check what happens once you already have the existing MachineSets, would make the most sense

sunzhaohua2:
Hi @JoelSpeed @theobarberbany can you help to review this?

JoelSpeed (commented):
When writing the CPMS E2Es, we used https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest/komega to wrap a lot of the fetching of resources, in particular the things like checking the spec/status of a certain field become

```
Eventually(komega.Object(machine)).Should(HaveField("Status.AuthoritativeAPI", Equal("ClusterAPI"))
```

Updating objects is also fairly simple, and uses retry logic to make sure we are not susceptible to flakes

```
Eventually(komega.Update(machine, func() {
  machine.Spec.AuthoritativeAPI = "ClusterAPI"
}).Should(Succeed())
}
```

I wonder if we could avoid a lot of boiler plate if we adopted the same library here as well?

This PR is also huge! I wonder if we could break it down into stages, start by merging the framework and util bits and then different sections of the tests suites. Might make it a bit easier to review/iterate on.

Also, where I've made suggestions, please see if those might apply in other places. A lot of transforms here could be replaced simply with `HaveField`. A lot of the fetching of objects could use `komega` combined with `HaveField` to extract the fields we care about

PR #288, opened 2025-04-29
OCPBUGS-55349,OCPBUGS-54970: Ensure CAPI machine when machine authority is CAPI

theobarberbany:
When we have a machine with authoritativeAPI of CAPI, but no CAPI machine exists we want to run an initial MAPI -> CAPI sync.

This updates the logic in the machinesync controller to do this, starting the newly created CAPI machine unpaused.


theobarberbany:
Yep, Damiano and I tested on a live cluster.

damdo (commented):
Worked on this with @theobarberbany 

/approve

leaving LGTM to @JoelSpeed

damdo:
This is a legitimate bug in techpreview that is impairing functionality, and doesn't directly affect stable OpenShit releasing efforts

PR #290, opened 2025-05-09
OCPBUGS-55985: fix: machine: propagate non-authoritative deletion

damdo:
When a non-authoritative machine is issued a deletion, the deletion should be propagated to the authoritative machine as well.

Context: https://redhat-internal.slack.com/archives/GE2HQ9QP4/p1746777766970039

### DIFF
--- a/pkg/controllers/machinesync/machine_sync_controller.go
+++ b/pkg/controllers/machinesync/machine_sync_controller.go
@@ -982,7 +982,20 @@ func (r *MachineSyncReconciler) fetchCAPIInfraResources(ctx context.Context, cap
 //nolint:funlen,gocognit,cyclop
 func (r *MachineSyncReconciler) reconcileMAPItoCAPIMachineDeletion(ctx context.Context, mapiMachine *machinev1beta1.Machine, capiMachine *capiv1beta1.Machine, infraMachine client.Object) (bool, error) {
 	if mapiMachine.DeletionTimestamp.IsZero() {
-		return false, nil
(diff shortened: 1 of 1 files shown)


pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -982,7 +982,20 @@ func (r *MachineSyncReconciler) fetchCAPIInfraResources(ctx context.Context, cap
...
+		}
+
+		// The MAPI authoritative machine is not being deleted, but the CAPI non-authoritative one is.
+		// Issue a deletion also to the MAPI authoritative machine.
+		if err := r.Client.Delete(ctx, mapiMachine); err != nil {
+			return false, fmt.Errorf("failed to delete Machine API machine: %w", err)
+		}
+
+		// Return true to force a requeue, to allow the deletion propagation.
+		return true, nil
```
JoelSpeed, line 998: The delete on the mapi machine will add an event to the queue, I don't think that requeueing actually does anything

damdo: Yes, the `true` there doesn't add a new event on the queue, But does just return and wait for an event triggered by the Delete. This is a snipped of the logic is doing in the caller of this func:

https://github.com/openshift/cluster-capi-operator/blob/ac5aa33f14d786ee00a8ed1647a7e99ccdb9f017/pkg/controllers/machinesync/machine_sync_controller.go#L411-L415

JoelSpeed (commented):
/lgtm
/approve

PR #273, opened 2025-03-18
TRT-2048: Revert "OCPCLOUD-2792: corecluster: set controller level conditions + tests"

xueqzhan:
Reverts openshift/cluster-capi-operator#256; tracked by https://issues.redhat.com/browse/OCPBUGS-52848

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are reverting this breaking change to get CI and/or nightly payloads flowing again.

This PR has caused samples operator to fail during installation starting from this payload: https://sippy.dptools.openshift.org/sippy-ng/release/4.19/tags/4.19.0-0.nightly-2025-03-05-232901/pull_requests

Slack thread: https://redhat-internal.slack.com/archives/C01CQA76KMX/p1741634385702679

To unrevert this, revert this PR, and layer an additional separate commit on top that addresses the problem. Before merging the unrevert, please run these jobs on the PR and check the result of (job/X or job/X, test/Y tuple) to confirm the fix has corrected the problem:

/payload-job periodic-ci-openshift-release-master-ci-4.19-e2e-gcp-ovn-techpreview-serial

CC: @damdo

PR #292, opened 2025-05-12
Apply provider manifests with SSA

mdbooth:
This change uses Server-Side Apply to apply provider manifests in place of the logic from library-go, which uses an Update with custom kind-specific client-side merge logic. Using SSA here should be equivalent to how these manifests would be applied by any other client.

The merge logic from library-go should no longer be required as long as the manifests are not specifying any fields which will be overwritten by an existing controller.

In particular, there should be no conflict with service-ca setting caBundle in various places, as long as the specified manifests do not include a caBundle. However, note that due to a validation bug in older versions of k8s, some CRDs do still specify an empty caBundle in their CRDs. These would have to be removed for this to work.

The expected flow of reconciles between cluster-capi-operator and service-ca-operator becomes:

* cluster-capi-operator: apply initial manifests
* service-ca-operator: adds caBundle to CRDs and validating webhooks
* cluster-capi-operator: triggered by update to manifests it managed applies manifests again
* The final update will produce no change as long as no managed fields have been updated, so there will be no further reconciles.

However, objects with the provider label were being reconciled too often due to insufficient filtering. This PR includes a second commit which addresses that.

### DIFF
--- a/pkg/controllers/capiinstaller/capi_installer_controller.go
+++ b/pkg/controllers/capiinstaller/capi_installer_controller.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-capi-operator/pkg/controllers"
@@ -268,6 +269,13 @@ func (r *CapiInstallerController) SetupWithManager(mgr ctrl.Manager) error {
 			&corev1.ConfigMap{},
 			handler.EnqueueRequestsFromMapFunc(toClusterOperator),
 			builder.WithPredicates(configMapPredicate(r.ManagedNamespace, r.Platform)),
+		).
+		// We reconcile all Deployment changes because we intend to reflect the
+		// status of any created Deployment in the ClusterOperator status.
+		Watches(
+			&appsv1.Deployment{},
+			handler.EnqueueRequestsFromMapFunc(toClusterOperator),
+			builder.WithPredicates(ownedPlatformLabelPredicate(r.ManagedNamespace, r.Platform)),
 		)
 
 	// All of the following watches share the ownedPlatformLabelPredicate.
@@ -275,7 +283,6 @@ func (r *CapiInstallerController) SetupWithManager(mgr ctrl.Manager) error {
 		obj       client.Object
 		namespace string
 	}{
-		{&appsv1.Deployment{}, r.ManagedNamespace},
 		{&admissionregistrationv1.ValidatingWebhookConfiguration{}, notNamespaced},
 		{&admissionregistrationv1.MutatingWebhookConfiguration{}, notNamespaced},
 		{&admissionregistrationv1.ValidatingAdmissionPolicy{}, notNamespaced},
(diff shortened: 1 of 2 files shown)


damdo:
I'll try and review this one in the coming days. Also as a domain expert it'd probably be good to have a review from @JoelSpeed

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -327,14 +269,20 @@ func (r *CapiInstallerController) SetupWithManager(mgr ctrl.Manager) error {
 			&corev1.ConfigMap{},
 			handler.EnqueueRequestsFromMapFunc(toClusterOperator),
 			builder.WithPredicates(configMapPredicate(r.ManagedNamespace, r.Platform)),
+		).
+		// We reconcile all Deployment changes because we intend to reflect the
+		// status of any created Deployment in the ClusterOperator status.
+		Watches(
+			&appsv1.Deployment{},
+			handler.EnqueueRequestsFromMapFunc(toClusterOperator),
+			builder.WithPredicates(ownedPlatformLabelPredicate(r.ManagedNamespace, r.Platform)),
```
mdbooth, line 278: I remember when we discussed this we decided to leave this in because of functionality we intend to implement in the future. Honestly I'd leave it out and add it back when we add something which needs it. Until then it's just adding noise.

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -177,55 +173,19 @@ func (r *CapiInstallerController) reconcile(ctx context.Context, log logr.Logger
...
-		}
-	}
-
 	var errs error
 
-	for i, r := range res {
-		if r.Error != nil {
-			errs = errors.Join(errs, fmt.Errorf("error applying CAPI provider component %q at position %d: %w", r.File, i, r.Error))
+	for i, providerObject := range providerObjects {
+		err := r.Patch(ctx, providerObject, client.Apply, client.ForceOwnership, client.FieldOwner("cluster-capi-operator.openshift.io/installer"))
```
mdbooth, line 184: bike shedding opportunity: what to call this field owner? I think this is good, but I'm just calling it out.

Should probably move the field owner to a constant somewhere.

pkg/controllers/capiinstaller/capi_installer_controller.go
damdo: Yes. See my comment above re:
`TODO: Deployments State/Conditions should influence the overall ClusterOperator Status`.

I think ideally we should be looking at doing this considering the upcoming GAing of the operator.

Maybe something we can tackle in tandem with the un-revert of: https://github.com/openshift/cluster-capi-operator/pull/273

In this PR we should at least see if we are introducing anything that would prevent us easily doing that.

pkg/controllers/capiinstaller/capi_installer_controller.go
damdo, line 184: Is there any obvious difference that we need to be aware of/account for, between `r.Patch & client.Apply + SSA` vs `resourceapply.ApplyDeployment`?

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -177,55 +173,19 @@ func (r *CapiInstallerController) reconcile(ctx context.Context, log logr.Logger
...
+		if err != nil {
+			gvk := providerObject.GroupVersionKind()
```
JoelSpeed, line 186: Why not use the GVK string function and then join that to the name?

pkg/controllers/capiinstaller/capi_installer_controller.go
```diff
@@ -177,55 +173,19 @@ func (r *CapiInstallerController) reconcile(ctx context.Context, log logr.Logger
...
+			name := strings.Join([]string{gvk.Group, gvk.Version, gvk.Kind, providerObject.GetName()}, "/")
```
JoelSpeed, line 187: Does the namespace matter?

pkg/controllers/capiinstaller/watch_predicates.go
```diff
@@ -32,11 +32,12 @@ func clusterOperatorPredicates() predicate.Funcs {
 		return ok && clusterOperator.GetName() == clusterOperatorName
 	}
 
+	// We only want to be reconciled on creation of the cluster operator,
+	// because we wait for it before reconciling. The Create event also fires
+	// when the manager is started, so this will additionally ensure we are
+	// called at least once at startup.
```
JoelSpeed, line 38: I thought it was a generic event triggered by the `List`? Not create?

pkg/controllers/capiinstaller/capi_installer_controller.go
mdbooth: I looked at it at the time, and convinced myself the answer is no. I could do that again...

However, intuitively this is fine. Consider that this is how all client tooling applies this. Deployments do not require special consideration. Perhaps they once did.

PR #295, opened 2025-05-12
OCPBUGS-54705: mapi2capi: Create new machine template when changing MAPI providerSpec

RadekManak:
This PR adds missing logic when updating MAPI MachineSet that is AuthorativeAPI=MachineAPI.

Because CAPI Infrastructure machine templates are immutable, we need to recreate it.

The logic is as follows:
1. Create a new infrastructure machine template with hash of spec in its name to avoid name conflict.
2. Modify the CAPI machineSet to use the new machine template.
3. Delete old infrastructure machine template.

This PR also adds the `machine.openshift.io/cluster-api-machineset` label to infrastructure machine templates converted from MAPI. This label serves 2 purposes.

- It allows listing old infrastructure machine templates that were created from conversion and should be deleted. When the reconcile fails at step 3. Without the label, the CAPI machineSet after step 2 no longer points to the old infrastructure machine template, and it would get orphaned.
- We can implement admission webhook that ensures only the machineSet matching the label can use the template. This restriction will prevent use of converted templates, that could be deleted when MAPI machineset is updated, by other CAPI machine sets.

I think we should treat the admission webhook as a feature rather than this bug.

### DIFF
--- a/pkg/controllers/common_consts.go
+++ b/pkg/controllers/common_consts.go
@@ -46,4 +46,7 @@ const (
 
 	// ReasonAuthoritativeAPIChanged indicates that sync state is stale due to a change of authoritativeAPI.
 	ReasonAuthoritativeAPIChanged = "AuthoritativeAPIChanged"
+
+	// MachineSetOpenshiftLabelKey is the key for label referring to a Machine API MachineSet.
+	MachineSetOpenshiftLabelKey = "machine.openshift.io/cluster-api-machineset"
 )
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -34,12 +34,14 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/go-test/deep"
 	machinev1applyconfigs "github.com/openshift/client-go/machine/applyconfigurations/machine/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
@@ -61,6 +63,9 @@ var (
 
 	// errUnexpectedInfraMachineTemplateType is returned when we receive an unexpected InfraMachineTemplate type.
 	errUnexpectedInfraMachineTemplateType = errors.New("unexpected InfraMachineTemplate type")
+
+	// errUnexpectedInfraMachineTemplateListType is returned when we receive an unexpected InfraStructureMachineTemplateList type.
+	errUnexpectedInfraMachineTemplateListType = errors.New("unexpected InfraMachineTemplateList type")
 
 	// errUnexpectedInfraClusterType is returned when we receive an unexpected InfraCluster type.
 	errUnexpectedInfraClusterType = errors.New("unexpected InfraCluster type")
@@ -135,7 +140,7 @@ func (r *MachineSetSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		).
 		Watches(
 			infraMachineTemplate,
-			handler.EnqueueRequestsFromMapFunc(util.ResolveCAPIMachineSetFromObject(r.MAPINamespace)),
+			handler.EnqueueRequestsFromMapFunc(util.ResolveCAPIMachineSetFromInfraMachineTemplate(r.MAPINamespace)),
 			builder.WithPredicates(util.FilterNamespace(r.CAPINamespace)),
 		).
 		Complete(r); err != nil {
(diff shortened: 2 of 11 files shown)


mdbooth (commented):
I have a couple of high level concerns. One is mentioned below: this change doesn't seem to consider CAPI->MAPI->CAPI round-tripping. If that's important, we'll need a fix.

The other is around Machine ownership. I haven't checked, but I want to confirm that CAPI InfraMachines are owned by the *MachineSet*, and not directly associated with the *InfraMachineTemplate* which they were templated from. If so, we don't have a problem because the MachineSet doesn't change. However, if there is any association between an InfraMachine and the InfraMachineTemplate which it was created from we should be handling that somewhere in this PR. I _suspect_ this is fine and CAPI just doesn't check, same as MAPI.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -318,10 +320,53 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 		return result, fmt.Errorf("unable to ensure CAPI machine set: %w", err)
 	}
 
+	if result, err := r.deleteOutdatedCAPIInfraMachineTemplates(ctx, mapiMachineSet, newCAPIInfraMachineTemplate); err != nil {
+		return result, fmt.Errorf("unable to delete outdated Cluster API infrastructure machine templates: %w", err)
```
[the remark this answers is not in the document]
JoelSpeed: But you can populate `ctrl.Result{}` with potential other options, for example to requeue after a certain period? So it does matter to be passed back up in case we want to control that from a lower level?

It may not be needed for every function we call so perhaps in this example it's best to remove?

pkg/conversion/test/fuzz/fuzz.go
```diff
@@ -204,6 +205,14 @@ func CAPI2MAPIMachineSetRoundTripFuzzTest(scheme *runtime.Scheme, infra *configv
 		// Expect(infraMachineTemplate.Status).To(Equal(in.infraMachineTemplate.Status))
 
 		capiMachineSet.Finalizers = nil
+
+		Expect(capiMachineSet.Spec.Template.Spec.InfrastructureRef.Name).To(
+			MatchRegexp("^" + regexp.QuoteMeta(in.machineSet.Spec.Template.Spec.InfrastructureRef.Name) + "(-[a-f0-9]{8})?$"),
+		)
+
+		// The name gets hash of spec added during conversion. Remove it before further comparison.
+		capiMachineSet.Spec.Template.Spec.InfrastructureRef.Name = in.machineSet.Spec.Template.Spec.InfrastructureRef.Name
```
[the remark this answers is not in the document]
JoelSpeed: If the content of the input to the hash function is stable, wouldn't the output 8 bytes be stable, and therefore the name in the ref should be able to round trip?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -526,8 +541,13 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 				})
 
 				It("should create the CAPI infra machine template", func() {
-					capiInfraMachineTemplate := capav1builder.AWSMachineTemplate().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
-					Eventually(k.Get(capiInfraMachineTemplate)).Should(Succeed())
+					capiMachineSet = capiv1resourcebuilder.MachineSet().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
+					Eventually(func() error {
+						_ = k8sClient.Get(ctx, client.ObjectKeyFromObject(capiMachineSet), capiMachineSet)
```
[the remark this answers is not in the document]
JoelSpeed: Yeah lets return this error if it happens

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -318,10 +320,53 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
...
 		consts.ReasonResourceSynchronized, messageSuccessfullySynchronizedMAPItoCAPI, &mapiMachineSet.Generation)
 }
 
+// deleteOutdatedCAPIInfraMachineTemplates deletes infra machine templates that have MAPI machine label of the current MachineSet and don't have the current computed hash.
+func (r *MachineSetSyncReconciler) deleteOutdatedCAPIInfraMachineTemplates(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, newCAPIInfraMachineTemplate client.Object) (ctrl.Result, error) { //nolint:unparam
+	logger := log.FromContext(ctx)
+
+	machineSetLabelSelector := client.MatchingLabels(map[string]string{"machine.openshift.io/cluster-api-machineset": mapiMachineSet.Name})
+
+	infraMachineTemplates := &capiv1beta1.MachineSetList{}
```
JoelSpeed, line 337: Why are we listing MachineSets? So we are deleting the MachineSet which removes the template and then it all gets regenerated? Is that not a little heavy handed?

pkg/controllers/machinesetsync/machineset_sync_controller.go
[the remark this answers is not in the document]
mdbooth: But not here: it's only used in the error path, so controller-runtime will only ever complain if it's not empty.

pkg/controllers/machinesetsync/machineset_sync_controller.go
mdbooth: I completely missed that this is listing MachineSets, not infra machine templates. I can't work out how the tests pass!

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -627,7 +673,7 @@ func (r *MachineSetSyncReconciler) createOrUpdateCAPIInfraMachineTemplate(ctx co
 
 	logger.Info("Changes detected for CAPI infra machine template. Updating it", "diff", fmt.Sprintf("%+v", capiInfraMachineTemplatesDiff))
 
-	if err := r.Update(ctx, newCAPIInfraMachineTemplate); err != nil {
+	if err := r.Create(ctx, newCAPIInfraMachineTemplate); err != nil {
```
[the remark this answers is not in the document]
mdbooth: To update, I think this is wrong either way because we'll be called on reconciliation of an existing object, in which case this will always fail.

You could write a CreateOrUpdate method, or you could just replace this with a server-side apply which already has create-or-update semantics.

pkg/conversion/test/fuzz/fuzz.go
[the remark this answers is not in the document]
mdbooth: It will only round trip if the original infra machine template was named according to the hash function. If it was user-generated, it won't be.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -309,31 +339,63 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
...
+		client.MatchingFieldsSelector{Selector: notNewCAPIInfraMachineTemplateNameFieldSelector},
+		client.MatchingLabelsSelector{Selector: machineSetMAPILabelSelector},
+	}
+
+	infraMachineTemplate, _, err := initInfraMachineTemplateAndInfraClusterFromProvider(r.Platform)
+	if err != nil {
+		return fmt.Errorf("failed to get infrastructure machine template from platform type: %w", err)
+	}
+
+	if err = r.DeleteAllOf(ctx, infraMachineTemplate, listOptions...); err != nil {
```
RadekManak, line 371: This delete is called always. While this does not cause an issue, it feels wrong.
I think we might still want to list first to have feedback on what is to be deleted. The logger below also prints always, regardless if anything was deleted.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -550,8 +623,14 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 				})
 
 				It("should create the CAPI infra machine template", func() {
-					capiInfraMachineTemplate := capav1builder.AWSMachineTemplate().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
-					Eventually(k.Get(capiInfraMachineTemplate)).Should(Succeed())
+					capiMachineSet = capiv1resourcebuilder.MachineSet().WithName(mapiMachineSet.Name).WithNamespace(capiNamespace.Name).Build()
+					Eventually(func() error {
+						err := k8sClient.Get(ctx, client.ObjectKeyFromObject(capiMachineSet), capiMachineSet)
+						Expect(err).NotTo(HaveOccurred())
```
JoelSpeed, line 629: Return the error rather than doing `Expect` inside this anonymous function

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -306,34 +336,66 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
...
 	}
 
-	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachineSet, corev1.ConditionTrue,
-		consts.ReasonResourceSynchronized, messageSuccessfullySynchronizedMAPItoCAPI, &mapiMachineSet.Generation)
+	notNewCAPIInfraMachineTemplateNameFieldSelector := fields.OneTermNotEqualSelector("metadata.name", newCAPIInfraMachineTemplate.GetName())
+
+	listOptions := []client.DeleteAllOfOption{
+		client.InNamespace(r.CAPINamespace),
+		client.MatchingFieldsSelector{Selector: notNewCAPIInfraMachineTemplateNameFieldSelector},
+		client.MatchingLabelsSelector{Selector: machineSetMAPILabelSelector},
```
JoelSpeed, line 363: Is it possible that the same template will be shared by multiple machinesets? Do we need to add a validation that only one MachineSet can point to any template at a time?

You can do this with a VAP

pkg/controllers/machinesetsync/machineset_sync_controller.go
JoelSpeed: I agree, can we list, check for items in the list, and then call the delete when there is a non-zero length list please? Otherwise we are making API calls regularly that do nothing, which is bad practice 

And we can also avoid the log spam too!

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -587,66 +650,54 @@ func (r *MachineSetSyncReconciler) applySynchronizedConditionWithPatch(ctx conte
 }
 
 // createOrUpdateCAPIInfraMachineTemplate creates a CAPI infra machine template from a MAPI machine set, or updates if it exists and it is out of date.
-func (r *MachineSetSyncReconciler) createOrUpdateCAPIInfraMachineTemplate(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, infraMachineTemplate client.Object, newCAPIInfraMachineTemplate client.Object) (ctrl.Result, error) { //nolint:unparam
+func (r *MachineSetSyncReconciler) createOrUpdateCAPIInfraMachineTemplate(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, infraMachineTemplate client.Object, newCAPIInfraMachineTemplate client.Object) error {
```
JoelSpeed, line 653: When we are updating. Is there a path that allows the objectmeta (labels/annotations) to be updated without deleting/recreating the InfraMachineTemplate?

Reason I ask, is that the objectmeta is not included in the spec hash, so in theory we could generate two identically named templates, when the only difference is is objectmeta, and I think this would break if we aren't handling this case specifically as an update to the existing resource

pkg/controllers/machinesetsync/machineset_sync_controller.go
RadekManak: Right now yes. The goal is to prevent that via VAP. Because the Infra template kind needs to be parameterized I will need to update capiinstaller to template and deploy manifests from this repo. I'd prefer to do this in another PR. This one is already large and need to be backported to 4.19.

```
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: SyncedInfraMachineTemplateSingleUsePolicy
spec:
  failurePolicy: Fail
  paramKind:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: {{ ".InfraMachineTemplate" }}
  matchConstraints:
    resourceRules:
    - apiGroups:   ["cluster.x-k8s.io"]
      apiVersions: ["v1beta1"]
      operations:  ["CREATE", "UPDATE"]
      resources:   ["machinesets"]
  validations:
    - expression: |
        !(
          object.spec.template.spec.infrastructureRef.kind == params.kind &&
          object.spec.template.spec.infrastructureRef.name == params.metadata.name &&
          (has(object.spec.template.spec.infrastructureRef.namespace) ? object.spec.template.spec.infrastructureRef.namespace : "") == (has(params.metadata.namespace) ? params.metadata.namespace : "") &&
          object.metadata.name != params.metadata.labels['machine.openshift.io/cluster-api-machine-set']
        )
      message: "The referenced infrastructure template has the 'machine.openshift.io/cluster-api-machine-set' label, and this MachineSet's name does not match the label's value."
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: SyncedInfraMachineTemplateSingleUsePolicyBinding
spec:
  policyName: SyncedInfraMachineTemplateSingleUsePolicy
  validationActions: [Deny]
  paramRef:
    selector:
      matchExpressions:
        - key: machine.openshift.io/cluster-api-machine-set
          operator: Exists
    parameterNotFoundAction: Allow
  matchResources:
    namespaceSelector: {"openshift-cluster-api"}
    objectSelector: {}
```

pkg/controllers/machinesetsync/machineset_sync_controller.go
RadekManak: I think this is now fixed with use of Patch. If only labels or annotations differ, they will be updated in place.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -672,22 +788,22 @@ func (r *MachineSetSyncReconciler) createOrUpdateCAPIMachineSet(ctx context.Cont
...
-				return ctrl.Result{}, utilerrors.NewAggregate([]error{createErr, condErr})
+				return utilerrors.NewAggregate([]error{createErr, condErr})
 			}
 
-			return ctrl.Result{}, createErr
+			return createErr
 		}
 
-		logger.Info("Successfully created CAPI machine set")
+		logger.Info("Successfully created CAPI machine set", "name", newCAPIMachineSet.Name, "infraMachineTemplate", newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Name)
```
theobarberbany, line 797: <3 better logs

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -603,66 +731,54 @@ func (r *MachineSetSyncReconciler) applySynchronizedConditionWithPatch(ctx conte
...
 // createOrUpdateCAPIInfraMachineTemplate creates a CAPI infra machine template from a MAPI machine set, or updates if it exists and it is out of date.
-func (r *MachineSetSyncReconciler) createOrUpdateCAPIInfraMachineTemplate(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, infraMachineTemplate client.Object, newCAPIInfraMachineTemplate client.Object) (ctrl.Result, error) { //nolint:unparam
+func (r *MachineSetSyncReconciler) createOrUpdateCAPIInfraMachineTemplate(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, infraMachineTemplate client.Object, newCAPIInfraMachineTemplate client.Object) error {
 	logger := log.FromContext(ctx)
 
-	if infraMachineTemplate == nil {
-		if err := r.Create(ctx, newCAPIInfraMachineTemplate); err != nil {
-			logger.Error(err, "Failed to create CAPI infra machine template")
-			createErr := fmt.Errorf("failed to create CAPI infra machine template: %w", err)
+	if infraMachineTemplate != nil {
```
theobarberbany, line 737: Do we attempt creation any more? Does applying a patch `Create()` ? TIL.

theobarberbany (commented):
First look over, looks good. I have a couple of questions / comments. 

The e2e is still failing? 

`Cluster API Suite: [It] Cluster API AWS MachineSet should be able to run a machine with a default provider spec`

Peeking at the logs, it looks like this might've broken something :(  Might be worth hooking up to a debugger to try and work out whats going on.

pkg/controllers/common_consts.go
```diff
@@ -43,4 +43,7 @@ const (
 	// ReasonResourceSynchronized denotes that the resource is synchronized
 	// successfully.
 	ReasonResourceSynchronized = "ResourceSynchronized"
+
+	// MachineSetOpenshiftLabelKey is the key for label referring to a Machine API MachineSet.
+	MachineSetOpenshiftLabelKey = "machine.openshift.io/cluster-api-machineset"
```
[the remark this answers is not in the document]
RadekManak: This is the label we use in machine api. Machine API is a fork of cluster API and the label is this due to it.

pkg/controllers/machinesetsync/machineset_sync_controller.go
RadekManak: Yes. This uses server side apply with `client.Apply` and `Force: true`.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -321,30 +332,145 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
...
+
+	return outdatedTemplates, nil
+}
+
+// deleteOutdatedCAPIInfraMachineTemplates deletes infra machine templates that have MAPI machine label of the current MachineSet and don't have the current computed hash.
+func (r *MachineSetSyncReconciler) deleteOutdatedCAPIInfraMachineTemplates(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet, newCAPIInfraMachineTemplate client.Object) error {
+	logger := log.FromContext(ctx)
+
+	if mapiMachineSet == nil || len(mapiMachineSet.Name) == 0 {
+		return errNilMachineAPIMachineSet
```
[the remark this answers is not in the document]
RadekManak: I added this in a previous iteration as a sanity check because the function called deleteAll without checking anything. Now it lists the objects before proceeding with the deletion. I removed this.

RadekManak (commented):
Updated based on feedback from @theobarberbany.
I missed renaming the label in one place, which may have caused the E2E failure. Le't see if it is fixed now.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -78,6 +83,12 @@ var (
 
 	// errMachineAPIMachineSetOwnerReferenceConversionUnsupported.
 	errMachineAPIMachineSetOwnerReferenceConversionUnsupported = errors.New("could not convert Machine API machine set owner references to Cluster API")
+
+	// errNilMachineAPIMachineSet is returned when the Machine API machine set is nil.
```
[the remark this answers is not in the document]
RadekManak: These should have been wrapped with more information. These are deleted now.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -329,30 +334,137 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
...
 	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, mapiMachineSet, corev1.ConditionTrue,
 		controllers.ReasonResourceSynchronized, messageSuccessfullySynchronizedMAPItoCAPI, &mapiMachineSet.Generation)
 }
 
+// filterOutdatedInfraMachineTemplates takes infraMachineTemplatesList and constructs a slice of InfraMachineTemplates without newInfraMachineTemplate.
+func filterOutdatedInfraMachineTemplates(infraMachineTemplateList client.ObjectList, newInfraMachineTemplate client.Object) ([]client.Object, error) {
+	outdatedTemplates := []client.Object{}
+
+	switch list := infraMachineTemplateList.(type) {
+	case *awsv1.AWSMachineTemplateList:
```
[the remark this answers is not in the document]
RadekManak: Only possible using reflection, which I tried to avoid.

The InfraMachineTemplate interface does not have Items variable. We get the correct type from initInfraMachineTemplateListAndInfraClusterListFromProvider as interface. Pass the object to list to populate and here we cast it back into concrete type to get the []client.Object which we can use normally.

PR #299, opened 2025-05-22
OCPBUGS-56604: fix: ensure CAPI machineset is created not paused when authoritative

damdo:
At the moment  when a MAPI MachineSet is created with `.spec.authoritativeAPI: ClusterAPI`, the authoritative CAPI MachineSet mirror is created as paused.

This should not be the case as otherwise both MachineSets engines will be paused and nothing will progress.

--

This partially relates to the machine sync bug fixed here: https://github.com/openshift/cluster-capi-operator/pull/288

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller.go
@@ -252,7 +252,9 @@ func (r *MachineSetSyncReconciler) syncMachineSets(ctx context.Context, mapiMach
 	case authoritativeAPI == machinev1beta1.MachineAuthorityMigrating:
 		logger.Info("Machine set is currently being migrated")
 		return ctrl.Result{}, nil
-
+	case authoritativeAPI == "":
+		logger.Info("Machine set status.authoritativeAPI is empty, will check again later", "AuthoritativeAPI", mapiMachineSet.Status.AuthoritativeAPI)
+		return ctrl.Result{}, nil
 	default:
 		logger.Info("Unexpected value for authoritativeAPI", "AuthoritativeAPI", mapiMachineSet.Status.AuthoritativeAPI)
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -259,9 +261,17 @@ func (r *MachineSetSyncReconciler) syncMachineSets(ctx context.Context, mapiMach
 }
 
 // reconcileMAPIMachineSetToCAPIMachineSet reconciles a MAPI MachineSet to a CAPI MachineSet.
+//
+//nolint:funlen
```
theobarberbany, line 265: Can we refactor anything in here (abstract out to helper funcs etc) to avoid this?

theobarberbany: Actually I think if you rebase this on top of the deletion PR and merge after that one has you'll avoid it as I've done some cleanup there.

damdo: This is a bug fix so I'd like to keep it small. Also I'd like to backport this to 4.19, so I don't want to change things up too much.
I think we can do the refactoring in yours yes, but I don't think this should be blocked on that one.

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -307,8 +317,11 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 	newCAPIMachineSet.SetNamespace(r.CAPINamespace)
 	newCAPIMachineSet.Spec.Template.Spec.InfrastructureRef.Namespace = r.CAPINamespace
 	newCAPIMachineSet.OwnerReferences = []metav1.OwnerReference{clusterOwnerRefence}
-	// Set the paused annotation on the new CAPI MachineSet, as we want to create it paused.
-	annotations.AddAnnotations(newCAPIMachineSet, map[string]string{capiv1beta1.PausedAnnotation: ""})
+
+	if authoritativeAPI == machinev1beta1.MachineAuthorityMachineAPI {
+		// Set the paused annotation on the new CAPI MachineSet, if the authoritativeAPI is Machine API.
```
theobarberbany, line 322: I might be tempted to rewrite this comment include the *why*: to call out we're explicitly handling the case where we run an initial MAPI -> CAPI sync

theobarberbany: Ah right, but what I meant was a comment explaining why the normal case was branched (from the POV of the principle of least surprise): when we're MachineAuthorityClusterAPI we don't want to set paused. I don't think that's immediately clear reading the code, especially with the way around we're having the normal case branching.(I think normal case branching makes sense for code simplicity)

PR #300, opened 2025-05-26
OCPBUGS-56487: fix: watch CAPI resources in the migration controllers

damdo:
The CAPI resources were not being watched by the migration controllers (both the machine and the machineset ones).
This meant that if the status of the CAPI resource changed from Paused: True to Paused: False or vice-versa, this would not get detected by the migration controllers, which would stay waiting in `Requested pausing for authoritative machine set` when moving a MachineSet authoritativeness from CAPI to MAPI.

### DIFF
--- a/pkg/controllers/common.go
+++ b/pkg/controllers/common.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


pkg/controllers/common.go
```diff
@@ -0,0 +1,61 @@
...
+func InitInfraMachineTemplateAndInfraClusterFromProvider(platform configv1.PlatformType) (client.Object, client.Object, error) {
+	switch platform {
+	case configv1.AWSPlatformType:
+		return &capav1beta2.AWSMachineTemplate{}, &capav1beta2.AWSCluster{}, nil
+	case configv1.PowerVSPlatformType:
+		return &capibmv1.IBMPowerVSMachineTemplate{}, &capibmv1.IBMPowerVSCluster{}, nil
+	default:
+		return nil, nil, fmt.Errorf("%w: %s", errPlatformNotSupported, platform)
+	}
+}
```
damdo, line 61: This is code motion, see the removals below.

RadekManak:
Looks good. I need to update my PR to reconcile correct machineSet when enqueued from InfraMachineTemplate event. As the name won't match with the added hask.

Here it is correct for now.
/lgtm
/approve

PR #309, opened 2025-06-02
OCPBUGS-57040: Fix failure to retry unpause

mdbooth:
The migration controllers shortcut reconciliation if their status indicate prior success. Therefore setting this status must be the last thing the controller does after all work is complete.

We were unpausing after setting completion status. Therefore if we failed to set unpause due to, e.g. an epheral api-server issue, we would never retry.

### DIFF
--- a/pkg/controllers/machinemigration/machine_migration_controller.go
+++ b/pkg/controllers/machinemigration/machine_migration_controller.go
@@ -190,14 +190,14 @@ func (r *MachineMigrationReconciler) Reconcile(ctx context.Context, req reconcil
 		return ctrl.Result{}, nil
 	}
 
-	// Set the actual AuthoritativeAPI to the desired one, reset the synchronized generation and condition.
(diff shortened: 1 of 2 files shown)


damdo:
@mdbooth I tested this and it seems to be working 👍 
Here you are doing it for the machine_migration_controller, you'll need to do the same for the machineset one.

mdbooth:
> @mdbooth I tested this and it seems to be working 👍 Here you are doing it for the machine_migration_controller, you'll need to do the same for the machineset one.

Good point. Lets do both at the same time rather than proliferate PRs. I'll update this.

mdbooth:
@damdo Updated to cover both controllers.

This will conflict with https://github.com/openshift/cluster-capi-operator/pull/313, but it's simpler so lets merge it first regardless.

damdo:
/retitle OCPBUGS-57040: Fix failure to retry unpause

damdo (commented):
I tested this manually and was able to verify it does work as intended.

/approve
/lgtm

PR #310, opened 2025-06-02
migration: Don't shortcut reconciliation when status is not set

mdbooth:
We were assuming that any object without authoritativeAPI in status was
already in the expected migration state. This change removes that
assumption and allows the migration controller to reconcile the object
before marking it as complete. This also ensures that the migration
condition is set correctly, as the shortcut doesn't fully update the
object status.

### DIFF
--- a/pkg/controllers/machinemigration/machine_migration_controller.go
+++ b/pkg/controllers/machinemigration/machine_migration_controller.go
@@ -122,17 +122,6 @@ func (r *MachineMigrationReconciler) Reconcile(ctx context.Context, req reconcil
 		return ctrl.Result{}, nil
 	}
 
-	// If authoritativeAPI status is empty, it means it is the first time we see this resource.
-	// Set the status.authoritativeAPI to match the spec.authoritativeAPI.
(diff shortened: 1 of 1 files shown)


JoelSpeed:
> We were assuming that any object without authoritativeAPI in status was
already in the expected migration state.

I disagree with the assessment of this PR currently. And think we need to keep this portion of the code. As far as I can tell the issues you have picked up are key parts of the design of the mechanism.

/hold for discussion

On upgrade, we expect the `status.authoritativeAPI` to be empty. We also say expect that controllers observing an empty `status.authoritativeAPI` should do nothing.

With this change, the `status.authoritativeAPI` will stay empty until there has been a successful synchronisation. Which means that reconciliation will be blocked on any object where the synchronisation cannot be completed, which, by design, we intend to use for any feature that hasn't yet been implemented (feature gaps).

**It is ok** for the `status.authoritativeAPI` to be set, and for the synchronised resource to not be present/up to date. This is expected in a lot of cases, eg user makes a change to add an unmigratable configuration. This is the purpose of the synchronised condition itself, to explain to users when there is an issue. We still need to show which resource we are observing as authoritative in this state.

PR #311, opened 2025-06-03
fix: synchronized condition should reset on migration

damdo:
Superseded by https://github.com/openshift/cluster-capi-operator/pull/313

---
Synchronized condition should reset on successful migration.
To then let the sync controllers take ownership of the condition and switch it back to True when a new sync has happened.
It is sufficient for us to unset the True and move to Unknown for this.


damdo:
/close

if favour of #313

PR #313, opened 2025-06-03
OCPBUGS 57013: Fix reset sync status in migration controllers

mdbooth:
Fixes a bug where we indended to remove the Synchronized condition on
completion of a migration, but instead took ownership of all conditions except
Synchronized.

Takes the opportunity to refactor the generation of the sync controller's
status in to common code using generic types. We can now set status using
common code in all 4 places it is set (machine sync, machineset sync, machine
migration, machineset migration).


damdo:
@mdbooth you may want to rebase this, now that  #309 merged

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -144,6 +144,20 @@ func (r *MachineMigrationReconciler) Reconcile(ctx context.Context, req reconcil
 		return ctrl.Result{}, nil
 	}
 
+	// Check if the synchronizedGeneration matches the old authoritativeAPI's resource current generation.
+	//
+	// This MUST be checked BEFORE setting status.authoritativeAPI to Migrating,
+	// because after that the sync controller will not run to update it and we
+	// will hang.
```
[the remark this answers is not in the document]
mdbooth: The sync controller doesn't reconcile if we're in `Migrating`. I think the current behaviour is desirable, because in that case the required direction of sync is indeterminate: we don't want to be writing changes to the sync target if it's active.

If we allowed sync during Migrating, then there's a possibility that a subsequent change to the MAPI object could be written to an unpaused CAPI object, which would be undesirable. It's safer just to cut it off, and effectively move immediately to the post-migration situation where they'll (eventually) be overwritten. And then we can further improve the UX by adding a VAP to prevent modification on admission.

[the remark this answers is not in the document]
JoelSpeed: More thinking on this and re-reading the EP, I think the change proposed here makes sense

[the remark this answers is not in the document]
damdo: We verified this together manually and with the debugger

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -133,12 +134,16 @@ func (r *MachineMigrationReconciler) Reconcile(ctx context.Context, req reconcil
...
-		return ctrl.Result{}, fmt.Errorf("unable to get synchronizedCondition for %s/%s: %w", mapiMachine.Namespace, mapiMachine.Name, err)
-	} else if cond != corev1.ConditionTrue {
-		logger.Info("Machine not synchronized with latest authoritative generation, will retry later")
+	// Check that the resource is synchronized and up-to-date.
+	//
+	// This MUST be checked BEFORE setting status.authoritativeAPI to Migrating,
+	// because after that the sync controller will not run to update it and we
+	// will deadlock.
+	if isSynchronized, err := r.isSynchronized(ctx, mapiMachine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to check synchronizedGeneration matches old authority: %w", err)
```
damdo, line 143: The semantics of isSynchornized cover more ground now
```suggestion
		return ctrl.Result{}, fmt.Errorf("unable to check the resource is synchronized and up-to-date with its authority: %w", err)
```

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -177,24 +182,14 @@ func (r *MachineSetMigrationReconciler) Reconcile(ctx context.Context, req recon
...
 	// Make sure the new authoritative resource has been requested to unpause.
 	if err := r.ensureUnpauseRequestedOnNewAuthoritativeResource(ctx, mapiMachineSet); err != nil {
 		return ctrl.Result{}, fmt.Errorf("unable to ensure the new AuthoritativeAPI has been un-paused: %w", err)
 	}
 
 	// Set the actual AuthoritativeAPI to the desired one, reset the synchronized generation and condition.
-	if err := r.setNewAuthoritativeAPIAndResetSynchronized(ctx, mapiMachineSet, mapiMachineSet.Spec.AuthoritativeAPI); err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to apply authoritativeAPI to status with patch: %w", err)
+	if err := synccommon.ApplyAuthoritativeAPIAndResetSyncStatus[*machinev1applyconfigs.MachineSetStatusApplyConfiguration](ctx, r.Client, controllerName, machinev1applyconfigs.MachineSet, mapiMachineSet, mapiMachineSet.Spec.AuthoritativeAPI); err != nil {
+		return ctrl.Result{}, fmt.Errorf("resetting sync status: %w", err)
```
damdo, line 192: This sounds a bit vaguer now, I would prefer being more explicit in what is happening here
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to apply authoritativeAPI and reset sync status: %w", err)
```

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -346,76 +341,50 @@ func (r *MachineMigrationReconciler) requestOldAuthoritativeResourcePaused(ctx c
 	return updated, nil
 }
 
-func (r *MachineMigrationReconciler) isSynchronizedGenerationMatchingOldAuthoritativeAPIGeneration(ctx context.Context, mapiMachine *machinev1beta1.Machine) (bool, error) {
-	//nolint:wsl
+func (r *MachineMigrationReconciler) isSynchronized(ctx context.Context, mapiMachine *machinev1beta1.Machine) (bool, error) {
+	// Check if the Synchronized condition is set to True.
+	// If it is not, this indicates an unmigratable resource and therefore should take no action.
+	if cond, err := util.GetConditionStatus(mapiMachine, string(controllers.SynchronizedCondition)); err != nil {
+		return false, fmt.Errorf("unable to get synchronizedCondition for %s/%s: %w", mapiMachine.Namespace, mapiMachine.Name, err)
```
damdo, line 348: I know this is my wrongdoing, but while you are at it, let's break down `synchronizedCondition` into two words
```suggestion
		return false, fmt.Errorf("unable to get synchronized condition for %s/%s: %w", mapiMachine.Namespace, mapiMachine.Name, err)
```

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -133,12 +134,16 @@ func (r *MachineSetMigrationReconciler) Reconcile(ctx context.Context, req recon
...
-		return ctrl.Result{}, fmt.Errorf("unable to get synchronizedCondition for %s/%s: %w", mapiMachineSet.Namespace, mapiMachineSet.Name, err)
-	} else if cond != corev1.ConditionTrue {
-		logger.Info("Machine set not synchronized with latest authoritative generation, will retry later")
+	// Check that the resource is synchronized and up-to-date.
+	//
+	// This MUST be checked BEFORE setting status.authoritativeAPI to Migrating,
+	// because after that the sync controller will not run to update it and we
+	// will deadlock.
+	if isSynchronized, err := r.isSynchronized(ctx, mapiMachineSet); err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to check synchronizedGeneration matches old authority: %w", err)
```
damdo, line 143: Same here

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -306,18 +301,34 @@ func (r *MachineSetMigrationReconciler) requestOldAuthoritativeResourcePaused(ct
 	return updated, nil
 }
 
-func (r *MachineSetMigrationReconciler) isSynchronizedGenerationMatchingOldAuthoritativeAPIGeneration(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet) (bool, error) {
-	//nolint:wsl
+func (r *MachineSetMigrationReconciler) isSynchronized(ctx context.Context, mapiMachineSet *machinev1beta1.MachineSet) (bool, error) {
+	// Check if the Synchronized condition is set to True.
+	// If it is not, this indicates an unmigratable resource and therefore should take no action.
+	if cond, err := util.GetConditionStatus(mapiMachineSet, string(controllers.SynchronizedCondition)); err != nil {
+		return false, fmt.Errorf("unable to get synchronizedCondition for %s/%s: %w", mapiMachineSet.Namespace, mapiMachineSet.Name, err)
```
damdo, line 308: Same here

damdo (commented):
Thanks a lot, this fixes some hefty issues and dedupes code 👍 

I left some minor comments on error messages and logs, nothing on the logic itself, which looks good to me.

pkg/controllers/machinemigration/machine_migration_controller.go
```diff
@@ -346,76 +341,50 @@ func (r *MachineMigrationReconciler) requestOldAuthoritativeResourcePaused(ctx c
...
-		// Value is disallowed by the openAPI schema validation.
 	}
 
-	return false, nil
+	// Should have been prevented by validation
+	return false, fmt.Errorf("%w: %s", controllers.ErrInvalidSpecAuthoritativeAPI, mapiMachine.Spec.AuthoritativeAPI)
 }
 
 // applyStatusAuthoritativeAPIWithPatch updates the resource status.authoritativeAPI using a server-side apply patch.
 func (r *MachineMigrationReconciler) applyStatusAuthoritativeAPIWithPatch(ctx context.Context, m *machinev1beta1.Machine, authority machinev1beta1.MachineAuthority) error {
```
JoelSpeed, line 388: Do we really need to wrap this? Why not just calll the synccommon directly?

pkg/controllers/machinesetmigration/machineset_migration_controller.go
```diff
@@ -327,55 +338,13 @@ func (r *MachineSetMigrationReconciler) isSynchronizedGenerationMatchingOldAutho
...
-		WithSynchronizedGeneration(0)
-
-	ac := machinev1applyconfigs.MachineSet(ms.Name, ms.Namespace).WithStatus(statusAc)
-
-	if err := r.Status().Patch(ctx, ms, util.ApplyConfigPatch(ac), client.ForceOwnership, client.FieldOwner(controllerName+"-AuthoritativeAPI")); err != nil {
-		return fmt.Errorf("failed to patch Machine API machine set status with authoritativeAPI %q: %w", authority, err)
-	}
-
-	return nil
+	return synccommon.ApplyAuthoritativeAPI[*machinev1applyconfigs.MachineSetStatusApplyConfiguration](ctx, r.Client, controllerName, machinev1applyconfigs.MachineSet, ms, authority)
```
JoelSpeed, line 349: Again, why wrap instead of direct call to the function?

pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -564,52 +562,10 @@ func (r *MachineSetSyncReconciler) convertMAPIToCAPIMachineSet(mapiMachineSet *m
...
-
-	msAc := machinev1applyconfigs.MachineSet(mapiMachineSet.GetName(), mapiMachineSet.GetNamespace()).
-		WithStatus(statusAc)
-
-	if err := r.Status().Patch(ctx, mapiMachineSet, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner(controllerName+"-SynchronizedCondition")); err != nil {
-		return fmt.Errorf("failed to patch MAPI machine set status with synchronized condition: %w", err)
-	}
-
-	return nil
+	return synccommon.ApplySyncStatus[*machinev1applyconfigs.MachineSetStatusApplyConfiguration](
```
JoelSpeed, line 565: Just call this directly at the call site?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -1333,50 +1331,8 @@ func compareCAPIInfraMachines(platform configv1.PlatformType, infraMachine1, inf
...
-
-	msAc := machinev1applyconfigs.Machine(mapiMachine.GetName(), mapiMachine.GetNamespace()).
-		WithStatus(statusAc)
-
-	if err := r.Status().Patch(ctx, mapiMachine, util.ApplyConfigPatch(msAc), client.ForceOwnership, client.FieldOwner(controllerName+"-SynchronizedCondition")); err != nil {
-		return fmt.Errorf("failed to patch Machine API machine status with synchronized condition: %w", err)
-	}
-
-	return nil
+	return synccommon.ApplySyncStatus[*machinev1applyconfigs.MachineStatusApplyConfiguration](
```
JoelSpeed, line 1334: Call directly?

JoelSpeed (commented):
Minor nites but otherwise LGTM

pkg/controllers/machinemigration/machine_migration_controller.go
JoelSpeed: Not a blocking nit

pkg/controllers/synccommon/applyconfiguration.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// syncObjApplyConfiguration is an apply configuration for objects managed by
+// the sync controller. This is currently MAPI Machine and MachineSet.
+type syncObjApplyConfiguration[objPT any, statusPT syncStatusApplyConfiguration[statusPT]] interface {
+	WithStatus(statusPT) objPT
+	WithResourceVersion(string) objPT
+}
+
+// syncObjApplyConfigurationP asserts that a syncObjApplyConfiguration is a pointer to a specific concrete type.
```
theobarberbany, line 31: my head's hurting a little (I think this is just generics) - why do we need to assert we have pointers to concrete types?

PR #321, opened 2025-06-19
Adds MAPI Machine VAP and test suites

theobarberbany:


pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +758,353 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+        ])
+    - name: check-authoritativeAPI-clusterapi
+      expression: 'object.status.authoritativeAPI == "ClusterAPI"'
+    - name: check-param-match
+      expression: 'object.metadata.name == params.metadata.name'
+
+  variables:
+    # label maps
+    - name: newLabels
+      expression: "object.metadata.?labels.hasValue() ? object.metadata.?labels.value()     : {}"
```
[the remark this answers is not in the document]
theobarberbany: CEL is really confusing here IMO. This is just a variable to extract labels from `object` safely.

The .? is an optional, so we don't panic, and can call `hasValue()` which gives us a safe way to check if labels is populated at all. 

That is then the first part of a conditional: `cond ? a : b`

Which in our case reads as: 

If we have a value for labels, return a (`object.metadata.?labels.value()`), otherwise return `{}`.

This means e.g if we have an update where we delete all labels/annotations, we don't panic trying to evaluate the CEL.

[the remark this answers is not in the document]
theobarberbany: I guess we could simplify this to`let labelOpt = obj.metadata.?labels`

and then it becomes this which is easier to read

`labelOpt.hasValue() ? labelOpt.value() : {}`

RadekManak:
I see this is just tests which include the VAP as strings. Do you plan to implement the manifest deployment in this PR? I would preffer for the tests to load the VAP yaml from manifests.

theobarberbany:
Yep, the plan is to do that @RadekManak, was just getting this out for early eyes on it :)

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
...
+    namespace: openshift-cluster-api
+    parameterNotFoundAction: Deny
+    selector: {}
+  policyName: machine-api-machine-vap
+  validationActions: [Deny]
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: machine-api-machine-vap
```
damdo, line 20: Will we have separate ones for machines and machinesets?

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
...
+    - name: check-only-non-service-account-requests
+      expression: >-
+        !(request.userInfo.username in [
+            "system:serviceaccount:openshift-machine-api:machine-api-controllers",
+            "system:serviceaccount:openshift-cluster-api:machine-api-migration"
+        ])
+    - name: check-authoritativeAPI-clusterapi
+      expression: 'object.status.authoritativeAPI == "ClusterAPI"'
+    - name: check-param-match
+      expression: 'object.metadata.name == params.metadata.name'
```
damdo, line 46: We could add a comment to say why we only consider these requests

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: machine-api-machine-vap
+spec:
+  matchResources:
+    namespaceSelector:
+      matchLabels:
+        name: openshift-machine-api
```
damdo, line 9: Does this mean we will need to have a label with key `name` and value `openshift-machine-api` on the namespace?

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
...
+            : !p[1].hasValue()
+        )
+
+  # All validations must evaluate to TRUE
+  validations:
+    # Only spec.authoritativeAPI may change
+    - expression: "object.spec == oldObject.spec || variables.specLockedExceptAuthoritativeAPI"
+      message:  "You may only modify spec.authoritativeAPI. Any other change inside .spec is not allowed."
+
+    # Guard machine.openshift.io/* and kubernetes.io/* labels
```
damdo, line 84:
```suggestion
    # Guard against machine.openshift.io/* and kubernetes.io/* labels changes
```

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
...
+              ((k in variables.oldLabels ? variables.oldLabels[k] : null) != variables.newLabels[k])
+          ) ||
+          variables.oldLabels.exists(k,
+              (k.startsWith('machine.openshift.io') || k.startsWith('kubernetes.io')) &&
+              !(k in variables.newLabels)
+          )
+        )
+      message: "Cannot modify or delete any machine.openshift.io/* or kubernetes.io/* label."
+
+    # Guard machine.openshift.io/* annotations
```
damdo, line 98:
```suggestion
    # Guard against machine.openshift.io/* annotations changes
```

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
...
+          variables.newAnn.exists(k,
+              k.startsWith('machine.openshift.io') &&
+              ((k in variables.oldAnn ? variables.oldAnn[k] : null) != variables.newAnn[k])
+          ) ||
+          variables.oldAnn.exists(k,
+              k.startsWith('machine.openshift.io') &&
+              !(k in variables.newAnn)
+          )
+        )
+      message: "Cannot modify or delete any machine.openshift.io/* annotation."
```
damdo, line 110: Aren't there any k8s.io annotations?
IIRC we have the paused one for example. This is MAPI I know so that might not apply here, but might be worth checking

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
...
+    # Param-controlled labels (labels on the CAPI machine) may change only to match the value on the CAPI Machine
+    - expression: >
+        !variables.paramLabels.exists(k,
+          ((k in variables.newLabels ? variables.newLabels[k] : null) !=
+           (k in variables.oldLabels ? variables.oldLabels[k] : null))
+          &&
+          ((k in variables.newLabels ? variables.newLabels[k] : null) !=
+           variables.paramLabels[k])
+        )
+      message: "Cannot modify a Cluster API controlled label except to match the Cluster API mirrored machine."
```
damdo, line 121: Hm not sure I follow here.

manifests/0000_30_cluster-api_13_vap.yaml
```diff
@@ -0,0 +1,121 @@
+apiVersion: admissionregistration.k8s.io/v1
```
damdo, line 1: As we talked about on that Slack thread, we can't deploy these like so, we'll need to put them in a transport configmap for conditional deploy based on the MAPI Migration feature gate.

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+			policy        *admissionregistrationv1.ValidatingAdmissionPolicy
+		)
+
+		BeforeEach(func() {
+			By("Unmarshalling the VAP/VAPB yamls")
+
+			policyBinding = &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}
+			policy = &admissionregistrationv1.ValidatingAdmissionPolicy{}
+
+			manifests, err := os.Open("../../../manifests/0000_30_cluster-api_13_vap.yaml")
```
damdo, line 776: This will need to change once we decide where to put them

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: yep, but IIRC I chose this as on a recent IPI install we already did

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: I think this is the best approach, yep. 

Because all expressions will be evaluated, and need to pass true (and e.g if fields don't exist on one of the CRDs we'll hit a rutime error in CEL) trying to handle both in one VAP would introduce a load of extra complexity to an already complex VAP 😂 

So in the interests of simplicity, sanity and maintainability I think we should stick to one per CR

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: I looked into this yesterday btw (and with the help of our AI friends) 

This would seemingly be the (only single expression) we would need to VAP two CRDs that have one embedded in the other (e.g like with machines/ machineset templates) 

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: forbid-foo-everywhere
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
    - apiGroups: ["example.com"]
      apiVersions: ["v1"]
      resources: ["foos","bars"]        # both kinds
      operations: ["CREATE","UPDATE"]
  validations:
  - # choose the right test based on the target kind
    expression: |
      request.resource.resource == 'foos'
        ? !has(object.spec.foo)
        : request.resource.resource == 'bars'
          ? !has(object.spec.template.spec.foo)
          : true
    message: 'Foo must not be set (directly or via template)'
```

Which I think (even in this relatively simple) example is way more complicated than an expression `!has(object.spec.foo)`

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: Ah yep - I'll add this to my list.

I'm also not sure (from the card here) on what these exactly entail

```
Labels and Annotations that are not material to operation
  Are not important for operation of Cluster API or Machine API
```

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: So this is enforcing: `Are not being modified to not match the Authoritative resource`

Which I've interpreted as: `Are not being modified to not match the Authoritative resource === Are being modified away from matching Authoritative`

manifests/0000_30_cluster-api_13_vap.yaml
damdo: Yeah makes sense to me

manifests/0000_30_cluster-api_13_vap.yaml
damdo: You mean for doing both machine and machinesets right?
Not for the label/name of the namespace? 🤔 Did you mean it to post on the thread above? Or am I confused

manifests/0000_30_cluster-api_13_vap.yaml
damdo: Not sure either, but probably best to not lock it down too much

manifests/0000_30_cluster-api_13_vap.yaml
damdo: Could you provide an example as well, I think it would help me understand better 😄

manifests/0000_30_cluster-api_13_vap.yaml
damdo: See: https://github.com/openshift/cluster-capi-operator/pull/321/files#r2197472127

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+			// Status controller doesn't run in envtest, its part of KCM - we've got to sleep?
+
+			// 'On my machine' 1s seems to be about right for the time required
+			// for the VAP to be loaded into the API Server
+			time.Sleep(1 * time.Second)
+
+		})
+
+		AfterEach(func() {
+			By("Removing the VAP and it's binding")
```
damdo, line 844:
```suggestion
			By("Removing the VAP and its VAP binding")
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+			Expect(k8sClient.Delete(ctx, policyBinding)).Should(Succeed())
+			Expect(k8sClient.Delete(ctx, policy)).Should(Succeed())
+		})
+
+		Context("with AuthoritativeAPI Machine API", func() {
```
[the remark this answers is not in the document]
damdo: I would change also the other Contexts to make it clear it is set in status already.

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				expectVAPError(err, "You may only modify spec.authoritativeAPI")
+			})
+
+			It("updating the spec.authoritativeAPI should be allowed", func() {
+				Eventually(k.Update(mapiMachine, func() {
+					mapiMachine.Spec.AuthoritativeAPI = machinev1beta1.MachineAuthorityMachineAPI
+				}), timeout).Should(Succeed())
+			})
+
+			Context("protected labels", func() {
```
damdo, line 891:
```suggestion
			Context("when trying to update metadata.labels", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				It("rejects modification of a machine.openshift.io label", func() {
```
damdo, line 892:
```suggestion
				It("rejects modification of the protected machine.openshift.io label", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+						err = k.Update(mapiMachine, func() {
+							mapiMachine.Labels["machine.openshift.io/instance-type"] = "m5.large"
+						})()
+
+						return err
+					}, timeout).ShouldNot(Succeed())
+					expectVAPError(err, "Cannot modify or delete any machine.openshift.io/* or kubernetes.io/* label")
+				})
+
+				It("rejects deletion of a machine.openshift.io label", func() {
```
damdo, line 904:
```suggestion
				It("rejects deletion of the protected machine.openshift.io label", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+						err = k.Update(mapiMachine, func() {
+							delete(mapiMachine.Labels, "machine.openshift.io/instance-type")
+						})()
+
+						return err
+					}, timeout).ShouldNot(Succeed())
+					expectVAPError(err, "Cannot modify or delete any machine.openshift.io/* or kubernetes.io/* label")
+				})
+
+				It("allows modification of an unrelated label", func() {
```
damdo, line 916:
```suggestion
				It("allows modification of a non-protected label", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+						err = k.Update(mapiMachine, func() {
+							delete(mapiMachine.Annotations, "machine.openshift.io/instance-state")
+						})()
+
+						return err
+					}, timeout).ShouldNot(Succeed())
+					expectVAPError(err, "Cannot modify or delete any machine.openshift.io/* annotation")
+				})
+
+				It("allows modification of an unrelated annotation", func() {
```
damdo, line 948:
```suggestion
				It("allows modification of a non-protected annotation", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+					Eventually(k.Update(mapiMachine, func() {
+						mapiMachine.Annotations["bar"] = "baz"
+					}), timeout).Should(Succeed())
+				})
+			})
+
+			Context("Cluster API controlled labels", func() {
```
damdo, line 955:
```suggestion
			Context("when trying to update Cluster API owned metadata.labels", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				It("allows changing a label to match the param machine", func() {
```
damdo, line 956:
```suggestion
				It("allows changing a metadata label to match the param machine", func() {
```

pkg/controllers/machinesync/suite_test.go
```diff
@@ -63,9 +65,48 @@ var _ = BeforeSuite(func() {
 
 	logf.SetLogger(textlogger.NewLogger(textlogger.NewConfig()))
 
+	// This is useful for debugging VAP related issues
```
damdo, line 68: Could we add below this comment instructions on how to read this debugging info for posterity?

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: Yeah, I've not really covered this outside of cluster-x or machine.openshift.io labels /annotations.

cc @JoelSpeed I'm probably missing something

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: I posted on the wrong thread 🤦

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: I looked into this yesterday btw (and with the help of our AI friends) 

This would seemingly be the (only single expression) we would need to VAP two CRDs that have one embedded in the other (e.g like with machines/ machineset templates) 

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: forbid-foo-everywhere
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
    - apiGroups: ["example.com"]
      apiVersions: ["v1"]
      resources: ["foos","bars"]        # both kinds
      operations: ["CREATE","UPDATE"]
  validations:
  - # choose the right test based on the target kind
    expression: |
      request.resource.resource == 'foos'
        ? !has(object.spec.foo)
        : request.resource.resource == 'bars'
          ? !has(object.spec.template.spec.foo)
          : true
    message: 'Foo must not be set (directly or via template)'
```

Which I think (even in this relatively simple) example is way more complicated than an expression `!has(object.spec.foo)` 

(This was meant to go here)

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -733,6 +759,238 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+			Expect(manifests.Close()).To(Succeed())
+
+			By("Updating the namespaces in the binding")
+			// Set the label on the mapi namespace so the
+			// selector applies the VAP to machines in it
+			Eventually(k.Update(mapiNamespace, func() {
+				mapiNamespace.SetLabels(map[string]string{
+					"name": "openshift-machine-api",
+				})
+			})).Should(Succeed())
```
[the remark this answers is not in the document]
theobarberbany: I'll give this a go. 

In tests, we still have the problem of needing to set the namespace label manually though - as we use `GenerateName`, so the namespace name isn't predictable.

manifests/0000_30_cluster-api_13_vap.yaml
JoelSpeed: You may want to look at creating variables and composing those, as you can then create policies that are much clearer that cover multiple similar resources. For example, I've seen VAPs cover Deployments, DameonSets and Statefulsets with just slightly different ways they set the variables up

theobarberbany: Ooh - I'd not considered that, do you have any examples?

pkg/controllers/machinesync/machine_sync_controller_test.go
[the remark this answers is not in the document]
damdo: Any movement on this front? :)

pkg/controllers/machinesync/suite_test.go
theobarberbany: Have moved to testutils package. PTAL :)

damdo:
@theobarberbany I think we could move this to a draft PR until we nail the transport configmap bits, WDYT

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -738,6 +765,258 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+			clusterAPIAdmissionPolicies := &corev1.ConfigMap{}
+			clusterAPICustomAdmissionPolicies := &corev1.ConfigMap{}
+
+			policyBinding = &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}
+			policy = &admissionregistrationv1.ValidatingAdmissionPolicy{}
+
+			configMaps, err := os.Open("../../../manifests/0000_30_cluster-api_09_admission-policies.yaml")
+			Expect(err).ToNot(HaveOccurred())
+
+			// Order is important as the yaml decoder.Decode unmarshals
```
theobarberbany, line 788: Given we likely expect to ship a bunch of across the `cluster-api-custom-admission-policies` config map, I think relying on ordering will become increasingly brittle. 

Maybe we can decode into a client.Object, read stream, and look for the the `type` and `name` that we want?

cc @damdo

damdo: Agree, good idea!

theobarberbany: Could also store things in a fake client / map to make getting easy

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,194 @@
...
+    
+        # Param-controlled labels (labels on the CAPI machine) may change only to match the value on the CAPI Machine
+        - expression: >
+            !variables.paramLabels.exists(k,
+              ((k in variables.newLabels ? variables.newLabels[k] : '') !=
+               (k in variables.oldLabels ? variables.oldLabels[k] : ''))
+              &&
+              ((k in variables.newLabels ? variables.newLabels[k] : '') !=
+               variables.paramLabels[k])
+            )
```
JoelSpeed, line 155: What if you went for `variables.paramLabels.all(k, variables.newLables[?k].hasValue() ? variables.newLables[k] == variables.paramLabels[k] : !variables.oldLabels[?k].hasValue())`

Check each label in the CAPI side is present and has the same value as the CAPI side, or, is missing in the old representation? Would that achieve the same thing? Would it be more succinct? 

It means you don't have to have the `orValue('')` style you have now which means you don't need to worry about whether a label is set deliberately to the empty string or not?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -738,6 +765,258 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				expectVAPError(err, "You may only modify spec.authoritativeAPI")
+			})
+
+			It("updating the spec.authoritativeAPI should be allowed", func() {
+				Eventually(k.Update(mapiMachine, func() {
+					mapiMachine.Spec.AuthoritativeAPI = machinev1beta1.MachineAuthorityMachineAPI
+				}), timeout).Should(Succeed(), "expected success when updating spec.authoritativeAPI")
+			})
+
+			Context("when trying to update metadata.labels", func() {
```
JoelSpeed, line 917: Test a label where the original value is `""` and one where you take a non empty value and try to change it to `""`

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,194 @@
...
+            ])
+        - name: check-authoritativeAPI-clusterapi
+          expression: 'object.status.authoritativeAPI == "ClusterAPI"'
+        - name: check-param-match
+          expression: 'object.metadata.name == params.metadata.name'
+    
+      variables:
+        # label maps
+        - name: newLabels
+          expression: "object.metadata.?labels.hasValue() ? object.metadata.?labels.value()     : {}"
```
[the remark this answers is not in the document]
theobarberbany: TIL this is a nice macro/ helper func! :D

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: Sorry - lost this! 

e.g 

CAPI (param) has labels `"foo":"bar"`, this would allow our MAPI to set `"foo":"bar"`, or to change from whatever was previously set to `"foo":"bar"`, but it would disallow any deviation from whatever is set on the CAPI label, so moving to `"foo":"baz"` would be denied. 

So we disallow common labels across machines from deviating - no shifting away from authority.

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,194 @@
...
+    metadata:
+      name: machine-api-machine-vap
+    spec:
+      matchResources:
+        namespaceSelector:
+          matchLabels:
+            name: openshift-machine-api
+      paramRef:
+        namespace: openshift-cluster-api
+        parameterNotFoundAction: Deny
```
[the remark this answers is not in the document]
theobarberbany: Hmm - maybe we want to `Allow` if we don't find the param. Are there any negative impacts of that?  Maybe this is the source of @sunzhaohua2 's bug. with MachineSet scaleups.

manifests/0000_30_cluster-api_13_vap.yaml
damdo: Ok got it thanks!

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,194 @@
...
+          expression: 'object.metadata.name == params.metadata.name'
+    
+      variables:
+        # label maps
+        - name: newLabels
+          expression: "object.metadata.?labels.hasValue() ? object.metadata.?labels.value()     : {}"
+        - name: oldLabels
+          expression: "oldObject.metadata.?labels.hasValue()   ? oldObject.metadata.?labels.value()   : {}"
+        - name: paramLabels
+          expression: "params.metadata.labels"
```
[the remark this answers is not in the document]
theobarberbany: My initial thinking here was yes, as the CAPI controllers will be authoritative - and they set some default labels. 

I can see potential for this to fail though, e.g if we mirror MAPI -> CAPI, switch authority and then try to make another change (before the CAPI controllers populate labels). 

I'll update this to check in the same way the others do.

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,194 @@
...
+            ].all(p,
+              p[0].hasValue()
+                ? p[1].hasValue() && p[0].value() == p[1].value()
+                : !p[1].hasValue()
+            )
+    
+      # All validations must evaluate to TRUE
+      validations:
+        # Only spec.authoritativeAPI may change
+        - expression: "object.spec == oldObject.spec || variables.specLockedExceptAuthoritativeAPI"
```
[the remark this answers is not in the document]
JoelSpeed: Logically, it shouldn't, that implies an issue with `specLockedExceptAuthoritativeAPI` doesn't it?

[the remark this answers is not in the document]
theobarberbany: Yeah, removing the clause at the top + the `object.spec == oldObject.spec` is equivalent - and works. Have updated.

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,194 @@
...
+          message:  "You may only modify spec.authoritativeAPI. Any other change inside .spec is not allowed."
+    
+        # Guard machine.openshift.io/* and kubernetes.io/* labels
+        - expression: >
+            !(
+              variables.newLabels.exists(k,
+                  (k.startsWith('machine.openshift.io') || k.startsWith('kubernetes.io')) &&
+                  ((k in variables.oldLabels ? variables.oldLabels[k] : '') != variables.newLabels[k])
```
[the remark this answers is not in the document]
theobarberbany: It works 🎉 ! 

Also works with 'null' - hopefully typechecking is happy and we've killed 2 birds with one stone :D

manifests/0000_30_cluster-api_09_admission-policies.yaml
theobarberbany: Yep, I think this makes sense. 

For review context: What we're changing from is: `there exists no key in parameter labels, where that key also exists and is changing in newLabels, and that key in newLabels doesn't match the key in parameter labels`

Which... grokking `Check each label in the CAPI side is present and has the same value as the CAPI side, or, is missing in the old representation` I agree that's going to do the same thing.

theobarberbany: Actually... they're subtly different? (I think)

> where that key also exists and is changing in newLabels,

With my original expression, we allow mismatches e.g between `param[k]` and `newLabels[k]` when we don't change `oldLabels[k]`. 

The new one, because of the `paramLabels.all(k` would force users to either: 

Update the value of the label on the MAPI Machine to match the param (`paramLabels[k] == newLabels[k]`), or deny the policy..

Are we ok with that?

manifests/0000_30_cluster-api_09_admission-policies.yaml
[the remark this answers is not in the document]
theobarberbany: Have updated `parameterNotFoundAction: True`

pkg/controllers/machinesync/machine_sync_controller_test.go
theobarberbany: Using the new `variables.oldLabels[?k].orValue(null)` style (with null) I think fixes all of these issues. 

Added tests anyway.

manifests/0000_30_cluster-api_09_admission-policies.yaml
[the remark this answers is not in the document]
theobarberbany: Looks like the most recent CI run failed because of this: 

```
2025-07-22T17:34:30.567749685Z E0722 17:34:30.567729       1 controller.go:341] "msg"="Reconciler error" "error"="machines.machine.openshift.io \"ci-op-rt78xjh8-c3c99-zljqf-master-0\" is forbidden: ValidatingAdmissionPolicy 'machine-api-machine-vap' with binding 'machine-api-machine-vap' denied request: failed to configure binding: no params found for policy binding with `Deny` parameterNotFoundAction" "controller"="machine-controller" "name"="ci-op-rt78xjh8-c3c99-zljqf-master-0" "namespace"="openshift-machine-api" "object"={"name":"ci-op-rt78xjh8-c3c99-zljqf-master-0","namespace":"openshift-machine-api"} "reconcileID"="a41fe412-b610-4efe-bde9-187e6f62ea9b"

```

The exemption: 

```yaml
 - name: check-only-non-service-account-requests
          expression: >-
            !(request.userInfo.username in [
                "system:serviceaccount:openshift-machine-api:machine-api-controllers",
                "system:serviceaccount:openshift-cluster-api:cluster-capi-operator"
            ])
```

must've not even been executing because we were denying on the parameter not being found.

theobarberbany:
cc @damdo @JoelSpeed 

The installer controller is no longer erroring when reconciling our `cluster-api-custom-admission-policies` transport config map, and the job [succeeds](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/321/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1948143603679236096). However, looking at the must gather (cluster scoped resources -> admissionregistration.k8s.io -> validatingadmissionpolicies) I don't see the VAP/VAPB. 

```bash
2025-07-23T23:28:13.660931264Z I0723 23:28:13.660899       1 capi_installer_controller.go:148] "processing CAPI provider ConfigMap" logger="CapiInstallerController" controller="CapiInstallerController" controllerGroup="config.openshift.io" controllerKind="ClusterOperator" ClusterOperator="cluster-api" namespace="" name="cluster-api" reconcileID="6de228c9-b374-4293-b2a0-3167aaf5547d" configmapName="cluster-api-custom-admission-policies" providerType="core" providerName="openshift-cluster-api" providerVersion="v0.0.0"
```

I'm pretty sure it's actually being applied, as we were failing a previous run due to the misconfiguration of `        parameterNotFoundAction: Deny` where the MAPI machine controllers couldn't reconcile anything as they were blocked by the VAP.

Is there an update needed to the way we build the must-gather to pick these up?

manifests/0000_30_cluster-api_09_admission-policies.yaml
JoelSpeed: Ahh right, yeah, so allowing it to not change makes sense. I think I was thinking that the CAPI controllers would have updated the labels in production, but this seems safe and makes sense to ratchet, LGTM

manifests/0000_30_cluster-api_13_vap.yaml
theobarberbany: Not sure if 'these' is the service accounts, or the entire `matchConditions` block. Here's the explanation :) 

- Anything migration or mapi controller is excluded, we trust them.
- We only care (for this VAP) if the MAPI machine is non-authoritative.
- The final one is basically a filter, the VAP binding will provide every matching object of the correct kind to the VAP policy. This means we match the correct instance of our parameter and therefore evaluate the policy.

JoelSpeed:
> Is there an update needed to the way we build the must-gather to pick these up?

Must gather picks up the resource to copy down based on the relatedObjects in the cluster operator, so we might need to update the output of relatedObjects

pkg/admissionpolicy/testutils/util.go
```diff
@@ -0,0 +1,261 @@
...
+
+	By("Unmarshalling the admission policies in each configmap")
+
+	// each ConfigMap produces a list of client.Objects obtained from that configMap
+	mapObjs := map[string][]client.Object{}
+
+	for _, configMap := range configMapByName {
+		objs := []client.Object{}
+
+		if components, ok := configMap.Data["components"]; ok && len(components) > 0 { //nolint: nestif
```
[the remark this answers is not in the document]
theobarberbany: I think we get linter complaints, as we're not in a `_test.go`, and are just a normal package.

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,165 @@
...
+      # Requests must satisfy every matchCondition to reach the validations
+      matchConditions:
+        - name: check-only-non-service-account-requests
+          expression: >-
+            !(request.userInfo.username in [
+                "system:serviceaccount:openshift-machine-api:machine-api-controllers",
+                "system:serviceaccount:openshift-cluster-api:cluster-capi-operator"
+            ])
+        - name: check-authoritativeAPI-clusterapi
+          expression: 'object.status.authoritativeAPI == "ClusterAPI"'
```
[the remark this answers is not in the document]
theobarberbany: Ah yep, that makes sense!

pkg/admissionpolicy/testutils/util.go
```diff
@@ -0,0 +1,261 @@
...
+	// ClusterAPIAdmissionPolicies is the name of the ClusterAPIAdmissionPolicies transport config map.
+	ClusterAPIAdmissionPolicies string = "cluster-api-admission-policies"
+
+	// ClusterAPICustomAdmissionPolicies is the name of the ClusterAPICustomAdmissionPolicies transport config map.
+	ClusterAPICustomAdmissionPolicies string = "cluster-api-custom-admission-policies"
+
+	// ClusterAPIAWSAdmissionPolicies is the name of the ClusterAPIAWSAdmissionPolicies transport config map.
+	ClusterAPIAWSAdmissionPolicies string = "cluster-api-aws-admission-policies"
+
+	yamlChunk = 4 << 10
```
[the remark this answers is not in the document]
theobarberbany: 4×2^10=4096. Having a const of 4096 almost certainly is clearer, but I thought this was cool 😂

manifests/0000_30_cluster-api_09_admission-policies.yaml
```diff
@@ -0,0 +1,167 @@
...
+      name: machine-api-machine-vap
+    spec:
+      matchResources:
+        namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: openshift-machine-api
+      paramRef:
+        namespace: openshift-cluster-api
+        # We 'Allow' here as we don't want to block MAPI Machine 
+        # functionality when no CAPI machine (param) exists.
```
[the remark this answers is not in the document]
theobarberbany: Yep, will add.

go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/metal3-io/cluster-api-provider-metal3/api v1.8.6
 	github.com/onsi/ginkgo/v2 v2.22.2
 	github.com/onsi/gomega v1.36.2
-	github.com/openshift/api v0.0.0-20250320170726-75d64d71980b
+	github.com/openshift/api v0.0.0-20250529074221-97812373b6b4
```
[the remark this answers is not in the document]
theobarberbany: Ah - i pulled `release-4.19` as i'd not rebased yet, and then rebased. Will update.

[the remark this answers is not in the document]
theobarberbany: cc @damdo  - is bumping in this PR ok ?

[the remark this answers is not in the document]
theobarberbany: https://github.com/openshift/cluster-capi-operator/pull/341 will fix, and ill rebase before merge here.

pkg/admissionpolicy/testutils/vap_status_controller.go
```diff
@@ -0,0 +1,257 @@
...
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+// ControllerName has "Status" in it to differentiate this controller with the other that runs in API server.
+const ControllerName = "validatingadmissionpolicy-status"
+
+// Controller is the ValidatingAdmissionPolicy Status controller that reconciles the Status field of each policy object.
+// This controller runs type checks against referred types for each policy definition.
```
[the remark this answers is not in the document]
theobarberbany: Hmm, this is copypasta of the upstream code. I'm using it as a drop in - don't think we should change it? There must be another one for MAPs already?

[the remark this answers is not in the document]
theobarberbany: Hmm, this is copypasta of the upstream code. I'm using it as a drop in - don't think we should change it? There must be another one for MAPs already?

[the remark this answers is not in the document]
theobarberbany: Not without vendoring all of k/k :(

pkg/controllers/machinesync/suite_test.go
```diff
@@ -72,6 +75,13 @@ var _ = BeforeSuite(func() {
 	Expect(cfg).NotTo(BeNil())
 	Expect(k8sClient).NotTo(BeNil())
 
+	testScheme = scheme.Scheme
+
+	// Start VAP Status Controller
+	By("starting the ValidatingAdmissionPolicy status controller")
+	vapCleanup, err = admissiontestutils.StartVAPStatusController(ctx, cfg, testScheme)
+	Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
theobarberbany: Q: How is this different from calling `vapCleanup` in an `AfterSuite` block?

pkg/operatorstatus/operator_status.go
```diff
@@ -209,6 +209,8 @@ func (r *ClusterOperatorStatusClient) RelatedObjects() []configv1.ObjectReferenc
 		{Group: "cluster.x-k8s.io", Resource: "clusters", Namespace: r.ManagedNamespace},
 		{Group: "cluster.x-k8s.io", Resource: "machines", Namespace: r.ManagedNamespace},
 		{Group: "cluster.x-k8s.io", Resource: "machinesets", Namespace: r.ManagedNamespace},
+		{Group: "admissionregistration", Resource: "validatingadmissionpolicy"},
+		{Group: "admissionregistration", Resource: "validatingadmissionpolicybinding"},
```
damdo, line 213: It might be best to specify the resource names instead of catching all the VAP(B) in the cluster?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -738,6 +754,276 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				newObj, ok := obj.DeepCopyObject().(client.Object)
+				Expect(ok).To(BeTrue())
+
+				err := k8sClient.Create(ctx, newObj)
+				if err != nil && !apierrors.IsAlreadyExists(err) {
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
+
+			By("waiting for VAP to be ready")
```
damdo, line 780:
```suggestion
			By("Waiting for VAP to be ready")
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -738,6 +754,276 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+
+			It("updating the spec should be allowed", func() {
+				Eventually(k.Update(mapiMachine, func() {
+					mapiMachine.Spec.ObjectMeta.Labels = map[string]string{"foo": "bar"}
+				}), timeout).Should(Succeed(), "expected success when updating the spec")
+			})
+
+		})
+
+		Context("with status.AuthoritativeAPI: Cluster API", func() {
```
damdo, line 900:
```suggestion
		Context("with status.AuthoritativeAPI: ClusterAPI", func() {
```

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -738,6 +754,276 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+			mapiMachine = mapiMachineBuilder.WithName("test-machine").WithLabels(map[string]string{
+				"machine.openshift.io/cluster-api-cluster":      "ci-op-gs2k97d6-c9e33-2smph",
+				"machine.openshift.io/cluster-api-machine-role": "worker",
+				"machine.openshift.io/cluster-api-machine-type": "worker",
+				"machine.openshift.io/cluster-api-machineset":   "ci-op-gs2k97d6-c9e33-2smph-worker-us-west-2b",
+				"machine.openshift.io/instance-type":            "m6a.xlarge",
+			}).WithAnnotations(map[string]string{
+				"machine.openshift.io/instance-state": "running",
+			}).Build()
+			Expect(k8sClient.Create(ctx, mapiMachine)).Should(Succeed())
```
damdo, line 851: Should we use Eventually?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -738,6 +754,276 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				"machine.openshift.io/cluster-api-machine-type": "worker",
+				"machine.openshift.io/cluster-api-machineset":   "ci-op-gs2k97d6-c9e33-2smph-worker-us-west-2b",
+				"machine.openshift.io/instance-type":            "m6a.xlarge",
+				"cluster.x-k8s.io/cluster-name":                 "ci-op-gs2k97d6-c9e33-2smph",
+				"cluster.x-k8s.io/set-name":                     "ci-op-gs2k97d6-c9e33-2smph-worker-us-west-2b",
+				"node-role.kubernetes.io/worker":                "",
+			}).WithAnnotations(map[string]string{
+				"machine.openshift.io/instance-state": "running",
+			}).Build()
+			Expect(k8sClient.Create(ctx, capiMachine)).Should(Succeed())
```
damdo, line 866: Should we use Eventually?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -738,6 +754,276 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+			By("Applying the objects found in clusterAPICustomAdmissionPolicies")
+			for _, obj := range transportConfigMaps[admissiontestutils.ClusterAPICustomAdmissionPolicies] {
+				// Deep‑copy so the loop variable isn’t mutated by the client
+				newObj, ok := obj.DeepCopyObject().(client.Object)
+				Expect(ok).To(BeTrue())
+
+				err := k8sClient.Create(ctx, newObj)
+				if err != nil && !apierrors.IsAlreadyExists(err) {
+					Expect(err).NotTo(HaveOccurred())
+				}
```
damdo, line 777: Should we use Eventually? Not sure how feasible it is with the already exists error being there

pkg/controllers/machinesync/suite_test.go
```diff
@@ -72,6 +75,13 @@ var _ = BeforeSuite(func() {
 	Expect(cfg).NotTo(BeNil())
 	Expect(k8sClient).NotTo(BeNil())
 
+	testScheme = scheme.Scheme
+
+	// Start VAP Status Controller
+	By("starting the ValidatingAdmissionPolicy status controller")
```
damdo, line 81:
```suggestion
	By("Starting the ValidatingAdmissionPolicy status controller")
```

damdo (commented):
A bunch of nits but looking good!

pkg/controllers/machinesync/machine_sync_controller_test.go
theobarberbany: https://github.com/openshift/cluster-capi-operator/pull/321/files#r2257610280

pkg/controllers/machinesync/machine_sync_controller_test.go
theobarberbany: Have updated, I think it works :D

pkg/operatorstatus/operator_status.go
theobarberbany: I've updated with explicitly set names now, let's tackle this when we add more / in another PR :)

PR #323, opened 2025-06-24
OCPCLOUD-2992: add machineset migration e2e utils

sunzhaohua2:


damdo (commented):
@sunzhaohua2  Looks very good, left mostly comments about adding retry to make the suite more robust to brief failures, and some notes on error wrapping messages

Thanks!

e2e/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/cluster-capi-operator/e2e
 
-go 1.22.7
+go 1.23.0
+
+toolchain go1.24.3
```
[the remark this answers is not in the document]
damdo: The ART tooling that does the building actually ignores them as it sets `TOOLCHAIN=local` for the go compiler.
They shouldn't hurt us, but they are meaningless in our use case, so better to leave them out if possible yeah

e2e/framework/machine.go
```diff
@@ -54,3 +60,46 @@ func FilterRunningMachines(machines []*clusterv1.Machine) []*clusterv1.Machine {
...
 }
+
+// GetMachine get a machine by its name from the cluster API namespace.
+func GetMachine(cl client.Client, name string) (*clusterv1.Machine, error) {
+	machine := &clusterv1.Machine{}
+	key := client.ObjectKey{Namespace: CAPINamespace, Name: name}
+
+	if err := cl.Get(ctx, key, machine); err != nil {
+		return nil, fmt.Errorf("error querying api for machine object: %w", err)
+	}
```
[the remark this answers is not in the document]
theobarberbany: Komega + eventually gives us retries for free, I'd argue for getting machines we just use that?

e2e/framework/machine.go
```diff
@@ -54,3 +60,46 @@ func FilterRunningMachines(machines []*clusterv1.Machine) []*clusterv1.Machine {
...
+
+	return machine, nil
+}
+
+
+// DeleteMachines deletes the specified machines and returns an error on failure.
+func DeleteMachines(cl client.Client, machines ...*clusterv1.Machine) error {
```
theobarberbany, line 78: What about using the gomega primitives here? I think it's more readable + will keep our e2es looking like our units / integrations (less mental load to debug / review)

```go
for _, machine := range machines {
  Eventually(func() error {
    return cl.Delete(ctx, machine)
  }, time.Minute, RetryShort).Should(SatisfyAny(
    Succeed(),
    WithTransform(apierrors.IsNotFound, BeTrue()),
  ), "Delete should succeed, or machine should not be found")
}
```

theobarberbany: This `Eventually` also removes the need for the wait func below (no longer needed)

e2e/framework/machineset.go
```diff
@@ -190,3 +200,76 @@ func GetMachinesFromMachineSet(cl client.Client, machineSet *clusterv1.MachineSe
...
+}
+
+// ScaleMachineSet scales a machineSet with a given name to the given number of replicas.
+func ScaleMachineSet(name string, replicas int) error {
+	scaleClient, err := getScaleClient()
+	if err != nil {
+		return fmt.Errorf("error calling getScaleClient %w", err)
+	}
+
+	scale, err := scaleClient.Scales(CAPINamespace).Get(ctx, schema.GroupResource{Group: "cluster.x-k8s.io", Resource: "MachineSet"}, name, metav1.GetOptions{})
```
theobarberbany, line 233: why are we hardcoding the namespace?

e2e/framework/machineset.go
```diff
@@ -190,3 +200,76 @@ func GetMachinesFromMachineSet(cl client.Client, machineSet *clusterv1.MachineSe
...
+// GetLatestMachineFromMachineSet returns the new created machine by a given machineSet.
+func GetLatestMachineFromMachineSet(cl client.Client, machineSet *clusterv1.MachineSet) (*clusterv1.Machine, error) {
+	machines, err := GetMachinesFromMachineSet(cl, machineSet)
+	if err != nil {
+		return nil, fmt.Errorf("error getting machines: %w", err)
+	}
+
+	var machine *clusterv1.Machine
+
+	newest := time.Date(2020, 0, 1, 12, 0, 0, 0, time.UTC)
```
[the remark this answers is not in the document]
sunzhaohua2: I copied from cluster-api-actuator-pkg :) https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/framework/machinesets.go#L400

e2e/framework/machine.go
```diff
@@ -54,3 +58,52 @@ func FilterRunningMachines(machines []*clusterv1.Machine) []*clusterv1.Machine {
...
+	machine := &clusterv1.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+
+	Eventually(komega.Get(machine), time.Minute, RetryShort).Should(SatisfyAny(
+		Succeed(),
+		WithTransform(apierrors.IsNotFound, BeTrue()),
```
JoelSpeed, line 73: If we just created a machine, we wouldn't want to have it return `NotFound` right? What's the use case for this util if it will return an empty Machine when the Machine isn't found?

e2e/framework/machineset.go
```diff
@@ -166,12 +168,24 @@ func WaitForMachineSet(cl client.Client, name string) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
 
-	Expect(cl.Get(ctx, key, machineSet)).To(Succeed())
+	Eventually(komega.Get(machineSet), time.Minute, RetryShort).Should(SatisfyAny(
+		Succeed(),
+		WithTransform(apierrors.IsNotFound, BeTrue()),
```
JoelSpeed, line 186: Again, if we are fetching an object, why would we want to allow it to not be created?

e2e/framework/machinetemplate.go
```diff
@@ -0,0 +1,83 @@
...
+	awsMachineTemplate := &awsv1.AWSMachineTemplate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+
+	Eventually(komega.Get(awsMachineTemplate), time.Minute, RetryShort).Should(SatisfyAny(
+		Succeed(),
+		WithTransform(apierrors.IsNotFound, BeTrue()),
```
JoelSpeed, line 31: Are we sure we want to allow the object to not be found?

e2e/framework/machineset.go
sunzhaohua2: For AfterAll clean resource, before I first GetMachineSet then delete, if machineset already was deleted in testing, will return 404, so add this.  I will remove this and update cleanup step, we can DeleteMachineSets directly.

JoelSpeed:
Units failed, but code changes LGTM

PR #339, opened 2025-07-29
NO-JIRA: Get openshift/api directory from packages

mdbooth:
This causes the envtests to automatically use a locally overridden openshift/api package. It continues to use the vendor directory by default while we use vendoring, but will also continue working seamlessly if we ever stop using vendoring.


mdbooth:
Fixed lint. unit test failure looked unrelated, possibly a flake?

mdbooth:
The unit test failure above was this unrelated flake: https://issues.redhat.com/browse/OCPBUGS-59955

PR #340, opened 2025-07-30
NO-JIRA: Restructure machine_sync_controller_test to enable sticky scroll

mdbooth:
VSCode's sticky scroll feature only works for a file of ginkgo tests
when there is only a single top-level declaration.


damdo (commented):
Weird

/approve
/lgtm

mdbooth:
> Weird

Weird, indeed. It would be good to get to the bottom of exactly why this happens because this is an incredibly useful feature when navigating these tests.

PR #342, opened 2025-07-31
NO-JIRA: Improve ergonomics of running tests

mdbooth:
This PR is a roundup of several minor enhancements:

* VSCode: Default stickyScroll to foldingProviderModel

This has the advantage of working for Ginkgo tests, where it is incredibly useful. The default 'outline' model does not work there. It can still be overridden in user configuration.

* Allow test package override in make unit

This allows you to run `make unit TEST_DIRS=./pkg/controllers/capiinstaller/...` and it will run only those tests.

* Set KUBEBUILDER_ASSETS when running tests from VS Code

The KUBEBUILDER_ASSETS environment variable is required to run envtests. This change causes VS Code to read a local environment file called `.localtestenv` when running tests, and adds a make target to generate it.

Note that this currently requires manually running `make localtestenv` before attempting to invoke tests from VS Code. There's probably a way to modify the default launch configuration to do this automatically.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,9 +39,16 @@ migration:
 	# building migration
 	go build -o bin/machine-api-migration cmd/machine-api-migration/main.go
 
+.PHONY: localtestenv
+localtestenv: .localtestenv
+
+KUBEBUILDER_ASSETS ?= $(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin --index https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml)
+.localtestenv: Makefile
+	echo "KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS)" > $@
+
 TEST_DIRS ?= ./pkg/... ./manifests-gen/...
-unit:
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin --index https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml)" ./hack/test.sh "$(TEST_DIRS)" 10m
+unit: .localtestenv
+	./hack/test.sh "$(TEST_DIRS)" 10m
 
 .PHONY: e2e
 e2e:
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@
 # vendor/
 
 bin/
+/.localtestenv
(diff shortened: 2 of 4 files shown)


JoelSpeed:
> The KUBEBUILDER_ASSETS environment variable is required to run envtests.

Required, apart from the fact that it has a default value 🤔  With the changes you've proposed, will that break existing workflows that rely on the default value of `KUBEBUILDER_ASSETS`?

Makefile
```diff
@@ -39,8 +39,18 @@ migration:
...
+
+KUBEBUILDER_ASSETS ?= $(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin --index https://raw.githubusercontent.com/openshift/api/master/envtest-releases.yaml)
+.localtestenv: Makefile
+	echo "KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS)" > $@
+
+TEST_DIRS ?= ./pkg/... ./manifests-gen/...
+unit: .localtestenv
+	# Launch "./hack/test.sh ..." with additional environment variables
+	# loaded from .localtestenv
+	xargs -I '{}' -x -- env '{}' ./hack/test.sh "$(TEST_DIRS)" 5m < .localtestenv
```
damdo, line 53: We need to make sure this is compatible with standard macOS xargs (BSD) vs GNU.

mdbooth: @damdo Can you run it for me? `make unit` should do the trick.

damdo: I run nixpks one, so I use the gnu version by default. Better to ask @JoelSpeed or @theobarberbany

theobarberbany: `xargs -I '{}' -x env '{}' ./hack/test.sh "./pkg/... ./manifests-gen/..." 5m < .localtestenv` ?

damdo: What is breaking exactly? looks like both `-x` and `-I` are there?

nrb: Looks like it's the `--` separator

mdbooth: Unfortunately `-n1` wouldn't be correct here, as it would transform this into a separate test invocation for each local env var.

mdbooth: I preferred the 'just use gxargs on Mac' solution: it sounds much more maintainable. Shall we add something to the Makefile which detects `gxargs` and uses it in preference to `xargs` if it's present?

damdo: Sounds the next best solution then yes

damdo:
/override ci/prow/e2e-aws-ovn

This PR doesn't change payload code

PR #348, opened 2025-08-27
OCPCLOUD-2709: Implement MAPI2CAPI conversion of loadbalancers

RadekManak:
## Summary by CodeRabbit

* **New Features**
  * Runtime-configurable CAPI/MAPI namespaces; AWS infra now derives and populates primary + optional secondary control‑plane load balancers from provider specs and exposes conversions for LB types.

* **Bug Fixes**
  * Worker machines rejected for declaring LBs; control‑plane machines may use LBs. Improved control‑plane endpoint port handling and degraded-condition reporting when control plane is missing.

* **Refactor**
  * Provider-spec retrieval and namespace usage unified under controller context across platforms; namespaces propagated into created resources and logs.

* **Tests**
  * Expanded tests for namespaces, control‑plane discovery, LB ordering/validation, MachineSync conversion, and CAPI↔MAPI LB translations.

* **Chores**
  * Operator cache includes Machine objects; test env adds ControlPlaneMachineSet CRD.

### DIFF
--- a/pkg/controllers/infracluster/aws.go
+++ b/pkg/controllers/infracluster/aws.go
@@ -17,34 +17,50 @@ package infracluster
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/go-logr/logr"
 	cerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
 	awsv1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	mapiv1beta1 "github.com/openshift/api/machine/v1beta1"
+)
+
+var (
+	// ErrNoControlPlaneLoadBalancerConfigured indicates there is no control plane load balancer configuration present.
+	ErrNoControlPlaneLoadBalancerConfigured = errors.New("no control plane load balancer configured")
+	// ErrNilProviderSpec indicates a nil provider spec raw extension.
+	ErrNilProviderSpec = errors.New("provider spec is nil")
+	// ErrInvalidNumberOfControlPlaneLoadBalancers indicates an invalid number of control plane load balancers has been configured.
+	ErrInvalidNumberOfControlPlaneLoadBalancers = errors.New("invalid number of control plane load balancers")
+	// ErrUnsupportedLoadBalancerType indicates an unsupported load balancer type.
+	ErrUnsupportedLoadBalancerType = errors.New("unsupported load balancer type")
 )
 
 // ensureAWSCluster ensures the AWSCluster cluster object exists.
 func (r *InfraClusterController) ensureAWSCluster(ctx context.Context, log logr.Logger) (client.Object, error) {
-	target := &awsv1.AWSCluster{ObjectMeta: metav1.ObjectMeta{
+	awsCluster := &awsv1.AWSCluster{ObjectMeta: metav1.ObjectMeta{
 		Name:      r.Infra.Status.InfrastructureName,
-		Namespace: defaultCAPINamespace,
+		Namespace: r.CAPINamespace,
 	}}
 
 	// Checking whether InfraCluster object exists. If it doesn't, create it.
-
-	if err := r.Get(ctx, client.ObjectKeyFromObject(target), target); err != nil && !cerrors.IsNotFound(err) {
+	if err := r.Get(ctx, client.ObjectKeyFromObject(awsCluster), awsCluster); err != nil && !cerrors.IsNotFound(err) {
 		return nil, fmt.Errorf("failed to get InfraCluster: %w", err)
 	} else if err == nil {
-		return target, nil
+		return awsCluster, nil
 	}
 
-	log.Info(fmt.Sprintf("AWSCluster %s/%s does not exist, creating it", target.Namespace, target.Name))
+	log = log.WithValues("AWSCluster", klog.KObj(awsCluster))
+	log.Info("AWSCluster does not exist, creating it")
 
 	apiURL, err := url.Parse(r.Infra.Status.APIServerInternalURL)
 	if err != nil {
@@ -60,10 +76,35 @@ func (r *InfraClusterController) ensureAWSCluster(ctx context.Context, log logr.
 		return nil, fmt.Errorf("infrastructure PlatformStatus should not be nil: %w", err)
 	}
 
-	target = &awsv1.AWSCluster{
+	providerSpec, err := r.getAWSMAPIProviderSpec(ctx, r.Client)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain MAPI ProviderSpec: %w", err)
+	}
+
+	awsCluster, err = r.newAWSCluster(providerSpec, apiURL, int32(port))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get AWSCluster: %w", err)
+	}
+
+	if err := r.Create(ctx, awsCluster); err != nil {
+		return nil, fmt.Errorf("failed to create AWSCluster: %w", err)
+	}
+
+	log.Info("AWSCluster successfully created")
+
+	return awsCluster, nil
+}
+
+func (r *InfraClusterController) newAWSCluster(providerSpec *mapiv1beta1.AWSMachineProviderConfig, apiURL *url.URL, port int32) (*awsv1.AWSCluster, error) {
+	controlPlaneLoadBalancer, secondaryControlPlaneLoadBalancer, err := extractLoadBalancerConfigFromMAPIAWSProviderSpec(providerSpec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract control plane load balancer configuration: %w", err)
+	}
+
+	target := &awsv1.AWSCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      r.Infra.Status.InfrastructureName,
-			Namespace: defaultCAPINamespace,
+			Namespace: r.CAPINamespace,
 			// The ManagedBy Annotation is set so CAPI infra providers ignore the InfraCluster object,
 			// as that's managed externally, in this case by this controller.
 			Annotations: map[string]string{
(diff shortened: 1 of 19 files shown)


RadekManak:
@coderabbitai full review

pkg/controllers/infracluster/azure.go
```diff
@@ -51,7 +51,7 @@ func (r *InfraClusterController) ensureAzureCluster(ctx context.Context, log log
 	// Get the Azure Bootstrap Credentials Secret.
 	// This is created by the Cluster Credential Operator and should always exist. We expect to always find it, if not, we error.
 	capzManagerBootstrapSecret := &corev1.Secret{}
-	if err := r.Get(ctx, client.ObjectKey{Namespace: defaultCAPINamespace, Name: capzManagerBootstrapCredentials}, capzManagerBootstrapSecret); err != nil && !cerrors.IsNotFound(err) {
+	if err := r.Get(ctx, client.ObjectKey{Namespace: r.CAPINamespace, Name: capzManagerBootstrapCredentials}, capzManagerBootstrapSecret); err != nil && !cerrors.IsNotFound(err) {
 		return nil, fmt.Errorf("failed to get Azure Boostrap Credentials Secret: %w", err)
 	}
```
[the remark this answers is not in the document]
nrb: I don't think we want to change that behavior in the scope of this PR.

pkg/controllers/infracluster/aws.go
```diff
@@ -84,14 +122,77 @@ func (r *InfraClusterController) ensureAWSCluster(ctx context.Context, log logr.
...
+
+// convertMAPILoadBalancerTypeToCAPI converts MAPI AWSLoadBalancerType to CAPI LoadBalancerType.
+func convertMAPILoadBalancerTypeToCAPI(mapiType mapiv1beta1.AWSLoadBalancerType) awsv1.LoadBalancerType {
+	switch mapiType {
+	case mapiv1beta1.ClassicLoadBalancerType:
+		return awsv1.LoadBalancerTypeClassic
+	case mapiv1beta1.NetworkLoadBalancerType:
+		return awsv1.LoadBalancerTypeNLB
+	default:
+		// Default to classic for unknown types
```
theobarberbany, line 195: Should we log that we've made this assumption?

pkg/controllers/infracluster/aws.go
```diff
@@ -84,14 +124,91 @@ func (r *InfraClusterController) ensureAWSCluster(ctx context.Context, log logr.
...
+	if err := yaml.Unmarshal(rawProviderSpec, providerSpec); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal MAPI ProviderSpec: %w", err)
+	}
 
-	return target, nil
+	return providerSpec, nil
+}
+
+// extractLoadBalancerConfigFromMAPIProviderSpec extracts 1-2 control plane load balancers from a MAPI machine's provider spec.
+// The first load balancer is treated as the internal load balancer, and the second (if present) as the secondary one.
```
nrb, line 151: This is backwards, at least from what I intended when I wrote the code upstream. CAPA only made external LBs, and the goal was to allow users to create internal ones as a secondary option.

Did you find that OpenShift sets the first as internal and the second as internal? It's very possible that the installer does this, because the names chosen upstream don't rely convey the public/private split well. As long as we're consistent with how they're used, I don't think it matters too much.

pkg/controllers/infracluster/aws.go
```diff
@@ -84,14 +122,77 @@ func (r *InfraClusterController) ensureAWSCluster(ctx context.Context, log logr.
...
+
+		return &awsv1.AWSLoadBalancerSpec{
+			Name:             &lbPrimary.Name,
+			LoadBalancerType: convertMAPILoadBalancerTypeToCAPI(lbPrimary.Type),
+		}, nil, nil
+	case 2:
+		lbFirst := providerSpec.LoadBalancers[0]
+		lbSecond := providerSpec.LoadBalancers[1]
+		// Prefer the load balancer with "-int" suffix as primary when two are present.
+		if strings.HasSuffix(lbSecond.Name, "-int") && !strings.HasSuffix(lbFirst.Name, "-int") {
```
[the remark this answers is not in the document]
nrb: Yeah, ideally if/when CAPA's networking types are revisited, the load balancers can be either an array as Radek says, or we can have better field names that communicate internal/external.

iirc that I initially named the field that was being added to something like `InternalLoadBalancer`, but we decided to make it more generic upstream.

I think the actual consequence would not be their name, but whether the configuration on AWS is incorrect, like having 2 LBs that can't route externally. I'm not sure it's worth validating that here or not; that should be something the installer creates correctly and we shouldn't be modifying them. If someone or something else modifies them out of band, they'll likely have a broken cluster.

pkg/conversion/capi2mapi/aws.go
```diff
@@ -112,6 +115,11 @@ func (m machineAndAWSMachineAndAWSCluster) toProviderSpec() (*mapiv1.AWSMachineP
 		errors = append(errors, err)
 	}
 
+	mapiLoadBalancers, err := convertAWSClusterLoadBalancersToMAPI(fldPath, m.machine, m.awsCluster)
+	if err != nil {
+		errors = append(errors, err)
+	}
```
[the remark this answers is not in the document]
nrb: @RadekManak I don't think this is something we need to actually address, since we should be receiving what CAPI sets, which should never be `nil`, right? If somehow we get a nil controlplaneloadbalancer, something else is wrong in the cluster that conversion shouldn't fix.

nrb:
/retest-required

Tests appear to be stuck waiting for status for more than 24 hours.

pkg/controllers/infracluster/aws.go
nrb: Makes sense that if we have only 1 it's internal.

Since that's what OpenShift does and we're not really changing meanings from upstream fields, we're good to leave them as-is in this PR. The internal consistency in OpenShift is more important.

Also think Christian's comment should be incorporated.

damdo:
/test unit

Known flakes that we are talking in other PRs

RadekManak:
/test remaining-required
I think this should have been applied automatically now that all required first stage tests passed. Maybe the removed security job is still considered in this PR?

pkg/controllers/machinesync/aws.go
```diff
@@ -64,77 +62,54 @@ func (r *MachineSyncReconciler) ensureAWSLoadBalancerMatch(ctx context.Context,
...
 	}
 
-	errs := ensureExpectedLoadBalancer(lbfieldPath, &providerSpec, loadBalancersCopy, awsCluster.Spec.ControlPlaneLoadBalancer)
-
 	if awsCluster.Spec.SecondaryControlPlaneLoadBalancer != nil {
-		errs = append(errs, ensureExpectedLoadBalancer(lbfieldPath, &providerSpec, loadBalancersCopy, awsCluster.Spec.SecondaryControlPlaneLoadBalancer)...)
-	}
-
-	errs = append(errs, ensureNoRemainingLoadBalancers(lbfieldPath, &providerSpec, loadBalancersCopy)...)
+		if awsCluster.Spec.SecondaryControlPlaneLoadBalancer.Name == nil || *awsCluster.Spec.SecondaryControlPlaneLoadBalancer.Name == "" {
```
chrischdi, line 70:
```suggestion
		if ptr.Deref(awsCluster.Spec.SecondaryControlPlaneLoadBalancer.Name, "") == "" {
```

pkg/controllers/machinesync/aws.go
```diff
@@ -64,77 +62,54 @@ func (r *MachineSyncReconciler) ensureAWSLoadBalancerMatch(ctx context.Context,
...
+			return field.ErrorList{field.Invalid(lbfieldPath, providerSpec.LoadBalancers, "secondary control plane load balancer name is not configured on AWSCluster")}.
```
chrischdi, line 71: Should we also check the primary LB so the name is not empty?

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -554,6 +563,16 @@ func (r *MachineSyncReconciler) reconcileMAPIMachinetoCAPIMachine(ctx context.Co
 		controllers.ReasonResourceSynchronized, messageSuccessfullySynchronizedMAPItoCAPI, &mapiMachine.Generation)
 }
 
+// ensurePlatformMAPIToCAPIValidations verifies that shared CAPI resources are compatible before converting from MAPI -> CAPI.
+func (r *MachineSyncReconciler) ensurePlatformMAPIToCAPIValidations(ctx context.Context, mapiMachine *mapiv1beta1.Machine) error {
```
damdo, line 567: How about
```suggestion
func (r *MachineSyncReconciler) validateMAPIToCAPIPlatformSpecifics(ctx context.Context, mapiMachine *mapiv1beta1.Machine) error {
```

pkg/conversion/capi2mapi/aws.go
```diff
@@ -571,3 +579,56 @@ func handleAWSIdentityRef(fldPath *field.Path, identityRef *awsv1.AWSIdentityRef
...
+	loadBalancers = append(loadBalancers, internalLoadBalancerRef)
+
+	if awsCluster.Spec.SecondaryControlPlaneLoadBalancer != nil {
+		externalLoadBalancerRef, err := convertAWSLoadBalancerToMAPI(awsCluster.Spec.SecondaryControlPlaneLoadBalancer)
+		if err != nil {
+			return nil, field.Invalid(fldPath.Child("secondaryControlPlaneLoadBalancer"), awsCluster.Spec.SecondaryControlPlaneLoadBalancer, err.Error())
+		}
+
+		loadBalancers = append(loadBalancers, externalLoadBalancerRef)
+	}
```
damdo, line 606: We could append these errors

pkg/controllers/machinesync/aws_test.go
```diff
@@ -291,3 +292,126 @@ var _ = Describe("AWS load balancer validation during MAPI->CAPI conversion", fu
...
+		expectError           bool
+		expectedErrorMessages []string
+	}
+
+	DescribeTable("validate load balancer matching",
+		func(in validateLoadBalancerMatchTableInput) {
+			err := validateLoadBalancerReferencesAgainstExpected(in.actualLoadBalancers, in.expectedLoadBalancers, lbfieldPath)
+
+			if in.expectError {
+				Expect(err).ToNot(BeNil())
```
chrischdi, line 317:
```suggestion
			if len(in.expectedErrorMessages) {
```

So we can get rid of the `expectError` parameter.

damdo (commented):
Thanks @RadekManak mostly LGTM 🎉 
I picked up a couple of minor things, aside from the Degraded one which I believe is quite important

chrischdi (commented):
Some last nits, really happy how this shapes out :-)

huali9:
/verified later @huali9 
I still need time to thoroughly test this feature so I'd like to do post merge testing to unblock some merges.

damdo:
Fix for Openstack CAPI hasn't merged yet #381 so overriding this for now

PR #343, opened 2025-08-11
OCPCLOUD-2566: e2e tests prevent changes to non-authoritative Machines tests

miyadav:
@huali9 @theobarberbany PTAL when time permits.

Works well in local test- 
```
`miyadav@miyadav-thinkpadx1carbongen8:~/go/src/github.com/openshift/cluster-capi-operator$ go run ./hack/../vendor/github.com/onsi/ginkgo/v2/ginkgo -r -v --randomize-all --randomize-suites --keep-going --race --trace --timeout=10m --focus="with spec.authoritativeAPI: ClusterAPI, Prevent changes to non-authoritative Machines except from sync controller" ./e2e
2025/08/11 13:56:35 maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined
Running Suite: Cluster API Suite - /home/miyadav/go/src/github.com/openshift/cluster-capi-operator/e2e
======================================================================================================
Random Seed: 1754900795 - will randomize all specs

Will run 15 of 47 specs
------------------------------
[BeforeSuite] 
/home/miyadav/go/src/github.com/openshift/cluster-capi-operator/e2e/e2e_test.go:59
[BeforeSuite] PASSED [1.962 seconds]
------------------------------
SSS
P [PENDING]
[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] MachineSet Migration Tests Create MAPI MachineSets with spec.authoritativeAPI: MAPI and existing CAPI MachineSet with same name should reject creation of MAPI MachineSet with same name as existing CAPI MachineSet
/home/miyadav/go/src/github.com/openshift/cluster-capi-operator/e2e/machineset_migration_test.go:75
------------------------------
SSSSSS
P [PENDING]
[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] MachineSet Migration Tests Create MAPI MachineSets with spec.authoritativeAPI: CAPI and existing CAPI MachineSet with same name should verify that the non-authoritative MAPI MachineSet providerSpec has been updated to reflect the authoritative CAPI MachineSet mirror values
/home/miyadav/go/src/github.com/openshift/cluster-capi-operator/e2e/machineset_migration_test.go:159
------------------------------
SSSSSSSSSSSS
P [PENDING]
[...]

PR #355, opened 2025-09-08
OCPCLOUD-2566: VAP e2e tests

miyadav:
@huali9 @damdo @theobarberbany , thanks for already reviewing this earlier ( #343) , it became too long , hence made a new one , changes are same.


theobarberbany:
@miyadav  can we also make this a draft until we're ready to test :)

theobarberbany:
> when i was trying this was getting weird errors related to baremetal imports , I am not sure why , will review these again and see what could have been missed.

Ah I see - if you share those errors, maybe we can help? 

Also I think given this now is a 5000+ line change, we should try to have a separate PR for getting go workspaces configured for e2es, to avoid making the repository history overly complex :)

miyadav:
made changes to makefile now , `GOWORK=off` for earlier [failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/355/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1965031319729606656)
@theobarberbany PTAL when time permits , if these are safe changes to makefile

e2e/machine_migration_test.go
```diff
@@ -237,6 +256,179 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] Ma
...
+		Context("with spec.authoritativeAPI modification", func() {
+			It("should allow modification of authoritativeAPI from ClusterAPI to MachineAPI", func() {
+				By("Attempting to modify authoritativeAPI from ClusterAPI to MachineAPI")
+
+				// Create a new machine for this test to avoid affecting other tests
+				testMachineName := "machine-auth-change-83955"
+				testMachine := createMAPIMachineWithAuthority(ctx, cl, testMachineName, machinev1beta1.MachineAuthorityClusterAPI)
+
+				_, err := capiframework.GetMachine(cl, testMachineName, capiframework.CAPINamespace)
+				Expect(err).ToNot(HaveOccurred(), "CAPI Machine should be created")
```
theobarberbany, line 270: Style nit here (but non blocking, so mostly for future reference) 



```suggestion
				Expect(capiframework.GetMachine(cl, testMachineName, capiframework.CAPINamespace)).To(Succeed(), "CAPI Machine should be created")

```

e2e/machine_migration_test.go
```diff
@@ -237,6 +256,179 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] Ma
...
+					updatedMachine := currentMachine.DeepCopy()
+					updatedMachine.Spec.AuthoritativeAPI = machinev1beta1.MachineAuthorityMachineAPI
+
+					err = cl.Update(ctx, updatedMachine)
+					return err
+				}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(), "Should allow modification of authoritativeAPI")
+
+				// Verify the machine still exists after the change (don't wait for specific status)
+				_, err = mapiframework.GetMachine(cl, testMachineName)
+				Expect(err).ToNot(HaveOccurred(), "Machine should still exist after authoritativeAPI change")
```
theobarberbany, line 288: Can we use Komega here? 
```suggestion
				Eventually(k.Update(testMachine, func(){
				testMachine.Spec.AuthoritativeAPI = machinev1beta1.MachineAuthorityMachineAPI
				}), capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(), "Should allow modification of authoritativeAPI")
```

e2e/machine_migration_test.go
```diff
@@ -237,6 +256,179 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] Ma
...
+
+				// Clean up the test machine
+				DeferCleanup(func() {
+					By("Cleaning up test machine")
+					// Try to delete the MAPI machine first
+					if testMachine != nil {
+						mapiframework.DeleteMachines(ctx, cl, testMachine)
+					}
```
theobarberbany, line 296: Is there any reason we're not using `cleanupMachineResources(` as we do below?

e2e/machine_migration_test.go
miyadav: got it.

e2e/machine_migration_test.go
```diff
@@ -237,6 +256,179 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] Ma
...
+
+					updatedMachine := currentMachine.DeepCopy()
+					updatedMachine.Spec.AuthoritativeAPI = machinev1beta1.MachineAuthorityMachineAPI
+
+					err = cl.Update(ctx, updatedMachine)
+					return err
+				}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(), "Should allow modification of authoritativeAPI")
+
+				// Verify the machine still exists after the change (don't wait for specific status)
```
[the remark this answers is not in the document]
miyadav: it is for ginkgo retry loop in Eventually and other one is after the retries succeed, just to confirm machine update succeeds.

e2e/machine_migration_test.go
miyadav: The test was failing , for cleaning up , some conflict due to machine already in deleted status .. , to check that , just deleted mapi and it was cleaning up properly, we shall use cleanupMachineResources , let me check if that is some bug , or was just test problem.

e2e/machine_migration_test.go
miyadav: Thank you Theo using this , need to updated `machinev1beta1.MachineAuthorityMachineAPI` to `machinev1beta1.MachineAuthorityClusterAPI`

e2e/machine_migration_test.go
miyadav: In the above context only using one `testMachine`, hence cannot use the cleanresources as is , which is used below.

theobarberbany: I'm not sure I'm following here 😓 can we see some logs? My reading of `cleanupMachineResources` shows that we're basically running the same code here and in the function..

e2e/machine_migration_test.go
```diff
@@ -506,3 +683,71 @@ func cleanupMachineResources(ctx context.Context, cl client.Client, capiMachines
...
+	machine.Spec.ProviderSpec.Value.Raw = rawProviderSpec
+	return nil
+}
+func verifyUpdatePrevented(cl client.Client, machineName string, description string, updateFunc func(*machinev1beta1.Machine), expectedErrorSubstrings ...string) {
+	By(fmt.Sprintf("Attempting to %s in MAPI Machine", description))
+
+	currentMachine, err := mapiframework.GetMachine(cl, machineName)
+	Expect(err).NotTo(HaveOccurred(), "Failed to get current machine")
+	updatedMachine := currentMachine.DeepCopy()
+	updateFunc(updatedMachine)
```
theobarberbany, line 708: I think all of this could be just passing the updatefunc and machine to a `k.Update`, and we just pass the machine in question (rather than always passing machine.Name)

```go
					Eventually(k.Update(machine, updateFunc), timeout).Should(MatchError(ContainSubstring(expectedErrorSubstrings)))
```

and we move the logic for working out what `expectedErrorSubstrings` should be to above the call.

e2e/machine_migration_test.go
miyadav: I meant the cleanupresources taking args as capi and mapi machines) - , where as in here , we just check that we can modify authoritativeAPI field and then immediately clean up the mapi machine that is updated , so not needed to explicitly try to delete capimachinename as well , which is why simply deleting the updated mapi machine only.
`						cleanupMachineResources(
							ctx,
							cl,
							[]*clusterv1.Machine{testCapiMachine},
							[]*machinev1beta1.Machine{testMapiMachine},
						)`

e2e/machine_migration_test.go
miyadav: Let me add this .

PR #337, opened 2025-07-25
OCPCLOUD-2992: add migration e2e test for scaling/updating/deleting machineset

sunzhaohua2:
## Summary by CodeRabbit

- Tests
  - Expanded AWS-focused end-to-end coverage for Machine and MachineSet migration, including create/delete flows, scaling, authority switching, and synchronization/paused condition validation.
  - Added shared helpers and initialization to streamline test setup; validated parity between resources managed by different APIs.
  - Removed legacy and redundant migration suites; gated execution by platform and feature flag.

- Refactor
  - Centralized common test initialization and standardized condition constants.
  - Migrated test types to newer API versions and strengthened assertions; simplified helper signatures and error handling.

- Chores
  - Cleaned up obsolete AWS-specific test scaffolding.

PR #358, opened 2025-09-09
NO-JIRA: use go workspaces

miyadav:
@theobarberbany @damdo PTAL when time permits - 
cc @sunzhaohua2 @huali9 
Will let the regression runs and review results .


chrischdi (commented):
This PR should also add the `go.work` and `go.work.sum`. Also the tooling should then run `go work vendor` instead of `go mod vendor`.

miyadav:
hey @chrischdi , thanks , I have them [here](https://github.com/miyadav/cluster-capi-operator/blob/goworkusagefore2evendoring/go.work) , let me check if they are ignored , you are correct about `go work sync` , but then I don't get the tests working :( .

miyadav:
> Ah sorry for that, looks like the github ui fails for this PR 😂

sorry from me too :D , let me make two separate commits.

damdo:
I did try adding workspaces for this repo in the past and it didn't work very well.
Maybe things have changed now, but I remember the code editors going crazy.

chrischdi:
@miyadav : with the following `go.work`, we would have a working `go work sync` which also includes `manifests-gen` (which is then runnable via `go run ./manifests-gen`.

```
go 1.24.0

use (
        .
        ./e2e
        ./manifests-gen
)

replace (
        github.com/metal3-io/baremetal-operator => github.com/metal3-io/baremetal-operator v0.5.1
        sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v1.10.4
)
```

We should also not need `GOWORK=off` then and get the full benefits of actually using go.work. Also then: use `go work vendor` instead of `go mod vendor`.

sunzhaohua2:
I also tried go work in this [pr](https://github.com/openshift/cluster-capi-operator/pull/337), seems works well.

hack/vendor.sh
```diff
@@ -6,14 +6,20 @@ vendor() {
...
+echo "Syncing Go workspace"
+if ! go work sync; then
+  echo "Warning: go work sync failed due to dependency conflicts. This is expected with the current vsphere provider dependency."
+  echo "The workspace structure is in place for future use, but individual module vendoring will be used for builds."
+fi
 
-echo "Updating dependencies for manifests-gen"
-cd manifests-gen/ && vendor && cd -
+vendor_in_subdir "hack/tools"
+vendor_in_subdir "manifests-gen"
```
chrischdi, line 25: Shouldn't we be okay with:
* running go mod tidy for every go.mod
* running go mod verify for every go.mod
* running go work vendor

Instead of having three separate vendor directories?

theobarberbany:
If we can we have one shared vendor directory, that would be much nicer to work with IMO :)

chrischdi:
Last nit, otherwise lgtm:

* In `hack/vendor.sh` : should we add `set -e` to make it fail on errors?

Makefile
```diff
@@ -77,6 +77,10 @@ lint:
 vendor:
 	./hack/vendor.sh
 
+# Alias for vendor (used by verify-deps)
+.PHONY: deps
+deps: vendor
+
```
damdo, line 83: I can't find where this is used

I checked [here](https://github.com/openshift/release/blob/917a988aced215a7ab5eda3709067c0ba3de3dd6/ci-operator/config/openshift/cluster-capi-operator/openshift-cluster-capi-operator-main.yaml#L267-L272) on the `verify-deps` job config for ccapio, but that [doesn't seem to be using this make target](https://github.com/openshift/release/blob/917a988aced215a7ab5eda3709067c0ba3de3dd6/ci-operator/step-registry/go/verify-deps/go-verify-deps-commands.sh)

hack/vendor.sh
damdo: Did we come to an agreement on this one in the end @miyadav @chrischdi ?

damdo (commented):
It be nice to see this merged, so I just have some questions, we could still try and merge this an follow up on a separate PR depending on the answer :)

miyadav:
> @miyadav I am still seeing the `vendor` directory under `./e2e` whist I don't see it under `./hack/tools` and `./manifests-gen`. Is that expected? TY

Don't think we should them, after the ./hack/vendor.sh script is adjusted. I am testing it in local by removing it and checking if somewhere we are running `go mod vendor `in testing scripts for `e2e/go.mod` because of which it is present here.

Makefile
damdo: @theobarberbany I was not implying we should change the `verify-deps` one :)

I was just following the above comment and not undestanding what it meant
> `# Alias for vendor (used by verify-deps)`

If we it that alias is not used by anything and is just an alias I suggest we drop it

damdo (commented):
/lgtm

For retriggering the e2e

damdo:
/approve

Thanks a lot for setting this up @miyadav !

damdo:
/verified bypass

Units/CI passing should be ok, this is a tooling change

PR #360, opened 2025-09-12
OCPCLOUD-2714,OCPCLOUD-2807: e2e test for vap

miyadav:
This PR adds end-to-end tests to validate that the ValidatingAdmissionPolicy (VAP) correctly prevents users from setting CAPI-specific fields on CAPI Machines that are not supported by MAPI during the migration process.
Since the PR is not yet merged these test will fail for now , will pass as the PR merges.

cc  @sunzhaohua2 @huali9 

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end tests to ensure updates to unsupported Machine fields are blocked by admission policies.
  - Improved assertions to verify clear, user-facing error messages reference the preventing policy and indicate forbidden/denied requests.
  - Ensures cluster safeguards prevent unintended Machine modifications without impacting existing flows.


miyadav:
this might need updates after  #355 is merged

/hold

PR #364, opened 2025-09-19
OCPCLOUD-2992: refactor: split machine migration testing into multiple files

sunzhaohua2:
The original pr https://github.com/openshift/cluster-capi-operator/pull/337 is huge, it will be divided into several PRs, this PR is refactor machine migration.

## Summary by CodeRabbit

- Tests
  - Added new AWS end-to-end suites for machine migration covering CAPI- and MAPI-authoritative workflows: creation, synchronization, pause state, mirroring, and deletion scenarios.
  - Added shared migration condition constants and helpers for creating, validating, and cleaning up machines.
  - Removed legacy migration test files and clarified platform skip messaging.

- Refactor
  - Threaded context through test helpers and framework calls for more reliable API interactions.
  - Replaced many asynchronous waits with direct assertions for faster, more deterministic tests.


sunzhaohua2:
@theobarberbany can you help to take a look this? thanks!

e2e/framework/machineset.go
```diff
@@ -88,7 +88,7 @@ func CreateMachineSet(cl client.Client, params machineSetParams) *clusterv1.Mach
 		ms.Spec.Template.Spec.FailureDomain = &params.failureDomain
 	}
 
-	Expect(cl.Create(ctx, ms)).To(Succeed())
+	Expect(cl.Create(ctx, ms)).To(Succeed(), "Should have successfully created the MachineSet")
```
theobarberbany, line 91: can be more specific, 
```suggestion
	Expect(cl.Create(ctx, ms)).To(Succeed(), "Should have successfully created the CAPI MachineSet")
```

e2e/migration_common.go
```diff
@@ -0,0 +1,20 @@
+package e2e
+
+import (
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+const (
+	// SynchronizedCondition indicates that a machine resource has been successfully synchronized between MAPI and CAPI during migration..
+	SynchronizedCondition machinev1beta1.ConditionType = "Synchronized"
```
theobarberbany, line 10: Do we not export these from the MAPI libraries? Can we import them like we do with the CAPI one below? 


If not, please ignore!

e2e/migration_common.go
```diff
@@ -0,0 +1,20 @@
...
+	// MAPIPausedCondition represents the paused state for MAPI machines.
+	MAPIPausedCondition machinev1beta1.ConditionType = "Paused"
+	// CAPIPausedCondition represents the paused state for CAPI machines.
+	CAPIPausedCondition = clusterv1.PausedV1Beta2Condition
+
+	// CAPIGV represents the CAPI Machine GroupVersion.
+	CAPIGV = "cluster.x-k8s.io/v1beta1"
```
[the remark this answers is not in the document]
damdo: This comes from one of the suggestions I made, but I was wondering the same. If we were to pass down a client.Object in that func rather than a machinename string, we should be able to use `GetResourceVersion()` and detect that without these.

e2e/machine_migration_helpers.go
```diff
@@ -0,0 +1,291 @@
...
+	By(fmt.Sprintf("Creating a new MAPI machine with AuthoritativeAPI: %s in namespace: %s", authority, newMachine.Namespace))
+	Eventually(func() error {
+		return cl.Create(ctx, newMachine)
+	}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(), "Should have successfully created MAPI machine %s with AuthoritativeAPI: %s", newMachine.Name, authority)
+
+	return newMachine
+}
+
+func verifyMachineRunning(cl client.Client, machineName string, apiVersion string) {
+	// Parse the full Group/Version string to extract the API group.
```
theobarberbany, line 119:
```go
Expect(machineName).NotTo(BeEmpty(), "Machine name cannot be empty")
    Expect(apiVersion).NotTo(BeEmpty(), "API version cannot be empty")
```

e2e/machine_migration_capi_authoritative_test.go
```diff
@@ -0,0 +1,219 @@
...
+							[]*mapiv1beta1.Machine{newMapiMachine},
+						)
+					})
+				})
+				It("should delete MAPI Machine", func() {
+					mapiframework.DeleteMachines(ctx, cl, newMapiMachine)
+					mapiframework.WaitForMachinesDeleted(cl, newMapiMachine)
+				})
+
+				It("should verify the CAPI machine is deleted", func() {
```
theobarberbany, line 127: I think we're missing a case here: 

We should be able to delete a non authoritative MAPI machine without it deleting the CAPI machine if and only if there is no owner reference on the MAPI machine. 

I don't think we test this at the moment, so we need to add this case :)

e2e/machine_migration_helpers.go
```diff
@@ -0,0 +1,291 @@
...
+		Fail(fmt.Sprintf("unknown authoritativeAPI type: %v", authority))
+	}
+
+	Eventually(komega.Object(mapiMachine), capiframework.WaitMedium, capiframework.RetryMedium).Should(
+		HaveField("Status.Conditions", ContainElement(conditionMatcher)),
+		fmt.Sprintf("Should have found the expected Paused condition for MAPI Machine %s with authority: %s", mapiMachine.Name, authority),
+	)
+}
+
+func verifyCAPIMachinePausedCondition(capiMachine *clusterv1.Machine, authority mapiv1beta1.MachineAuthority) {
```
theobarberbany, line 216: 🎉I like this :D

theobarberbany:
A few more comments, but I think we're nearly there :)

e2e/migration_common.go
sunzhaohua2: yes, we cant export these like CAPI one.

e2e/machine_migration_capi_authoritative_test.go
sunzhaohua2: @theobarberbany this will be covered in machineset deletion, and there is a [bug](https://issues.redhat.com/browse/OCPBUGS-56897?focusedId=27631723&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27631723) for this, so we can't test this for now.

- Removing non-authoritative MAPI MachineSet
  - Create a MAPI MachineSet with specAPI MachineAPI and 1 replica
    - Check Machine is created and mirrored
  - Switch specAPI to ClusterAPI
    - Wait for migration to happen
  - Scale CAPI MachineSet to 2 replicas
    - Check new CAPI Machine is running
    - Check for paused non-authoritative MAPI mirror
  - Delete MAPI MachineSet -> this step **mapi machines will be deleted**. 
    - Check CAPI MachineSet not removed
    - Check both Machines and MIrrors remain
    - Neither MAPI machine should have owner reference
  - Delete MAPI Machine with specAPI ClusterAPI
    - Check CAPI mirror is not removed
  - Delete MAPI Machine with specAPI MachineAPI
    - Check CAPI mirror is removed
    - CAPI MachineSet should replace the CAPI instance
    - New CAPI Machine should not be mirrored to MAPI

e2e/machineset_migration_test.go
```diff
@@ -114,10 +108,7 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] Ma
 			})
 
 			It("should find that MAPI MachineSet has a CAPI MachineSet mirror", func() {
-				Eventually(func() error {
-					capiMachineSet, err = capiframework.GetMachineSet(cl, mapiMSAuthMAPIName, capiframework.CAPINamespace)
-					return err
-				}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(), "CAPI MachineSet mirror should exist")
+				capiMachineSet = capiframework.GetMachineSet(cl, mapiMSAuthMAPIName, capiframework.CAPINamespace)
```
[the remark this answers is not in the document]
sunzhaohua2: it is used in line 115 verifyCAPIPausedCondition(capiMachineSet, machinev1beta1.MachineAuthorityMachineAPI) :)

e2e/machine_migration_helpers.go
sunzhaohua2: I updated based on my understanding, please help to review if it make sense.

e2e/migration_common.go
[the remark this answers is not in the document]
sunzhaohua2: I used `switch m := machine.(type)` is this ok?

e2e/machine_migration_capi_authoritative_test.go
theobarberbany: Ah ok - I'd forgotten about that bug 🤦 

This looks good then!

e2e/powervs_test.go
```diff
@@ -68,7 +69,7 @@ var _ = Describe("Cluster API IBMPowerVS MachineSet", Ordered, func() {
 
 })
 
-func getPowerVSMAPIProviderSpec(cl client.Client) *mapiv1.PowerVSMachineProviderConfig {
+func getPowerVSMAPIProviderSpec(ctx context.Context, cl client.Client) *mapiv1.PowerVSMachineProviderConfig {
 	machineSetList := &mapiv1beta1.MachineSetList{}
 	Expect(cl.List(ctx, machineSetList, client.InNamespace(framework.MAPINamespace))).To(Succeed())
```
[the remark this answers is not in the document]
sunzhaohua2: there is a global ctx variable in e2e_test.go, it can't be removed because it is used in the BeforeEach/AfterEach/It test blocks.

e2e/machineset_migration_test.go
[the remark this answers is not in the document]
sunzhaohua2: Got it, I will update this in the following machineset pr

e2e/powervs_test.go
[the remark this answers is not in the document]
theobarberbany: Ah gotcha - we might want to rename it e.g `e2eCtx` so we don't shadow a common function parameter with a global variable.


(Explanation via gemini for learning purposes)

That's a **very good suggestion** and hits on a core principle of idiomatic Go programming, especially regarding the `context` package.

Here's an exploration and explanation of why renaming the global variable is a smart move in the context of Go and E2E testing:

### 1\. The Variable Shadowing Risk

  * **Go Convention:** In Go, the standard and overwhelming convention for a `context.Context` parameter passed into a function is the short, three-letter name: `ctx`.
    ```go
    func doSomething(ctx context.Context, arg string) error {
        // ... use local ctx here
    }
    ```
  * **The Problem:** If you have a global variable named `ctx`, and then you define a function that takes a parameter also named `ctx`, the function's parameter **shadows** the global variable within the function's scope. Any code inside that function will refer to the local parameter `ctx`, not the global one.
    ```go
    var ctx context.Context = context.Background() // Global variable

    func doSomething(ctx context.Context) { // Local parameter shadows the global
        // Inside this function, ctx refers to the parameter,
        // which might be different from the global variable.
        // It's ambiguous and error-prone.
    }
    ```
  * **The Consequence:** This ambiguity is a source of confusion and subtle bugs. A developer might *intend* to use the global `ctx` (e.g., if it holds application-wide values or a common timeout), but inadvertently use the local `ctx` parameter instead, or vice-versa. Renaming the global variable eliminates this ambiguity entirely.

### 2\. Clarity, Scope, and Intent

  * **Global Variables are Special:** Global variables, especially in Go, are generally discouraged unless they represent truly global, application-wide, or configuration state (like a logger, database pool, or base context). When you do use a global, you should give it a descriptive name that clearly communicates its special, broad scope.
  * **The Purpose of `e2eCtx`:** By renaming the global to `e2eCtx`, you immediately communicate two things to any developer reading the code:
    1.  This is a `context.Context`.
    2.  This specific context is tied to the **End-to-End (E2E)** testing scope.
  * **The Purpose of `ctx` (local):** The local function parameter `ctx` remains the idiomatic, short name, and it is understood to be the context relevant *for that specific function call*.

### 3\. The `context.Context` Idiom in Go

The Go idiom dictates that `context.Context` should be the **first parameter** of a function and named `ctx`. It's a fundamental part of the Go ecosystem, used for cancellation, deadlines, and passing request-scoped values.

By adhering to this standard for local parameters and using a more specific name for your global (like `e2eCtx`), you ensure your E2E test code is clear, idiomatic, and less likely to introduce bugs related to accidental variable shadowing.

**In short: Renaming the global `ctx` to `e2eCtx` is a best-practice move that improves code clarity and avoids a common class of Go programming errors.**

damdo:
/verified by @damdo 

E2Es are passing here: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/364/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/1972657659668598784

damdo:
/override ci/prow/e2e-openstack-ovn-techpreview

It's broken

PR #371, opened 2025-09-29
Updates InfraMachine watch_filters

theobarberbany:


theobarberbany:
@coderabbitai full review

pkg/controllers/machinesync/suite_test.go
```diff
@@ -64,7 +64,8 @@ func TestAPIs(t *testing.T) {
 var _ = BeforeSuite(func() {
 	klog.SetOutput(GinkgoWriter)
 
-	logf.SetLogger(textlogger.NewLogger(textlogger.NewConfig()))
+	logf.SetLogger(GinkgoLogr)
+	ctrl.SetLogger(GinkgoLogr)
```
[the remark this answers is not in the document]
theobarberbany: this just isn't true - the tests compile fine, and run locally

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -429,18 +462,39 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
 				})).Should(Succeed())
+
+				capiMachine = capiMachineBuilder.WithName(mapiMachine.Name).Build()
 			})
 
 			It("should successfully create the CAPI machine", func() {
-				Eventually(k.Get(
-					clusterv1resourcebuilder.Machine().WithName(mapiMachine.Name).WithNamespace(capiNamespace.Name).Build(),
-				)).Should(Succeed())
+				Eventually(k.Get(capiMachine)).Should(Succeed())
```
JoelSpeed, line 470: With this change, you are updating the `capiMachine` object, does that matter? With the previous, it was a separate instance so it didn't overwrite the outer context?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -429,18 +462,39 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+				capaMachine = capaMachineBuilder.WithOwnerReferences([]metav1.OwnerReference{
+					{
+						Kind:               machineKind,
+						APIVersion:         clusterv1.GroupVersion.String(),
+						Name:               capiMachine.Name,
+						UID:                capiMachine.UID,
+						BlockOwnerDeletion: ptr.To(true),
+						Controller:         ptr.To(false),
+					},
+				}).WithName(mapiMachine.Name).Build()
```
JoelSpeed, line 486: I don't understand the purpose of this, it's like you're trying to write the owner ref, but then we fetch the object next and that will overwrite everything in `capaMachine` so this has no effect?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -556,7 +621,7 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
 					})
 
 					It("should update the synchronized condition on the MAPI machine to True", func() {
-						Eventually(k.Object(mapiMachine), timeout).Should(
+						Eventually(k.Object(mapiMachine), 1000*timeout).Should(
```
JoelSpeed, line 624: Erm?

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -546,7 +611,7 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
 
 				Context("when MAPI machine with the same name and status.authoritativeAPI set to ClusterAPI is created", func() {
 					BeforeEach(func() {
-						mapiMachine = mapiMachineBuilder.WithName(capiMachine.Name).WithAuthoritativeAPI(mapiv1beta1.MachineAuthorityClusterAPI).Build()
+						mapiMachine = mapiMachineBuilder.WithGenerateName("").WithName(capiMachine.Name).WithAuthoritativeAPI(mapiv1beta1.MachineAuthorityClusterAPI).Build()
```
JoelSpeed, line 614: Why explicitly cancelling the generate name? Does an object created with both a name and generate name work?

pkg/util/watch_filters.go
```diff
@@ -37,10 +39,10 @@ const machineSetKind = "MachineSet"
...
 		klog.V(4).Info(
-			"reconcile triggered by object",
-			"objectType", fmt.Sprintf("%T", obj),
-			"namespace", obj.GetNamespace(),
-			"name", obj.GetName(),
+			"reconcile triggered by ",
+			"objectType: ", fmt.Sprintf("%T", obj),
+			"namespace: ", obj.GetNamespace(),
+			"name: ", obj.GetName(),
 		)
```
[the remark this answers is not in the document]
theobarberbany: Yep this is a good idea, will do once I'm back from PTO :)

pkg/controllers/machinesync/machine_sync_controller.go
```diff
@@ -282,11 +288,17 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
...
 	infraCluster, infraMachine, err := r.fetchCAPIInfraResources(ctx, sourceCAPIMachine)
-	if err != nil {
+	if err != nil { //nolint: nestif
 		fetchErr := fmt.Errorf("failed to fetch Cluster API infra resources: %w", err)
 
 		if existingMAPIMachine == nil {
 			r.Recorder.Event(sourceCAPIMachine, corev1.EventTypeWarning, "SynchronizationWarning", fetchErr.Error())
+			// Don't requeue for terminal configuration errors
+			if errors.Is(err, errInvalidInfraClusterReference) || errors.Is(err, errInvalidInfraMachineReference) {
+				logger.Info("Detected terminal configuration error - cluster name or infrastructure machine reference is empty, not requeuing", "error", err.Error())
```
chrischdi: Intentionally not using `logger.Error`?

chrischdi: (also for the others)

theobarberbany: good catch, will update!

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -429,18 +462,32 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+
+				capiMachine = capiMachineBuilder.WithName(mapiMachine.Name).Build()
 			})
 
 			It("should successfully create the CAPI machine", func() {
-				Eventually(k.Get(
-					clusterv1resourcebuilder.Machine().WithName(mapiMachine.Name).WithNamespace(capiNamespace.Name).Build(),
-				)).Should(Succeed())
+				Eventually(k.Get(capiMachine)).Should(Succeed())
 			})
```
mdbooth: I don't think this is testing anything. We already asserted that Create() succeeded in BeforeEach. Are we wondering if it was deleted since? I would just delete this test.

pkg/util/watch_filters.go
```diff
@@ -79,6 +81,38 @@ func ResolveCAPIMachineSetFromInfraMachineTemplate(namespace string) func(contex
...
+			if err != nil {
+				logger.Error(err, "Failed to parse GroupVersion", "APIVersion", ref.APIVersion)
+				continue
+			}
+
+			if ref.Kind == "Machine" && gv.Group == clusterv1.GroupVersion.Group {
+				requests = append(requests, reconcile.Request{
+					NamespacedName: client.ObjectKey{Namespace: namespace, Name: ref.Name},
+				})
+			}
```
mdbooth: Do we expect multiple Machine owner references? You should probably just return at this point.

This also means you don't need to append, you can return a slice you instantiate here, initialised with a single element.

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -235,11 +236,6 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
 	})
 
 	Context("when all the CAPI infra resources exist", func() {
-		BeforeEach(func() {
-			By("Creating the CAPI infra machine")
-			Eventually(k8sClient.Create(ctx, capaMachine)).Should(Succeed(), "capa machine should be able to be created")
-		})
-
```
damdo: Are we sure we want to remove this? It was the precondition in this context `"when all the CAPI infra resources exist"`

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -272,11 +268,28 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+							Kind:               machineKind,
+							APIVersion:         clusterv1.GroupVersion.String(),
+							Name:               capiMachine.Name,
+							UID:                capiMachine.UID,
+							BlockOwnerDeletion: ptr.To(true),
+							Controller:         ptr.To(false),
+						},
+					}).Build()
+
+					Expect(k8sClient.Create(ctx, capaMachine)).To(Succeed(), "capa machine should be able to be created")
```
damdo: I am not sure about this move. I get here we need the MAPI owner refs to be set. But couldn't we keep the original `			By("Creating the CAPI infra machine")` in the outer Context and here only do an OwnerRef update for the capaMachine?

I worry this move would render the Context structure false

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -514,21 +561,32 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+					// Note we don't create it on the API server yet.
+					noMachineSetCapaMachineBuilder := capaMachineBuilder.WithName("test-machine-no-machineset")
+					capaMachine = noMachineSetCapaMachineBuilder.Build()
+
+					capiMachine = capiMachineBuilder.WithName("test-machine-no-machineset").WithInfrastructureRef(corev1.ObjectReference{
+						Kind:      capaMachine.Kind,
+						Name:      capaMachine.GetName(),
+						Namespace: capaMachine.GetNamespace(),
+					}).Build()
+					Expect(k8sClient.Create(ctx, capiMachine)).Should(Succeed())
```
damdo: I'd rather use Eventuallys instead of Expect all over the place for these k8s api calls

pkg/util/watch_filters.go
```diff
@@ -79,6 +81,38 @@ func ResolveCAPIMachineSetFromInfraMachineTemplate(namespace string) func(contex
...
+// and uses owner references to determine the owning CAPI machine. If one is found, it returns a reconcile.Request
+// for the corresponding MAPI Machine in the MAPI namespace to trigger reconciliation of the mirror MAPI Machine.
+func ResolveCAPIMachineFromInfraMachine(namespace string) func(context.Context, client.Object) []reconcile.Request {
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		logger := logf.FromContext(ctx).WithValues(
+			"objectType", fmt.Sprintf("%T", obj),
+			"namespace", obj.GetNamespace(),
+			"name", obj.GetName(),
+		)
+		logger.V(4).Info("reconcile triggered")
```
damdo: Wouldn't we normally log this in the controller instead of here?

pkg/util/watch_filters.go
```diff
@@ -79,6 +81,38 @@ func ResolveCAPIMachineSetFromInfraMachineTemplate(namespace string) func(contex
...
+		requests := []reconcile.Request{}
+
+		for _, ref := range obj.GetOwnerReferences() {
+			gv, err := schema.ParseGroupVersion(ref.APIVersion)
+			if err != nil {
+				logger.Error(err, "Failed to parse GroupVersion", "APIVersion", ref.APIVersion)
+				continue
+			}
+
+			if ref.Kind == "Machine" && gv.Group == clusterv1.GroupVersion.Group {
```
damdo: Is there an equivalent kind for `clusterv1.GroupVersionKind.Kind` for Machine?

pkg/util/watch_filters.go
theobarberbany: yea, this is just a log line to output the values we set above

pkg/controllers/machinesync/suite_test.go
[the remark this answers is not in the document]
damdo: Let's not forget about this :)

sunzhaohua2:
@huali9 PTAL :)

PR #376, opened 2025-10-03
OCPCLOUD-2998: implement synchronizedAPI

RadekManak:
## Summary
- Gate migration transitions of `status.AuthoritativeAPI` toward `spec.AuthoritativeAPI` on `status.SynchronizedAPI` and `status.SynchronizedGeneration`, so authority only advances from a stable synchronized source state.
- Keep the transition through `Migrating` explicit and reset synchronized status after the final `status.AuthoritativeAPI` handoff so sync can re-establish state from the new authoritative API.
- Align machine and MachineSet migration behavior around the same state machine by moving the shared transition logic into `pkg/controllers/migrationcommon`.
- Add shared unit/integration coverage for the new migration flow and outage e2e coverage showing MachineSet migration can still advance with Cluster API controllers down, and that rollback remains possible while the target stays paused.

## Why
Before this PR, the migration controllers inferred the stable source of truth from `spec.AuthoritativeAPI` and `status.SynchronizedGeneration` alone.

That breaks down when a migration request changes mid-transition. Once `status.AuthoritativeAPI` is `Migrating`, `status.SynchronizedGeneration` does not say whether it reflects Machine API or Cluster API state, so the controller can no longer safely tell which side is stable enough to advance toward `spec.AuthoritativeAPI`.

In practice, that can block or mis-handle reversals of in-progress migration requests, especially when the target Cluster API side is unavailable. This PR makes the migration state explicit by tracking `status.SynchronizedAPI`, gating authority changes on the actual synchronized source state, and resetting sync status after the final handoff.

Implements [OCPCLOUD-2998](https://issues.redhat.com/browse/OCPCLOUD-2998).

## Reviewer Guide
This PR is large, but the behavior change is concentrated in a few places. It will likely be easier to review by area rather than as one continuous diff.

1. Start with `pkg/controllers/synccommon/syncstatus.go` and `pkg/controllers/synccommon/migratestatus.go`.
   These define the status contract used by migration: `status.SynchronizedAPI`, `status.SynchronizedGeneration`, and how synchronized status is reset after authority changes.

2. Then read `pkg/controllers/migrationcommon/controller.go`.
   This is the shared migration state machine that moves `status.AuthoritativeAPI` toward `spec.AuthoritativeAPI`, including the explicit `Migrating` transition.

3. Then review the thin controller adapters:
   - `pkg/controllers/machinemigration/...`
   - `pkg/controllers/machinesetmigration/...`
   These mostly wire Machines and MachineSets into the shared logic.

4. Review the remaining diff with that model in mind.
   Most of the PR size after that is test restructuring and expanded coverage around the shared state machine and outage scenarios.

## Test Plan
- Unit tests for `migrationcommon` and `synccommon`
- Updated machine and MachineSet migration controller tests
- Updated sync controller tests
- MachineSet outage e2e coverage for migration/rollback behavior with Cluster API controllers down

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added explicit synchronization state tracking for machines and MachineSets during migration between Machine API and Cluster API.
  * Enabled migration rollback support when service disruptions occur.

* **Tests**
  * Added comprehensive end-to-end coverage for migration behavior under outage scenarios.


RadekManak:
go: github.com/openshift/api@v0.0.0-20260105114749-aae5635a71a7 (replaced by ../api): reading ../api/go.mod: open /go/src/github.com/openshift/api/go.mod: no such file or directory

Yes, I switched to using a local project folders for development instead of pushing into my fork. I'll correct this when the API PR is merged.

mdbooth (commented):
We just discussed this. Unfortunately I don't think this approach is correct.

The purpose of the `SynchronizedAPI` field is to eliminate the heuristic in `isSynchronized` which would fail if we were reverting an incomplete migration. Specifically, the `SynchronizedAPI` field defines deterministically exactly which object the value of `SynchronizedGeneration` refers to. Therefore `SynchronizedAPI` should be written by the sync controller whenever we write `SynchronizedGeneration`.

Additionally, we should ensure that each field is owned by a particular controller and is never touched by any other controller. Therefore the migration controller should not write to this value even to initialise it. It should just wait for the sync controller to initialise it if it is required.

With these changes in place, the additional code path in the migration controller guarded by `IsMigrationCancellationRequested` is not required. As long as we can unambiguously determine the synchronisation state in all circumstances, the existing logic is already correct. Therefore we should just need to update the logic in `isSynchronized` to use the `SynchronizedAPI` field and therefore no longer require the invalid heuristic.

RadekManak:
/testwith openshift/cluster-capi-operator/main/e2e-aws-capi-techpreview https://github.com/openshift/machine-api-operator/pull/1442

RadekManak:
@coderabbitai full review

e2e/machineset_migration_disruptive_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	BeforeAll(func() {
+		if platform != configv1.AWSPlatformType {
+			Skip(fmt.Sprintf("Skipping tests on %s, this is only supported on AWS", platform))
+		}
+
+		if !capiframework.IsFeatureGateEnabled(ctx, cl, features.FeatureGateMachineAPIMigration) {
+			Skip("Skipping, this feature is only supported on MachineAPIMigration enabled clusters")
+		}
+
```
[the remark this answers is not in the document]
RadekManak: This is N/A. This controller is only deployed on standalone openshift clusters.
Where it this guideline defined?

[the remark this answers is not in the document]
RadekManak: I see.
https://www.youtube.com/watch?v=eWDrfcxWsuM

e2e/machine_migration_helpers.go
```diff
@@ -478,3 +478,12 @@ func verifyMachineSynchronizedGeneration(mapiMachine *mapiv1beta1.Machine, autho
 		Fail(fmt.Sprintf("unknown authoritativeAPI type: %v", authority))
 	}
 }
+
+// verifyMachineSynchronizedAPI verifies that the MAPI Machine's status.synchronizedAPI matches the expected value.
+func verifyMachineSynchronizedAPI(mapiMachine *mapiv1beta1.Machine, expectedSynchronizedAPI mapiv1beta1.SynchronizedAPI) {
+	By(fmt.Sprintf("Verifying MAPI Machine SynchronizedAPI is %s", expectedSynchronizedAPI))
```
stefanonardo, line 486:
```suggestion
func verifyMachineSynchronizedAPI(mapiMachine *mapiv1beta1.Machine, expectedSynchronizedAPI mapiv1beta1.SynchronizedAPI) {
    GinkgoHelper()

	By(fmt.Sprintf("Verifying MAPI Machine SynchronizedAPI is %s", expectedSynchronizedAPI))
```

e2e/machineset_migration_helpers.go
```diff
@@ -289,6 +484,15 @@ func verifyMAPIMachineSetSynchronizedCondition(mapiMachineSet *mapiv1beta1.Machi
 	}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed())
 }
 
+// verifyMachineSetSynchronizedAPI verifies that the MAPI MachineSet's status.synchronizedAPI matches the expected value.
+func verifyMachineSetSynchronizedAPI(mapiMachineSet *mapiv1beta1.MachineSet, expectedSynchronizedAPI mapiv1beta1.SynchronizedAPI) {
+	By(fmt.Sprintf("Verifying MAPI MachineSet SynchronizedAPI is %s", expectedSynchronizedAPI))
```
stefanonardo, line 491:
```suggestion
func verifyMachineSetSynchronizedAPI(mapiMachineSet *mapiv1beta1.MachineSet, expectedSynchronizedAPI mapiv1beta1.SynchronizedAPI) {
    GinkgoHelper()

	By(fmt.Sprintf("Verifying MAPI MachineSet SynchronizedAPI is %s", expectedSynchronizedAPI))
```

pkg/controllers/synccommon/syncstatus.go
```diff
@@ -124,22 +135,124 @@ func newSyncStatusApplyConfiguration[
...
+
+	return ""
+}
+
+// MigrationDirection determines the current and desired authorities for a migration.
+// When statusAuthority is Migrating, it uses SynchronizedAPI to infer the current authority.
+// On success, currentAuthority is guaranteed to be either MachineAPI or ClusterAPI.
+// Missing or invalid SynchronizedAPI values are returned as errors instead of
+// surfacing as a Migrating or otherwise unsupported current authority.
+func MigrationDirection(statusAuthority mapiv1beta1.MachineAuthority, synchronizedAPI mapiv1beta1.SynchronizedAPI, specAuthority mapiv1beta1.MachineAuthority) (mapiv1beta1.MachineAuthority, mapiv1beta1.MachineAuthority, bool, error) {
```
stefanonardo: nit for making it more readable for callers:

```suggestion
func MigrationDirection(statusAuthority mapiv1beta1.MachineAuthority, synchronizedAPI mapiv1beta1.SynchronizedAPI, specAuthority mapiv1beta1.MachineAuthority) (currentAuthority, desiredAuthority mapiv1beta1.MachineAuthority, isMigrating bool, err error) {
```

RadekManak:
@damdo Can we get this merged?

PR #455, opened 2026-02-05
SPLAT-2410:  Introduce Cluster API for OpenShift Installations on vSphere

AnnaZivkovic:
Enables vSphere support in the Machine API to Cluster API conversion framework. It implements bidirectional conversion between vSphere Machine API resources (Machines/MachineSets) and Cluster API resources (Machines/MachineTemplates), including support for vSphere-specific v1beta2 conditions.  

Changes                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
  - New conversion libraries:                                                                                                                                                                                                     
    - pkg/conversion/mapi2capi/vsphere.go - Converts Machine API to Cluster API                                                                                                                                 
    - pkg/conversion/capi2mapi/vsphere.go - Converts Cluster API to Machine API                                                                                                                     
  - Controller integration:                                                                                                                                                                                                       
    - Registered vSphere handlers in machinesync controller                                                                                                                                                                       
    - Registered vSphere handlers in machinesetsync controller                                                                                                                                                                    
    - Enabled vSphere provider in machine-api-migration binary                                                                                                                                                                    
  - Test coverage:                                                                                                                                                                                                                
    - Comprehensive unit tests for both conversion directions                                                                                                                                                    
    - Fuzz testing for conversion stability                                                                                                                                                                                                                                                                                                                                                                           
  - Added v1beta2 condition handling (pkg/util/conditions.go):                                                                                                                                                                    
    - GetV1Beta2Condition() - Retrieves conditions from status.v1beta2.conditions                                                                                                                                                 
    - GetV1Beta2ConditionStatus() - Gets condition status from v1beta2 location                                                                                                                                                   
    - GetConditionStatusFromInfraObject() - Intelligently checks both v1beta2 (vSphere) and v1beta1 (AWS/Azure/GCP) condition locations                                                                                           
    - EnsureCAPIV1Beta2Conditions() - Merges v1beta2 conditions during conversion                                                                                                                                                 
    - Backward compatibility: Maintains support for existing providers using v1beta1 conditions while adding v1beta2 support for vSphere                                                                                            
                                                                                                                                                         

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Adds VMware vSphere platform support across migration, machine and machineset sync, and conversion flows; includes providerSpec parsing and round‑trip conversions.
  * Startup now registers vSphere types and propagates feature‑gate handling to conditionally enable vSphere controllers.

* **Bug Fixes**
  * Uses infra‑aware condition/status retrieval in migration flows to correctly detect paused state.

* **Tests**
  * Adds extensive unit and fuzz tests for vSphere conversion and round‑trip scenarios.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,485 @@
...
+	if workspace.Server != "" || workspace.Datacenter != "" || workspace.Folder != "" ||
+		workspace.Datastore != "" || workspace.ResourcePool != "" {
+		mapiProviderConfig.Workspace = workspace
+	}
+
+	// Below this line are fields not converted from the CAPI VSphereMachine.
+
+	// ProviderID - Populated at a different level.
+	// FailureDomain - Handled via zone label in buildAdditionalMetadata.
+	// PowerOffMode - VSphereMachine lifecycle field, not part of the provider spec.
```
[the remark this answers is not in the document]
AnnaZivkovic: That was vague. These fields control runtime/operational behavior rather than the machine's infrastructure provisioning configuration. I updated the descriptions.

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,485 @@
...
+
+	mapiMachine, err := fromCAPIMachineToMAPIMachine(m.machine, additionalMachineAPIMetadataLabels, additionalMachineAPIMetadataAnnotations)
+
+	if err != nil {
+		errs = append(errs, err...)
+	}
+
+	mapiMachine.Spec.ProviderSpec.Value = vsphereSpecRawExt
+	// Note: ProviderStatus is not set during conversion, similar to OpenStack and PowerVS providers.
+	// The MAPI controller will manage the status at runtime.
```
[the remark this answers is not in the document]
JoelSpeed: I don't know why those other providers are allowing stale status. That is not the expectation of the overall architecture. I would expect this to be kept updated.

Please check how AWS is handling this

admission-policies/vsphere/unsupported-vsphere-spec-fields.yaml
```diff
@@ -0,0 +1,179 @@
...
+      expression: "object.kind == 'VSphereMachine' ? object.spec : object.spec.template.spec"
+    - name: specPath
+      expression: "object.kind == 'VSphereMachine' ? 'spec' : 'spec.template.spec'"
+  validations:
+    # Cluster-level configuration
+    - expression: "!has(variables.machineSpec.thumbprint) || variables.machineSpec.thumbprint == ''"
+      messageExpression: "variables.specPath + '.thumbprint is not supported - TLS validation is handled at cluster level'"
+
+    # CAPI-only runtime configuration
+    - expression: "!has(variables.machineSpec.customVMXKeys) || variables.machineSpec.customVMXKeys.size() == 0"
```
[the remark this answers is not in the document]
vr4manta: hmm, so having MAPI conversion add these controller controlled to CAPI would make sense so if you switch from MAPI to CAPI, the state of the VM will be consistent.  Adding additional parameters is what we may want to prevent then in this case.  We can do that property checking in the capi2mapi if i recall correctly.  Else we may want to look into making MAPI support exposing the ability to add these configs.

cmd/machine-api-migration/main.go
```diff
@@ -147,13 +149,27 @@ func checkFeatureGates(ctx context.Context, mgr ctrl.Manager) {
...
+func checkPlatformSupported(ctx context.Context, platform configv1.PlatformType, featureGateAccessor featuregates.FeatureGateAccess) {
 	switch platform {
 	case configv1.AWSPlatformType, configv1.OpenStackPlatformType:
 		klog.Infof("MachineAPIMigration: starting %s controllers", platform)
+	case configv1.VSpherePlatformType:
+		currentFeatureGates, err := featureGateAccessor.CurrentFeatureGates()
+		if err != nil {
+			klog.Error(err, "unable to get current feature gates")
+			os.Exit(1)
+		}
```
damdo: I'd refactor this a bit. 

- move out from `checkFeatureGates` the `getFeatureGates` into the main func and rename it to `getFeatureGatesAccessor()`. Make it return `featureGateAccessor`.
- Then also in the outer main right after have `currentFeatureGates, err := featureGateAccessor.CurrentFeatureGates()`
- rename `func checkFeatureGates` to ` func checkMachineAPIMigrationEnabled`
- in `func checkPlatformSupported` take in as argument `currentFeatureGates` rather than `featureGateAccessor`, this way when we add featuregates for the other platforms we don't need to refactor this again.

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+func FromVSphereMachineSetAndInfra(m *mapiv1beta1.MachineSet, i *configv1.Infrastructure) MachineSet {
+	return &vSphereMachineSetAndInfra{
+		machineSet:     m,
+		infrastructure: i,
+		vSphereMachineAndInfra: &vSphereMachineAndInfra{
+			machine: &mapiv1beta1.Machine{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels:      m.Spec.Template.ObjectMeta.Labels,      //nolint:staticcheck // ObjectMeta is a field, not embedded
+					Annotations: m.Spec.Template.ObjectMeta.Annotations, //nolint:staticcheck // ObjectMeta is a field, not embedded
+				},
```
damdo: We don't do this in other providers implementations. What's the reason for this?

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+		return nil, nil, nil, field.ErrorList{field.Invalid(field.NewPath("spec", "providerSpec", "value"), v.machine.Spec.ProviderSpec.Value, err.Error())}
+	}
+
+	capvMachine, warn, machineErrs := v.toVSphereMachine(vSphereProviderConfig)
+	if machineErrs != nil {
+		errs = append(errs, machineErrs...)
+	}
+
+	warnings := make([]string, 0, len(warn))
+	warnings = append(warnings, warn...)
```
damdo: I think we might not need this as there is only one warnings slice and it is coming out of `v.toVSphereMachine`

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+
+	// Run the full ToMachine conversion to check for errors
+	capiMachine, capvMachineObj, warns, machineErrs := v.toMachineAndInfrastructureMachine()
+	if machineErrs != nil {
+		errors = append(errors, machineErrs.ToAggregate().Errors()...)
+	}
+
+	warnings := make([]string, 0, len(warns))
+	warnings = append(warnings, warns...)
+
```
damdo: Same as above

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+// toVSphereMachine converts a MAPI VSphereMachineProviderConfig to a CAPI VSphereMachine.
+//
+//nolint:funlen
+func (v *vSphereMachineAndInfra) toVSphereMachine(providerSpec mapiv1beta1.VSphereMachineProviderSpec) (*vspherev1.VSphereMachine, []string, field.ErrorList) {
+	var errs field.ErrorList
+
+	fldPath := field.NewPath("spec", "providerSpec", "value")
+
+	// Convert network configuration
+	capiNetworkSpec, networkWarnings, networkErrs := convertMAPINetworkSpecToCAPI(fldPath.Child("network"), providerSpec.Network)
```
damdo: This is Vsphere specific. Better to make it reflect in the name.
`convertVsphere...`

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+	if len(networkErrs) > 0 {
+		errs = append(errs, networkErrs...)
+	}
+
+	warnings := make([]string, 0, len(networkWarnings))
+	warnings = append(warnings, networkWarnings...)
+
+	// Convert data disks
+	capiDataDisks, diskErrs := convertMAPIDataDisksToCAPI(fldPath.Child("dataDisks"), providerSpec.DataDisks)
```
damdo: Same as https://github.com/openshift/cluster-capi-operator/pull/455/changes#r3092640602

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+	if len(diskErrs) > 0 {
+		errs = append(errs, diskErrs...)
+	}
+
+	// Convert clone mode
+	capiCloneMode := convertMAPICloneModeToCAPI(providerSpec.CloneMode)
```
damdo: Same as https://github.com/openshift/cluster-capi-operator/pull/455/changes#r3092640602

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+func convertMAPICloneModeToCAPI(mapiMode mapiv1beta1.CloneMode) vspherev1.CloneMode {
+	switch mapiMode {
+	case mapiv1beta1.FullClone:
+		return vspherev1.FullClone
+	case mapiv1beta1.LinkedClone:
+		return vspherev1.LinkedClone
+	default:
+		return vspherev1.FullClone // Default to FullClone
+	}
+}
```
damdo: I see here we are defaulting to FullClone.
Doesn't the `mapiv1beta1.CloneMode` default match the `vspherev1.CloneMode` default?
If so, couldn't we leave it empty if it is empty?
In other places we deal with emptiness and defaulting explicitly. I think we should do that here too

```
		case "":
			// Default - no setting
		default:
			errs = append(errs, field.Invalid(fldPath.Index(i).Child("YYYYYY"), ... )
```

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+	capiDataDisks, diskErrs := convertMAPIDataDisksToCAPI(fldPath.Child("dataDisks"), providerSpec.DataDisks)
+	if len(diskErrs) > 0 {
+		errs = append(errs, diskErrs...)
+	}
+
+	// Convert clone mode
+	capiCloneMode := convertMAPICloneModeToCAPI(providerSpec.CloneMode)
+
+	spec := vspherev1.VSphereMachineSpec{
+		PowerOffMode: vspherev1.VirtualMachinePowerOpModeHard,
```
damdo: Is this always fixed?

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	// Set workspace fields if available
+	if providerSpec.Workspace != nil {
+		spec.Server = providerSpec.Workspace.Server
+		spec.Datacenter = providerSpec.Workspace.Datacenter
+		spec.Folder = providerSpec.Workspace.Folder
+		spec.Datastore = providerSpec.Workspace.Datastore
+		spec.ResourcePool = providerSpec.Workspace.ResourcePool
+		// VMGroup - MAPI-specific field for vm-host group based zoning, not supported in CAPV.
```
damdo, line 314: Is this a potential candidate to be added in upstream CAPV?
If not we should probably decide if it is worth erroring the conversion or throwing a warning. Somehow we need to let the user know about this.

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+			MemoryMiB:         providerSpec.MemoryMiB,
+			DiskGiB:           providerSpec.DiskGiB,
+			TagIDs:            providerSpec.TagIDs,
+			Network:           capiNetworkSpec,
+			DataDisks:         capiDataDisks,
+			// Server - Set below from Workspace.
+			// Datacenter - Set below from Workspace.
+			// Folder - Set below from Workspace.
+			// Datastore - Set below from Workspace.
+			// ResourcePool - Set below from Workspace.
```
damdo: So we say we set these in the Workspace. But if the workspace is not set on the MAPI side, do we still need to fill these in?

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+	if !reflect.DeepEqual(providerSpec.ObjectMeta, metav1.ObjectMeta{}) {
+		// We don't support setting the object metadata in the provider spec.
+		// It's only present for the purpose of the raw extension and doesn't have any functionality.
+		errs = append(errs, field.Invalid(fldPath.Child("metadata"), providerSpec.ObjectMeta, "metadata is not supported"))
+	}
+
+	// Only take action when a non-default credentials secret is being used in MAPI.
+	// If the user is using the default, then their CAPV secret will already be configured and no action is necessary.
+	if providerSpec.CredentialsSecret != nil &&
+		providerSpec.CredentialsSecret.Name != "" &&
```
damdo: I don't see this emptiness check in the AWS provider conversion for example. Is there a reason why this differs? Are we ok when there is no CredentialsSecret defined?

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+	return &vspherev1.VSphereMachine{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: vspherev1.GroupVersion.String(),
+			Kind:       vSphereMachineKind,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      v.machine.Name,
+			Namespace: capiNamespace,
+		},
+		Spec: spec,
```
damdo: No `Status: status,` conversion in this function?
Like: https://github.com/openshift/cluster-capi-operator/blob/317bfa4476b42232bd9979ef3676a2bf56c54c91/pkg/conversion/mapi2capi/aws.go#L331-L334

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+
+	// Only take action when a non-default credentials secret is being used in MAPI.
+	// If the user is using the default, then their CAPV secret will already be configured and no action is necessary.
+	if providerSpec.CredentialsSecret != nil &&
+		providerSpec.CredentialsSecret.Name != "" &&
+		providerSpec.CredentialsSecret.Name != DefaultVSphereCredentialsSecretName {
+		// Not convertable; need custom credential configuration
+		errs = append(errs, field.Invalid(fldPath.Child("credentialsSecret"), providerSpec.CredentialsSecret.Name, fmt.Sprintf("credentialsSecret does not match the default of %q, credentials must be configured at the cluster level", DefaultVSphereCredentialsSecretName)))
+	}
+
```
damdo, line 341: Have we evaluated if we need/want a similar check to https://github.com/openshift/cluster-capi-operator/blob/317bfa4476b42232bd9979ef3676a2bf56c54c91/pkg/conversion/mapi2capi/aws.go#L299-L305

pkg/conversion/mapi2capi/vsphere.go
```diff
@@ -0,0 +1,463 @@
...
+			Nameservers: device.Nameservers,
+		}
+
+		// Convert AddressesFromPools
+		if len(device.AddressesFromPools) > 0 {
+			addressesFromPools := make([]corev1.TypedLocalObjectReference, len(device.AddressesFromPools))
+			for j, pool := range device.AddressesFromPools {
+				addressesFromPools[j] = corev1.TypedLocalObjectReference{
+					APIGroup: &pool.Group,
+					Kind:     pool.Resource, // This might need adjustment based on actual mapping
```
damdo: Do we need to take any extra action based on this comment?

pkg/util/conditions.go
```diff
@@ -98,6 +132,50 @@ func GetConditionStatus(obj client.Object, conditionType string) (corev1.Conditi
...
+	}
+
+	// Both returned Unknown. If the errors are expected (missing fields), return Unknown with no error
+	// to avoid reconcile failures for providers that simply omit the condition.
+	// Otherwise, return the real error.
+	if err != nil && !errors.Is(err, errStatusConditionsNotInterface) {
+		return corev1.ConditionUnknown, err
+	}
+
+	return corev1.ConditionUnknown, nil
```
damdo: I need to think about this further. I want to make sure this is a robust approach given this is going to change for many providers.
i.e. in CAPA for example, we'll shortly have v1beta3 conditions on the status for the infra objects. I think we need to make this more flexible

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,552 @@
...
+		errs = append(errs, err...)
+	}
+
+	vsphereSpecRawExt, errRaw := RawExtensionFromInterface(mapaSpec)
+	if errRaw != nil {
+		return nil, nil, fmt.Errorf("unable to convert vSphere providerSpec to raw extension: %w", errRaw)
+	}
+
+	warnings := make([]string, 0, len(warn))
+	warnings = append(warnings, warn...)
```
damdo: Same here

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,552 @@
...
+
+	additionalMachineAPIMetadataLabels, additionalMachineAPIMetadataAnnotations := m.buildAdditionalMetadata()
+
+	mapiMachine, err := fromCAPIMachineToMAPIMachine(m.machine, additionalMachineAPIMetadataLabels, additionalMachineAPIMetadataAnnotations)
+
```
damdo, line 123: nit unnecessary empty line

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,552 @@
...
+	// Run the full ToMachine conversion so that we can check for
+	// any Machine level conversion errs in the spec translation.
+	mapiMachine, warn, err := m.ToMachine()
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	warnings := make([]string, 0, len(warn))
+	warnings = append(warnings, warn...)
```
damdo: same

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,552 @@
...
+	// GuestSoftPowerOffTimeout - Timeout for soft power operations, not part of machine provisioning spec.
+	// NamingStrategy - Controls VM naming behavior, not part of machine provisioning spec.
+	// Thumbprint - Ignore - Not required, TLS validation handled by cluster-level configuration.
+	// StoragePolicyName - Ignore - Not supported in MAPI VSphereMachineProviderSpec.
+	// Resources - Ignore - CPU/memory reservations, limits, and shares not supported in MAPI.
+	// AdditionalDisksGiB - Deprecated in CAPV, converted to DataDisks above.
+	// CustomVMXKeys - Ignore - Not supported in MAPI VSphereMachineProviderSpec.
+	// PciDevices - Ignore - Not supported in MAPI VSphereMachineProviderSpec.
+	// OS - Ignore - Not supported in MAPI VSphereMachineProviderSpec.
+	// HardwareVersion - Ignore - Not supported in MAPI VSphereMachineProviderSpec.
```
damdo: Any field that we are expected to set in MAPV (required), for which CAPV doesn't have a value and we need to make one up?

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,552 @@
...
+		errs = append(errs, field.Invalid(fldPath.Child("guestSoftPowerOffTimeout"), spec.GuestSoftPowerOffTimeout, "guestSoftPowerOffTimeout is not supported"))
+	}
+
+	if spec.NamingStrategy != nil {
+		// Not required for our use case.
+		errs = append(errs, field.Invalid(fldPath.Child("namingStrategy"), spec.NamingStrategy, "namingStrategy is not supported"))
+	}
+
+	return errs
+}
```
damdo, line 367: As this logic stands, if any of these is being set on CAPV, the conversion of that machine/machineset on MAPV will fail.

Is that always ok? Or are there certain cases where warning is also ok instead of erroring?

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,552 @@
...
+	// We check if addresses are set to determine if the VM has been provisioned
+	if len(m.vSphereMachine.Status.Addresses) == 0 {
+		return ""
+	}
+
+	if m.vSphereMachine.Status.Ready {
+		return "ready"
+	}
+
+	return "not-ready"
```
damdo: Do constants exist for these in MAPI/MAPV?

pkg/conversion/capi2mapi/vsphere.go
```diff
@@ -0,0 +1,552 @@
...
+	}
+
+	return []metav1.Condition{{
+		Type:    string(mapiv1beta1.MachineCreation),
+		Status:  metav1.ConditionFalse,
+		Reason:  mapiv1beta1.MachineCreationFailedConditionReason,
+		Message: message,
+		// LastTransitionTime will be set by the condition utilities.
+	}}
+}
```
damdo, line 470: This is again something we'll need to think about with regards to v1beta1 vs v1betaX

damdo (commented):
Thanks @AnnaZivkovic great work here!

I left some thoughts and some minor nits.
The two main themes we need to reason get right here are:
- ensure we have status conversion everywhere as well as spec
- understand what to do with v1beta1 vs v1betaX status conversion

pkg/conversion/mapi2capi/vsphere.go
AnnaZivkovic: The reason I enabled vSphere to copy Labels and Annotations from the MachineSet template (while AWS/OpenStack don't) was for roundtrip conversion consistency.   Tho, I can see not being necessary, but if we don't populate the template labels/annotations couldn't they be lost during conversion?

pkg/conversion/mapi2capi/vsphere.go
AnnaZivkovic: The defaults differ between MAPI and CAPV:                                                                                                                                                                          
                                                                                                                                                                                                                                  
  - **MAPI**: [Defaults to FullClone](https://github.com/openshift/api/blob/f9587f6e7c6067507c9b2436096ebd550ae93d7c/machine/v1beta1/types_vsphereprovider.go#L69)                                                                                            
  - **CAPV**: [Defaults to LinkedClone](https://github.com/openshift/cluster-api-provider-vsphere/blob/73c2570169977bf94949f180210facdd8253154d/apis/v1beta1/types.go#L108)

  
  From what I understand we can't leave CloneMode empty because the defaults are applied at runtime by the controllers, not via admission webhooks. MAPI's controller defaults to FullClone while CAPV's controller defaults to LinkedClone, so converting empty ->empty would change the VM provisioning behavior (FullClone->LinkedClone). We need to explicitly set FullClone to preserve the runtime behavior.

pkg/conversion/mapi2capi/vsphere.go
AnnaZivkovic: Since:
  1. MAPI doesn't have a PowerOffMode field (it's CAPV-specific)
  2. CAPV defaults to hard when omitted                                                                                                                                                                                           
  3. We validate that it must be hard or empty when converting back to MAPI
  
We can remove the hardcoded value and let CAPV apply its own default. This is cleaner because we're not being overly opinionated about CAPV-specific fields. I will make this change

pkg/conversion/mapi2capi/vsphere.go
AnnaZivkovic: Good catch! The comments were misleading. We shouldn't silently accept empty Workspace because that would cause CAPV to fail provisioning with no clear error. I think it is out of scope to support the CPMS with failuredomains at the moment. I created a story to track the work here https://redhat.atlassian.net/browse/SPLAT-2725

pkg/conversion/mapi2capi/vsphere.go
AnnaZivkovic: I was treating an empty secrete the same as nil and setting it as the default secrete, but for consistency Ill make it match AWS

pkg/conversion/mapi2capi/vsphere.go
AnnaZivkovic: This falls into later support for full functionality.

pkg/conversion/capi2mapi/vsphere.go
AnnaZivkovic: I dont think there are any required MAPI fields that we need to "make up" values for. All required MAPI fields (Template and Network) have direct equivalents in CAPV that are also required. The only fields we set defaults for are optional fields

pkg/conversion/mapi2capi/vsphere.go
AnnaZivkovic: I dont think we need it as unlike aws have exactly one region and all machines must be in that region.   Vsphere is a lot more flexible. It can support multiple vCenters and different failure domains can use different vCenters/datacenters all of which makes the logic more complicated to check.  If misconfigured a vSphere vm will fail with a clear error from vCenter anyway. I can add some validation if you think it would be beneficial.

pkg/conversion/capi2mapi/vsphere.go
AnnaZivkovic: Not that I know of. I added them as contents.

pkg/conversion/capi2mapi/vsphere.go
AnnaZivkovic: @damdo Do you have any ideas of how to approach this? I am not sure how to come up with a more generic solution in the scope of this pr.  My naive solution would be to have all of the platforms upgrade.

pkg/conversion/capi2mapi/vsphere.go
AnnaZivkovic: I moved some of them to warnings.

pkg/util/conditions.go
damdo: @AnnaZivkovic let's use this approach: https://gist.github.com/damdo/9e4803e0d7a03b1c26a4ca30d0439063

pkg/conversion/consts/consts.go
```diff
@@ -27,4 +27,7 @@ const (
 
 	// MAPIMachineMetadataAnnotationInstanceState is the annotation for the instance state of the machine is used as column for kubectl.
 	MAPIMachineMetadataAnnotationInstanceState = "machine.openshift.io/instance-state"
+
+	// VSphereInstanceStateReady indicates the vSphere instance is ready.
+	VSphereInstanceStateReady = "ready"
```
damdo: Is this only used in MAPI or also in CAPI?

pkg/conversion/mapi2capi/vsphere.go
damdo: cc. @AnnaZivkovic

pkg/conversion/mapi2capi/vsphere.go
damdo: Let's see if conversion works without and drop it for now. We can add it back later if there is a need for it.

damdo (commented):
Thank you for the patience here @AnnaZivkovic 

I left some follow-up comments and provided a more dynamic solution for conditions conversion.

damdo:
@AnnaZivkovic units are failing (fuzzy) PTAL

pkg/conversion/consts/consts.go
AnnaZivkovic: It's used in both conversion directions:
- MAPI->CAPI (mapi2capi/vsphere.go:526): Checks if MAPI's ProviderStatus.InstanceState equals "ready" to set CAPV's boolean Ready field
- CAPI->MAPI (capi2mapi/vsphere.go:408): When CAPV's Ready is true, sets MAPI's InstanceState to "ready"

It's the translation bridge between MAPI's string-based InstanceState and CAPV's boolean Ready. The constant lives in consts because both conversion directions need it.

damdo (commented):
Thanks a lot @AnnaZivkovic the implementation looks reasonable.
There might still be some edges we need to refine but this looks good enough for me as a first pass (TPNU).

/approve

damdo:
@AnnaZivkovic would you be able to investigate the failures on https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/455/pull-ci-openshift-cluster-capi-operator-main-e2e-vsphere-capi-techpreview/2079165004816322560

Thanks!

AnnaZivkovic:
> @AnnaZivkovic would you be able to investigate the failures on https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/455/pull-ci-openshift-cluster-capi-operator-main-e2e-vsphere-capi-techpreview/2079165004816322560
> 
> Thanks!

AnnaZivkovic:
I think We should add a dedicated `e2e-vsphere-capi-devpreview` job because `MachineAPIMigrationVSphere` is currently enabled only with `DevPreviewNoUpgrade`, while the existing vSphere CAPI job uses `TechPreviewNoUpgrade`. In the failed run, the migration process detected that the gate was disabled and did notstart, so the tests failed waiting for mirrored CAPI Machines and `Synchronized=True`. This will provide  vSphere migration coverage without incorrectly promoting the feature to TechPreview and TechPreview job should skip these migration-specific tests when the gate is disabled.

mdbooth:
> I think We should add a dedicated `e2e-vsphere-capi-devpreview` job because `MachineAPIMigrationVSphere` is currently enabled only with `DevPreviewNoUpgrade`, while the existing vSphere CAPI job uses `TechPreviewNoUpgrade`. In the failed run, the migration process detected that the gate was disabled and did notstart, so the tests failed waiting for mirrored CAPI Machines and `Synchronized=True`. This will provide vSphere migration coverage without incorrectly promoting the feature to TechPreview and TechPreview job should skip these migration-specific tests when the gate is disabled.

Agree. Please can you add that and link it here? Please can the -devpreview job also run only the vsphere CAPI migration tests, not the full suite. In fact, maybe call it -migration instead?

Also, please update the tests which required the Migration gate to use the correct gate.

AnnaZivkovic:
new e2e-vsphere-capi-customnoupgrade-migration job: https://github.com/openshift/release/pull/84593

damdo:
@AnnaZivkovic we need to start being able to run these on vsphere too right?
- https://github.com/AnnaZivkovic/cluster-capi-operator/blob/bd2d0a3d5b85068851aca6c70c9d0b076e2364da/e2e/machineset_migration_mapi_authoritative_test.go#L37-L39
- https://github.com/AnnaZivkovic/cluster-capi-operator/blob/bd2d0a3d5b85068851aca6c70c9d0b076e2364da/e2e/machineset_migration_capi_authoritative_test.go#L34-L36
- https://github.com/AnnaZivkovic/cluster-capi-operator/blob/bd2d0a3d5b85068851aca6c70c9d0b076e2364da/e2e/machine_migration_capi_authoritative_test.go#L35-L37
- https://github.com/AnnaZivkovic/cluster-capi-operator/blob/bd2d0a3d5b85068851aca6c70c9d0b076e2364da/e2e/machine_migration_mapi_authoritative_test.go#L35-L37

(there might be more scattered across the e2e base, I suggest searching for:
```
		if platform != configv1.AWSPlatformType {
			Skip(fmt.Sprintf("Skipping tests on %s, this is only supported on AWS", platform))
		}
```
in `e2e/*`)

e2e/migration_common.go
```diff
@@ -39,3 +70,56 @@ const (
 func generateName(prefix string) string {
 	return prefix + utilrand.String(5)
 }
+
+// newInfraMachineObject returns a stub infrastructure machine for the current
```
damdo: Would it be possible to pass the actual provider object here instead of constructing a fresh stub? `verifyResourceRemoved()` only performs a GET and expects `NotFound`, so the check could pass even if the `VSphereMachine` was never created. Retaining the object fetched during setup would make the deletion assertion more meaningful.

e2e/machine_sync_test.go
```diff
@@ -34,15 +33,13 @@ import (
 var _ = Describe("Machine Sync", Ordered, func() {
 	BeforeAll(func() {
 		switch platform {
-		case configv1.AWSPlatformType, configv1.OpenStackPlatformType:
+		case configv1.AWSPlatformType, configv1.OpenStackPlatformType, configv1.VSpherePlatformType:
```
damdo, line 36: Could this test also validate the underlying `VSphereVM` now that it runs on vSphere? `GetObjectFromContractVersionedRef()` resolves the CAPI `InfrastructureRef` to the `VSphereMachine`, so the current UID check does not cover VM recreation or a stuck VM finalizer.

The supplied must-gather provides an example from this run. `VSphereVM/machine-mapi-roundtrip-pcwrr` was captured with `deletionTimestamp: "2026-09-15T09:02:39Z"` and the `vspherevm.infrastructure.cluster.x-k8s.io` finalizer. CAPV later logged `failed to get VSphereMachine for VSphereVM: VSphereMachine.infrastructure.cluster.x-k8s.io "machine-mapi-roundtrip-pcwrr" not found`.

An assertion covering the `VSphereVM` could help identify this type of cleanup issue in the test.

e2e/machine_migration_helpers.go
```diff
@@ -443,6 +422,81 @@ func summarizeMAPIConditions(conditions []mapiv1beta1.Condition) string {
...
+
+	By(fmt.Sprintf("Creating a new CAPI AWSMachine in namespace: %s", newAWSMachine.Namespace))
+	Eventually(func() error {
+		return cl.Create(ctx, newAWSMachine)
+	}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(), "Should have successfully created AWSMachine %s/%s", newAWSMachine.Namespace, newAWSMachine.Name)
+
+	return newAWSMachine
+}
+
+func createVSphereInfraMachine(ctx context.Context, cl client.Client, referenceName, machineName string) *vspherev1.VSphereMachine {
```
damdo, line 473: Could we retain the created `VSphereMachine` and verify the corresponding `VSphereVM` during cleanup? The current migration cleanup verifies the provider machine, but an explicit VM check could help detect a provider VM that remains behind after the machine is removed.

e2e/machineset_migration_helpers.go
```diff
@@ -87,6 +77,63 @@ func createCAPIMachineSet(ctx context.Context, cl client.Client, replicas int32,
...
+	}
+
+	Eventually(func() error {
+		return cl.Create(ctx, awsMachineTemplate)
+	}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(), "Failed to create AWSMachineTemplate %s", awsMachineTemplate.Name)
+
+	return awsMachineTemplate
+}
+
+func createVSphereInfraMachineTemplate(ctx context.Context, cl client.Client, name string) *vspherev1.VSphereMachineTemplate {
```
damdo, line 117: Could MachineSet cleanup also verify that the `VSphereVM` objects created from this template are removed? Deleting the `VSphereMachineTemplate` confirms template cleanup, but it does not necessarily confirm that the provider VMs have been removed.

damdo (commented):
Could we extend the provider-agnostic portions of `machine_migration_vap_test.go` to vSphere as well? The policy locks the MAPI `spec` generically and does not depend on AWS.

One possible structure would be:

```go
BeforeAll(func() {
    switch platform {
    case configv1.AWSPlatformType, configv1.VSpherePlatformType:
    default:
        Skip(...)
    }
    skipUnlessMigrationEnabled()
})
```

The generic contexts could then run on both platforms:

* locked `spec` fields
* protected labels and annotations
* CAPI Machine forbidden fields
* MachineAPI-authority bypass

The existing AWS provider-spec context could remain AWS-only with a nested platform guard. If provider-spec coverage is useful for vSphere too, a small analogous test could use `VSphereMachineProviderSpec`, for example by modifying `template` or `numCPUs`. It may also be useful to use a platform-appropriate provider ID in the generic bypass test instead of the current AWS-formatted value.

PR #505, opened 2026-03-25
OCPBUGS-78476: Adds nutanix to images.json

theobarberbany:
We need nutanix to be in images.json for the new installer / manifests-gen tooling to work


nrb:
Is this still needed in light of https://github.com/openshift/cluster-capi-operator/pull/527?

mdbooth:
> Is this still needed in light of #527?

No

/close

PR #519, opened 2026-04-10
OCPCLOUD-3359,OCPCLOUD-3345: Add support for TLS envsubst substitution

mdbooth:
Depends on https://github.com/openshift/api/pull/2786

- **DNM: temporary o/api o/client-go overrides**
- **make vendor**
- **Set observedRevisionGeneration when writing revisions**
- **working**
- **Bump library-go**
- **make vendor**
- **update**
- **lints**

## Summary by CodeRabbit

* **New Features**
  * TLS-derived substitutions applied to installer revisions.
  * Automatic TLS profile detection with a watcher that triggers operator restart on profile changes.
  * New metrics Services, TLS secret mounts, and RBAC bindings to expose and secure operator metrics.

* **Improvements**
  * Structured logging and more consistent, failure-propagating startup behavior.
  * Dependency updates across modules for compatibility.

* **Tests**
  * New unit and e2e TLS negotiation tests covering multiple TLS profiles and handshake behaviors.

* **Bug Fixes**
  * Adjusted health probe port in a deployment manifest.


mdbooth:
@coderabbitai review

theobarberbany:

checking if it's an outage - ignore matt :)

e2e/tls_test.go
```diff
@@ -0,0 +1,447 @@
...
+			By("Restoring original APIServer TLS configuration")
+			restoreAdherence := apiServer.Spec.TLSAdherence
+			// tlsAdherence cannot be removed once set, so if the original was
+			// empty (unset), restore to LegacyAdheringComponentsOnly which is
+			// the behavioral equivalent.
+			if restoreAdherence == configv1.TLSAdherencePolicyNoOpinion {
+				restoreAdherence = configv1.TLSAdherencePolicyLegacyAdheringComponentsOnly
+			}
+			setTLSConfig(restoreAdherence, apiServer.Spec.TLSSecurityProfile)
+		})
```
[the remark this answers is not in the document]
mdbooth: This is a reasonable point, but it's hard to verify because there's no clear API signal that the change has been effected. This is why I deliberately didn't gate these tests on pod restarts but instead on the observed TLS behaviour. Lets see if it's a problem in practice. Other tests *should* be robust to these services restarting.

e2e/framework/util.go
```diff
@@ -107,3 +107,35 @@ func IsMachineAPIMigrationEnabled(ctx context.Context) bool {
...
+
+	desiredVersion := clusterVersion.Status.Desired.Version
+	if len(desiredVersion) == 0 && len(clusterVersion.Status.History) > 0 {
+		desiredVersion = clusterVersion.Status.History[0].Version
+	}
+
+	featureGate := &configv1.FeatureGate{
+		ObjectMeta: metav1.ObjectMeta{Name: "cluster"},
+	}
+	Eventually(cl.Get(ctx, client.ObjectKey{Name: "cluster"}, featureGate)).Should(Succeed(), "featureGate should be available")
```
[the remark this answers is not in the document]
mdbooth: I can't use komega here without a major refactor to use contexts. I've fixed it without using komega.

mdbooth:
This was an unrelated flake. I'm working on a separate bugfix for it, which will go in a separate PR.

https://github.com/openshift/cluster-capi-operator/pull/525

/test unit

e2e/framework/util.go
```diff
@@ -107,3 +108,41 @@ func IsMachineAPIMigrationEnabled(ctx context.Context) bool {
...
+		}
+		for _, enabled := range fg.Enabled {
+			if string(enabled.Name) == name {
+				return true
+			}
+		}
+	}
+
+	return false
+}
```
damdo: Haven't checked, but it feels like it could be possible to refactor IsMachineAPIMigrationEnabled to use this func?

pkg/commoncmdoptions/commonoptions.go
```diff
@@ -50,93 +58,116 @@ var (
...
-	LeaderElectionConfig config.LeaderElectionConfiguration
+// OperatorConfig contains configuration options common to all CAPI operator managers.
+type OperatorConfig struct {
+	CAPINamespace     *string
+	MAPINamespace     *string
+	OperatorNamespace *string
+
+	// ClusterTLSProfileSpec is the cluster-wide TLS profile spec. Operands will
+	// use this.
+	ClusterTLSProfileSpec configv1.TLSProfileSpec
```
damdo: As discussed in our chat, let's not expose this but rather leverage ManagerTLSOptions everywhere.

pkg/commoncmdoptions/commonoptions.go
```diff
@@ -50,93 +58,116 @@ var (
...
+
+	// ManagerTLSOptions are the TLS options functions used by the manager.  It
+	// defaults to the cluster-wide TLS profile, but can be overridden on the
+	// command line. Metrics and webhooks will use these.
+	ManagerTLSOptions []func(config *tls.Config)
```
damdo: If we are going to use this for both managers and operands, do we want to simply call it `TLSOptions`?

damdo (commented):
Thanks a lot! A couple of thoughts but looks good!

damdo (commented):
/approve
/lgtm

Great! Thanks for addressing the feedback!

damdo:
@mdbooth I think something is off here given all the CAPI TP jobs failed

damdo (commented):
/lgtm

Thanks @mdbooth !!

mdbooth:
The 2 OpenStack failures against this PR look like infra flakes. I would prefer not to override if we can get them to pass.

PR #528, opened 2026-04-24
NO-JIRA: fix: tolerate not-found upon infra machine deletion

damdo:
When the machine sync controller detects immutable spec changes on a CAPI infrastructure machine,
it deletes and recreates it.

If the infra machine has no finalizers at the time of deletion (e.g the CAPA controller already removed its own and the sync finalizer was not yet added), Delete removes the object immediately and the subsequent Update to strip finalizers fails with NotFound. This set the Synchronized condition to False with reason FailedToUpdateCAPIInfraMachine permanently which blocked the synchronization.

As such we want to tolerate NotFound errors on both the Delete and Update calls in ensureCAPIInfraMachineDeleted, if the object is already gone, as that is the desired outcome.

This also adds unit tests covering the NotFound tolerance on both calls,
and an integration test that deletes the infra machine out-of-band before a providerSpec update to verify the controller recovers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reconciler now treats already-deleted infrastructure machines as non-fatal, avoiding unnecessary reconciliation failures and allowing updates to proceed.
  * Ensures spec updates (e.g., instance type changes) are correctly applied after out-of-band deletions and reconciliation resumes.

* **Tests**
  * Added end-to-end and unit tests covering deletion + spec-mutation ordering and various failure/NotFound scenarios to improve reliability.

### DIFF
--- a/pkg/controllers/machinesync/machine_sync_controller_test.go
+++ b/pkg/controllers/machinesync/machine_sync_controller_test.go
@@ -525,6 +525,58 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
 					)
 				})
 			})
+
+			Context("when the CAPI infra machine is deleted and the MAPI machine providerSpec gets updated", func() {
+				BeforeEach(func() {
+					By("Waiting for initial synchronization to complete")
+					Eventually(k.Object(mapiMachine), timeout).Should(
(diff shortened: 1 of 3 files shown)


theobarberbany (commented):
/lgtm
/approve

PR #543, opened 2026-04-30
OCPBUGS-84815: fix: scope webhook to capi namespace, remove unused webhook endpoints

damdo:
See: https://redhat.atlassian.net/browse/OCPBUGS-84815

On unsupported platforms (e.g. IBMCloud), the openshift-capi-controllers ValidatingWebhookConfiguration is deployed by CVO but the webhook server does not start, causing Cluster creation in any namespace to fail with "connection refused". This blocks MCE/ACM which install CAPI independently and create Cluster objects in their own namespace.

Scope the remaining Cluster webhook to the openshift-cluster-api namespace via namespaceSelector so it no longer intercepts requests in other namespaces.

Remove the coreprovider, infrastructureprovider, and clusterctl provider webhook entries which had no corresponding handler code registered and were for CRDs no longer deployed by this operator.

## Summary by CodeRabbit

* **Chores**
  * Removed obsolete provider validation webhooks and simplified webhook configuration.
  * Restricted Cluster API validation to the designated namespace.
  * Registered additional CRD API types in the test runtime and updated test dependencies.

* **Tests**
  * Added end-to-end suites validating webhook namespace scoping and in-namespace admission behavior.


mdbooth:
I'd really like to see an e2e for this. It could be really simple: create and delete a Cluster in the `default` namespace. It should pass on all platforms, regardless of whether the VWC is present or even operating correctly. It should continue to pass when we replace this webhook with a VAP.

I understand the temptation to do the e2e later, but we should resist it.

e2e/webhook_test.go
```diff
@@ -0,0 +1,175 @@
...
+			},
+		}
+	})
+
+	It("should allow creating a non-webhook-compliant Cluster API Cluster outside the managed namespace", func() {
+		By("Creating a Cluster API Cluster in the default namespace")
+		Eventually(func() error {
+			return cl.Create(ctx, cluster)
+		}).WithContext(ctx).Should(Succeed(),
+			"Non-Webhook-Compliant Cluster API Cluster creation should be allowed outside the managed namespace")
```
mdbooth: You should either not use context, or use the context of the current node with a additional timeout on the Eventually and a timeout on the node. As written, this will take 10 minutes(?) to timeout.

My personal preference is the latter. That would mean:
* Add a `ctx context.Context` argument to the It() function.
* Add .WithTimeout(defaultEventuallyTimeout) to the Eventually
* Add defaultNodeTimeout to the It() call

Or remove the WithContext() from Eventually. That also works.

Same comment for every Eventually here.

e2e/webhook_test.go
```diff
@@ -0,0 +1,175 @@
...
+			},
+			Spec: clusterv1.ClusterSpec{
+				InfrastructureRef: clusterv1.ContractVersionedObjectReference{
+					Kind:     "UnsupportedInfraCluster", // unsupported platform
+					Name:     "webhook-scope-test",
+					APIGroup: "infrastructure.cluster.x-k8s.io",
+				},
+			},
+		}
+	})
```
mdbooth, line 70: Nit: This is common. I'd move it outside both contexts into a function which returns a new one.

e2e/webhook_test.go
```diff
@@ -0,0 +1,175 @@
...
+//   - The infrastructureRef kind "UnsupportedInfraCluster" is not in the allowed
+//     list (AWSCluster, AzureCluster, GCPCluster, etc.).
+//   - The cluster name "webhook-scope-test" does not match the cluster's
+//     infrastructureName, which the webhook requires for objects in the
+//     managed namespace.
+//
+// If the webhook were cluster-scoped, all three operations (create, update,
+// delete) would be rejected. The test proves the namespaceSelector is working
+// by asserting they succeed.
+var _ = Describe("Cluster API Webhook Namespace Scoping", Ordered, func() {
```
mdbooth: Neither of these need to be Ordered. These should be Serial, and BeforeAll should be a BeforeEach.

RadekCap:
> I'd really like to see an e2e for this.

We already have tests for it - https://github.com/stolostron/capi-tests
These tests caught this failure.

They're doing exactly what you're asking for: https://github.com/stolostron/capi-tests/tree/main/docs/test-analysis
(and much more)

PR #544, opened 2026-04-30
NO-JIRA: manifests-gen: scope provider webhooks to capi namespace

damdo:
Add a kustomize replacement that injects namespaceSelector on all ValidatingWebhookConfiguration and MutatingWebhookConfiguration entries, restricting them to the openshift-cluster-api namespace.

This applies to all CAPI providers that consume the manifests-gen kustomize component, preventing their webhooks from intercepting requests in other namespaces where the webhook server may not be running (e.g. on unsupported platforms).

A dummy ConfigMap is used as the value source for the kustomize replacement because kustomize replacements require a concrete resource to source field values from. The ConfigMap is annotated with config.kubernetes.io/local-config: "true" so it is not emitted in the final output.

## Summary by CodeRabbit

* **Chores**
  * Added a local configuration source to define a default namespace selector for webhooks.
  * Updated manifest generation to inject namespaceSelector values into validating and mutating webhook configurations so policies can target specific namespaces.
  * Marked a retained secret as local-only so it’s excluded from final output.

### DIFF
--- a/manifests-gen/kustomization.yaml
+++ b/manifests-gen/kustomization.yaml
@@ -3,6 +3,31 @@ kind: Component
 
 namespace: openshift-cluster-api
 
+resources:
+  - webhook-namespace-selector.yaml
+
+# Scope all provider webhooks to the openshift-cluster-api namespace.
+# This prevents webhooks from intercepting requests in other namespaces,
+# which would fail on unsupported platforms where the webhook server is not running.
+replacements:
+  - source:
+      kind: ConfigMap
+      name: webhook-namespace-selector
+      fieldPath: data.namespace
+    targets:
+      - select:
+          kind: ValidatingWebhookConfiguration
+        fieldPaths:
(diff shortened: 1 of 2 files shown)


mdbooth:
I suspect this will have 'interesting' edge cases. For example, I don't think we have any today, but I wonder what this would do to to cluster-scoped resources. I think at some point we're going to want to allow providers to be more selective about how they apply these transformations.

For now, though, this solves a real problem.
/lgtm
/approve

damdo:
Added:
- https://github.com/openshift/cluster-api-provider-aws/pull/605
- https://github.com/openshift/cluster-api/pull/292

To test this approach and validate it

mdbooth:
/verified by https://github.com/openshift/cluster-api-provider-aws/pull/605 and https://github.com/openshift/cluster-api/pull/292
/hold cancel

PR #551, opened 2026-05-07
NO-JIRA: Fix races in machineset sync units

mdbooth:
- **Fix test race deleting CAPA machine template**
- **Fix incorrect uses of k8sClient.Create in Eventually**
- **Fix machineset sync test races between MAPI and CAPI object creation**

## Summary by CodeRabbit

* **Tests**
  * Improved synchronization and reliability of MachineSet reconciliation tests for ClusterAPI scenarios by adding post-status-update hooks, waiting for controller-observed generation propagation, and strengthening deletion/finalizer sequencing to mirror real cleanup behavior.
  * Standardized test resource creation by routing all creates through an option-aware helper, including consistent dry-run enforcement and “already exists” tolerance for admission policy object setup.

### DIFF
--- a/pkg/controllers/machinesetsync/machineset_sync_controller_test.go
+++ b/pkg/controllers/machinesetsync/machineset_sync_controller_test.go
@@ -42,6 +42,7 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
 	capierrors "sigs.k8s.io/cluster-api/errors"
 
+	gomegatypes "github.com/onsi/gomega/types"
 	"github.com/openshift/cluster-api-actuator-pkg/testutils"
 	consts "github.com/openshift/cluster-capi-operator/pkg/controllers"
 	"github.com/openshift/cluster-capi-operator/pkg/controllers/machinesync"
@@ -154,6 +155,19 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 		mgrCancel()
 		// Wait for the mgrDone to be closed, which will happen once the mgr has stopped
 		<-mgrDone
+	}
+
+	// eventuallyManagerInformerCache is a helper function to wait for the
+	// manager's informer cache to observe an object in a particular state. This
+	// prevents a class of race condition where the test is trying to assert
+	// behaviour of the controller in the presence of existing state. We need to
+	// ensure that the controller has had a chance to observe the state the test
+	// just created first.
+	eventuallyManagerInformerCache := func(obj client.Object) gomegatypes.AsyncAssertion {
+		return Eventually(func(g Gomega) client.Object {
+			g.Expect(mgr.GetClient().Get(ctx, client.ObjectKeyFromObject(obj), obj)).To(Succeed())
+			return obj
+		})
 	}
 
 	BeforeEach(func() {
@@ -547,23 +561,44 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 	})
 
 	Context("when the MAPI machine set has MachineAuthority set to Cluster API", func() {
+		// afterMAPIUpdate can be set in a sub-context's BeforeEach to register
+		// initialisation to happen after the MAPI machine set is updated.
+		//
+		// NOTE: This is a workaround for the deeply nested test structure
+		// which makes it difficult to control the order of setup operations.
+		// Do not copy this pattern to new tests or extend its usage. New
+		// tests should be structured so that this kind of deferred setup is
+		// not necessary.
+		var afterMAPIUpdate func()
+
 		BeforeEach(func() {
 			By("Creating the MAPI machine set")
 
 			mapiMachineSet = mapiMachineSetBuilder.Build()
 			Eventually(kCreate(ctx, mapiMachineSet)).Should(Succeed())
+		})
 
+		// This triggers reconciliation, so we do it in JustBeforeEach to allow
+		// sub-contexts to make additional changes first without racing.
+		JustBeforeEach(func() {
 			By("Setting the MAPI machine set AuthoritativeAPI to ClusterAPI")
 			Eventually(k.UpdateStatus(mapiMachineSet, func() {
 				mapiMachineSet.Status.AuthoritativeAPI = mapiv1beta1.MachineAuthorityClusterAPI
-			})).Should(Succeed())
 
-			// We need to set the observed generation to the metadata generation
-			// to ensure the status is updated as that's a prerequisite for the status to be updated by the machinesetsync controller.
-			By("Setting the MAPI machine set observed generation to its metadata generation")
-			Eventually(k.UpdateStatus(mapiMachineSet, func() {
+				// We need to set the observed generation to the metadata
+				// generation to ensure the status is updated as that's a
+				// prerequisite for the status to be updated by the
+				// machinesetsync controller.
 				mapiMachineSet.Status.ObservedGeneration = mapiMachineSet.Generation
 			})).Should(Succeed())
+
+			if afterMAPIUpdate != nil {
+				afterMAPIUpdate()
+
+				// afterMAPIUpdate is not reinitialised between tests in this
+				// context, so we must clear it after use.
+				afterMAPIUpdate = nil
+			}
 		})
 
 		Context("when the CAPI machine set exists and the spec differs (replica count)", func() {
@@ -583,6 +618,12 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 				Eventually(k.UpdateStatus(capiMachineSet, func() {
 					capiMachineSet.Status.ObservedGeneration = capiMachineSet.Generation
 				})).Should(Succeed())
+
+				By("Waiting for the manager's informer cache to observe the CAPI machine set", func() {
+					eventuallyManagerInformerCache(capiMachineSet).Should(
+						HaveField("Status.ObservedGeneration", Equal(capiMachineSet.Generation)),
+					)
+				})
 			})
 
 			It("should set the sync finalizer on both the mapi and capi machine sets", func() {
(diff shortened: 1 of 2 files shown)


pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -882,7 +883,7 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 			})
 		})
 
-		Context("when the CAPI machine set has a non-zero deletion timestamp", func() {
+		Context("when the CAPI machine set is deleted", func() {
```
theobarberbany: nit: why the rename? lol 

is non-zero deletion timestamp not more specific to what's going on here?

pkg/controllers/machinesetsync/machineset_vap_test.go
```diff
@@ -72,11 +71,11 @@ var _ = Describe("MachineSet VAP Tests", func() {
 
 		mapiNamespace = corev1resourcebuilder.Namespace().
 			WithGenerateName("openshift-machine-api-").Build()
-		Eventually(k8sClient.Create(ctx, mapiNamespace)).Should(Succeed(), "mapi namespace should be able to be created")
+		Eventually(kCreate(ctx, mapiNamespace)).Should(Succeed(), "mapi namespace should be able to be created")
```
theobarberbany, line 74: 🎉 nice on catching these

pkg/controllers/machinesetsync/machineset_vap_test.go
```diff
@@ -232,11 +224,11 @@ var _ = Describe("MachineSet VAP Tests", func() {
 			testVersion := "1"
 			capiMachineSet.Spec.Template.Spec.Version = testVersion
 
-			Eventually(k8sClient.Create(ctx, capiMachineSet), timeout).Should(MatchError(ContainSubstring(".version is a forbidden field")))
+			Eventually(kCreate(ctx, capiMachineSet, client.DryRunAll), timeout).Should(MatchError(ContainSubstring(".version is a forbidden field")))
```
theobarberbany, line 227: Noice - this explains a flake I sometimes saw! 🎉 

We use this same pattern in the machine sync VAP tests, could we change/ add it there too?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -547,23 +561,38 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 	})
 
 	Context("when the MAPI machine set has MachineAuthority set to Cluster API", func() {
+		// afterMAPIUpdate can be set in a sub-context's BeforeEach to register
+		// initialisation to happen after the MAPI machine set is updated.
+		var afterMAPIUpdate func()
```
theobarberbany, line 572: This pattern ... I get the intent but makes understanding what code is executing where even worse (layer before each, justbefore each, contexts) then trace this through? 

BeforeEach -> Subcontext BeforeEach where this actually gets set -> JustBeforeEach where this gets called 😢 

Feels like we're hand rolling a fix to our (probably my) mis-structuring of tests :/ Can we break up the higher contexts so we avoid this?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -547,23 +561,38 @@ var _ = Describe("With a running MachineSetSync controller", func() {
...
+				// generation to ensure the status is updated as that's a
+				// prerequisite for the status to be updated by the
+				// machinesetsync controller.
 				mapiMachineSet.Status.ObservedGeneration = mapiMachineSet.Generation
 			})).Should(Succeed())
+
+			if afterMAPIUpdate != nil {
+				afterMAPIUpdate()
+
+				// afterMAPIUpdate is not reinitialised between tests in this
```
theobarberbany, line 598: If we start this pattern I think someone / an agent will forget to unset it and we'll spend a day or two chasing our tails

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
```diff
@@ -156,6 +157,19 @@ var _ = Describe("With a running MachineSetSync controller", func() {
 		<-mgrDone
 	}
 
+	// eventuallyManagerInformerCache is a helper function to wait for the
+	// manager's informer cache to observe an object in a particular state. This
+	// prevents a class of race condition where the test is trying to assert
+	// behaviour of the controller in the presence of existing state. We need to
+	// ensure that the controller has had a chance to observe the state the test
+	// just created first.
+	eventuallyManagerInformerCache := func(obj client.Object) gomegatypes.AsyncAssertion {
```
theobarberbany, line 166: This is good though :D

pkg/controllers/machinesetsync/machineset_vap_test.go
mdbooth: I probably wouldn't have spotted them if they hadn't flaked in a test run, tbh. Not the first time I've made this mistake, either. It's a tricky api.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
mdbooth: I'm *very* sympathetic to what you're saying. I was thinking the same. However, as you say the issue is structural: it's the nature of these deeply-nested tests. Fixing this structurally would amount to refactoring the entire test suite. I'm sympathetic to that, too, but it's not likely to happen in the near future.

This was honestly the simplest solution I could come up with without rewriting everything. It at least has a clear name, so it's explicit. If you can think of anything simpler I'd be in favour.

One change I'd be tempted to make to this, though, is to add a comment to explain this was added under duress, and should not be copy-pasta'd or extended. New tests should be structured so this is not necessary. Thoughts?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
mdbooth: You worried that an agent might delete this line? Not following...

theobarberbany: I think you addressed this in this comment: https://github.com/openshift/cluster-capi-operator/pull/551#discussion_r3225274555

My above comment wasn't all that clear, I meant if this becomes a pattern that gets copied - but a big do not copy this comment should work 🤞

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: Yeah... fair. It would be ugly and you're right.. wont be worked any time soon.

> One change I'd be tempted to make to this, though, is to add a comment to explain this was added under duress, and should not be copy-pasta'd or extended. New tests should be structured so this is not necessary. Thoughts?

I think this would be a pragmatic middle ground. 

I wonder if we can lint / catch a generalised version of this pattern somehow 🤔

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: Any chance of a big dont copy block above this?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: @mdbooth ^

mdbooth: I vaguely recall changing this because I find the logic hard to read, even in controllers. And I'm *extremely* familiar with it. 'is deleted' is the intent and you don't need to think about negations.

It's not important though. Do you really want it back?

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
mdbooth: Looks like I added this.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
mdbooth: I'm still not following. Can you be explicit about what the comment should contain?

mdbooth:
@theobarberbany can we merge this? I don't think there's anything important outstanding.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: Thats not what any of the logic in the controller gates on though right? It's not deleted until the finalizers are gone / there's a whole complex dance there getting ordering right? 

> And I'm extremely familiar with it.

I wrote the logic and tests 🍿 Not sure what you're implying here.

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
theobarberbany: Literally just a pointer / copy of the one where the pattern is started? 

```go
		// NOTE: This is a workaround for the deeply nested test structure
		// which makes it difficult to control the order of setup operations.
		// Do not copy this pattern to new tests or extend its usage. New
		// tests should be structured so that this kind of deferred setup is
		// not necessary.
```

belt and braces, we have to assume people without context using AI will be working here :(

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
damdo: This one is a bit subtle.
IIRC the thing is an interpretation issue object: exists but has deletion timestamp (just got issued a delete) != object does exist anymore (got issued a delete and now is gone).

When one says "it is deleted" it can leave it up for interpretation, which is why I think originally we went with `"when the CAPI machine set has a non-zero deletion timestamp"`.

Also we don't just use it here, it is used in other 3/4 places, so in case we change this, we'd need to change it everywhere.

Personally I'd prefer to also keep it like that if possible, unless there is a very strong argument against. TY

damdo (commented):
Thanks!

/approve
/lgtm

/verified by units

pkg/controllers/machinesetsync/machineset_sync_controller_test.go
damdo: This is now done.

damdo:
This is a units change, no need for e2es


/override ci/prow/e2e-aws-capi-techpreview ci/prow/e2e-aws-capi-techpreview-post-install ci/prow/e2e-aws-ovn-techpreview ci/prow/e2e-aws-ovn-techpreview-upgrade ci/prow/e2e-azure-capi-techpreview ci/prow/e2e-azure-ovn-techpreview-upgrade ci/prow/e2e-gcp-capi-techpreview ci/prow/e2e-openstack-capi-techpreview ci/prow/e2e-vsphere-capi-techpreview

PR #552, opened 2026-05-11
OCPCLOUD-2664: Update operatorstatus to write correct sub-Conditions

mdbooth:
## Summary by CodeRabbit

* **Refactor**
  * Streamlined operator status condition handling: Progressing is always tracked; Available is only set when explicitly determined.
  * Internal condition representation simplified; degraded condition removed.

* **Bug Fixes**
  * Prevents asserting Available=True before first explicit availability reconciliation.

* **Tests**
  * Updated and added tests to reflect new Available/Progressing semantics and condition lookup behavior.

### DIFF
--- a/pkg/operatorstatus/controller_status.go
+++ b/pkg/operatorstatus/controller_status.go
@@ -59,6 +59,14 @@ const (
 
 	// ReasonNonRetryableError indicates that the controller encountered a non-retryable error.
 	ReasonNonRetryableError = "NonRetryableError"
+
+	// ConditionAvailableSuffix is the suffix added to a controller prefix to
+	// form the controller's available condition type.
+	ConditionAvailableSuffix = "Available"
+
+	// ConditionProgressingSuffix is the suffix added to a controller prefix to
+	// form the controller's progressing condition type.
+	ConditionProgressingSuffix = "Progressing"
 )
 
 // CAPIFieldOwner returns a qualifiedclient.FieldOwner for the given qualifier.
@@ -68,16 +76,46 @@ func CAPIFieldOwner[S ~string](qualifier S) client.FieldOwner {
 	return client.FieldOwner(CAPIOperatorIdentifierDomain + "/" + qualifier)
 }
 
+type partialCondition struct {
+	status  configv1.ConditionStatus
+	reason  string
+	message string
+}
+
 // ReconcileResult represents the result of a controller's reconciliation.
 // As well as returning a reconcile.Result to controller-runtime, it can also
 // write the result as conditions on the ClusterOperator.
 type ReconcileResult struct {
 	ControllerResultGenerator
 
-	progressing  *configv1apply.ClusterOperatorStatusConditionApplyConfiguration
-	degraded     *configv1apply.ClusterOperatorStatusConditionApplyConfiguration
+	// a ReconcileResult must always have an explicit progressing condition
+	progressing partialCondition
+
+	// the available condition is optional, and will be maintained from the
+	// current state if not set explicitly
+	available *partialCondition
+
 	err          error
 	requeueAfter time.Duration
+}
+
+// Internal constructors for ReconcileResult. Callers do not set these properties directly.
+
+func newReconcileResult(controllerResultGenerator ControllerResultGenerator, progressingStatus configv1.ConditionStatus, reason, message string) ReconcileResult {
+	return ReconcileResult{
+		ControllerResultGenerator: controllerResultGenerator,
+		progressing:               partialCondition{progressingStatus, reason, message},
+	}
+}
+
+func (r ReconcileResult) withAvailable(status configv1.ConditionStatus, reason, message string) ReconcileResult {
+	r.available = &partialCondition{status, reason, message}
+	return r
+}
+
+func (r ReconcileResult) withError(err error) ReconcileResult {
+	r.err = err
+	return r
 }
 
 // Result returns a reconcile.Result for controller-runtime.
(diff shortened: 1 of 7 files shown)


pkg/operatorstatus/controller_status_test.go
```diff
@@ -393,52 +397,70 @@ func newFakeClient(objs ...client.Object) client.WithWatch {
...
+		},
+		{
+			name:               "Progressing with existing Available preserves Available",
+			existingConditions: existingAvailable,
+			result:             testResultGenerator.Progressing("installing components"),
+			wantProgressing:    expectedCondition{configv1.ConditionTrue, ReasonProgressing, "installing components"},
+			wantAvailable:      &expectedCondition{configv1.ConditionTrue, ReasonAsExpected, "Success"},
+		},
+		{
+			name:               "Error with existing Available preserves Available",
```
damdo, line 445:
```suggestion
			name:               "EphemeralError with existing Available preserves Available",
```

pkg/operatorstatus/controller_status_test.go
```diff
@@ -393,52 +397,70 @@ func newFakeClient(objs ...client.Object) client.WithWatch {
...
+			existingConditions: existingAvailable,
+			result:             testResultGenerator.Error(fmt.Errorf("connection refused")),
+			wantProgressing:    expectedCondition{configv1.ConditionTrue, ReasonEphemeralError, "connection refused"},
```
damdo, line 448: Although not introduced in this PR, I'm curious why `EphemeralError` vs `NonRetryableError` instead of `RetryableError` vs `NonRetryableError` :)

pkg/operatorstatus/controller_status.go
```diff
@@ -229,10 +242,43 @@ func (r *ReconcileResult) WriteClusterOperatorStatus(ctx context.Context, log lo
...
+	// During installation, we don't want to declare Available=True until the
+	// first time we have successfully reconciled. We achieve this by initially
+	// not setting the Available condition at all. The condition aggregator will
+	// interpret this correctly.
+	//
+	// After we have become Available for the first time, we don't want to
+	// declare Available=False except for conditions which require an
+	// administrator's intervention. Currently we set it for non-retryable
+	// errors. Otherwise we copy the previous state of the Available condition
+	// if it exists.
```
damdo, line 266: Here you hint at the fact that you followed the godoc (I guess [this one](https://github.com/openshift/api/blob/9b69e5fa27e94ac3e4f61a94b56e8a0034ea4aee/config/v1/types_cluster_operator.go#L151-L211)), when implementing this.

I know there is also a more recent enhancement describing this ([here](https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/clusteroperator.md#conditions)).

I did a quick diffing and I don't see any major differences between the two sources, so we should be good there.


> Key differences
>   1. Level of detail on Progressing
>     - The API is more prescriptive: it explicitly says operators should not report Progressing just because owned resources (DaemonSets, Deployments) are adjusting to a new node from cluster scale-up or a node rebooting during an upgrade. It also adds that if the observed cluster state has changed and the component is reacting (e.g., updated proxy config), Progressing should become true.
>     - The enhancement guide covers the same spirit but is briefer, focusing on the time constraints and the general "don't report Progressing for reconciliation of known state" rule.
>   2. Progressing message guidance
>     - The enhancement guide adds specific UX guidance not present in the API types: the Progressing message should be a "terse, human-readable message describing the current state of the object in 5-10 words."
>     - The API types have no such message formatting guidance.
>   3. Upgradeable semantics
>     - The API is more precise: it distinguishes between ClusterVersion scope (where the message explains which updates — minor or patch — are impacted) and ClusterOperator scope (where False specifically blocks minor OpenShift updates). It also explicitly states the CVO allows updates when the condition is missing, True, or Unknown.
>     - The enhancement guide covers the same idea but doesn't differentiate ClusterVersion vs. ClusterOperator scope as explicitly.
>   4. Degraded nuances
>     - Both sources agree on the semantics (persistent observation, can coexist with Available, should not oscillate, must not report during normal upgrades).
>     - The API adds more detail: it mentions that Degraded may be reported for persistent infrastructure failures requiring admin intervention (e.g., unhealthy control plane host), and that unexpected errors should be handled as operators mature.
>   5. Core vs. Optional classification
>     - The enhancement guide explicitly categorizes Available/Progressing/Degraded as "Core Conditions" and Upgradeable/EvaluationConditionsDetected as "Optional Conditions".
>     - The API types make no such distinction — all five are simply listed as constants.
>   6. EvaluationConditionsDetected
>     - The API provides more detail: it mentions using :: as a delimiter to concatenate multiple Reasons when evaluating multiple changes.
>     - The enhancement guide gives only a brief summary.
>

pkg/operatorstatus/controller_status.go
```diff
@@ -68,18 +76,48 @@ func CAPIFieldOwner[S ~string](qualifier S) client.FieldOwner {
...
 
-	progressing  *configv1apply.ClusterOperatorStatusConditionApplyConfiguration
-	degraded     *configv1apply.ClusterOperatorStatusConditionApplyConfiguration
+	// a ReconcileResult must always have an explicit progressing condition
+	progressing partialCondition
+
+	// the available condition is optional, and will be maintained from the
+	// current state if not set explicitly
+	available *partialCondition
+
```
damdo, line 97: So are we planning not to set degraded ever at all?

damdo (commented):
Thanks, mostly LGTM, a couple of nits, a couple of Qs.

pkg/operatorstatus/controller_status.go
mdbooth: Yes, in the aggregator. I'm thinking: any Progressing condition (error or not) which persists for more than duration X, where X is the colour of your bike shed.

pkg/operatorstatus/controller_status_test.go
mdbooth: I've been using these words for a while, so evolution mostly. I used to use 'Ephemeral' and 'Terminal'. But at some point I switched 'Terminal' to 'NonRetryable' because its purpose is more explicit.

I'm still ok with this, tbh. I've committed worse sins.

damdo (commented):
/approve
/lgtm

mdbooth:
/override e2e-openstack-ovn-techpreview

Permafailing

mdbooth:
/override pull-ci-openshift-cluster-capi-operator-main-e2e-openstack-ovn-techpreview

damdo:

Unrelated issue.

mdbooth:
Confirmed the conditions written in e2e-aws-capi-techpreview

/verified by CI

PR #567, opened 2026-05-20
OCPCLOUD-2899: Support converting taints between MAPI and CAPI

nrb:
This work relies on having CAPI v1.12, though it doesn't look like bumped yet, possibly because of the upstream CAPA bump (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5857) being in progress.

## Summary by CodeRabbit

* **New Features**
  * Taints are now properly converted and synchronized bidirectionally between machine configurations, with correct effect mapping and propagation handling.

* **Tests**
  * Added comprehensive test coverage for taint conversion, including single and multiple taints with proper effect mapping scenarios.

* **Chores**
  * Removed unused cluster-api dependency directive from module configuration.


tthvo:
> This work relies on having CAPI v1.12, though it doesn't look like bumped yet, possibly because of the upstream CAPA bump (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5857) being in progress.

:wave: May I ask some questions:
- Do you think we need to also enable feature gate [MachineTaintPropagation](https://github.com/kubernetes-sigs/cluster-api/blob/a2247ea336940fbaaf4b60fd38a49ab20dff414e/feature/feature.go#L88-L91) for CAPI controller? I see it is `false` by default :thinking:
- Will this `taints` field be also available in `v1beta1` API? I see upstream https://github.com/kubernetes-sigs/cluster-api/commit/4c81f12e0cd2a33171c0ff3a623a0ca81b0b90d4 brought it to `v1beta1` so probably it will in the next sync.

I'm asking because the installer still uses `v1beta1` to generate CAPI machines and we need to set node taints for edge machine pool.  Or I guess since `v1beta2` is now the storage version (right?), the installer should start using it instead :eyes:

nrb:
@tthvo 

> Do you think we need to also enable feature gate [MachineTaintPropagation](https://github.com/kubernetes-sigs/cluster-api/blob/a2247ea336940fbaaf4b60fd38a49ab20dff414e/feature/feature.go#L88-L91) for CAPI controller? I see it is false by default 🤔

Yeah, we'll need to turn that on.

> Will this taints field be also available in v1beta1 API? I see upstream https://github.com/kubernetes-sigs/cluster-api/commit/4c81f12e0cd2a33171c0ff3a623a0ca81b0b90d4 brought it to v1beta1 so probably it will in the next sync.

It looks like that backport was made against the 1.13 series, not 1.12? I don't see any mention of a git backport in https://github.com/kubernetes-sigs/cluster-api/pull/13455.

IMO we should try to move to v1beta2 if it's not too painful...but given the delay on CAPA, it might be.

tthvo:
> It looks like that backport was made against the 1.13 series, not 1.12? I don't see any mention of a git backport in https://github.com/kubernetes-sigs/cluster-api/pull/13455.
>
> IMO we should try to move to v1beta2 if it's not too painful...but given the delay on CAPA, it might be.

Arghh, sorry, I didn't look at the PR date... 🤦 Anywayy, +1 to use v1beta2! On the installer side, we did move to generate CAPI manifests with `v1beta2` for workers after all. There are few more places for control plane, which will eventually move too. Thanks!

damdo:
Hey @nrb what's next on this one? TY

PR #568, opened 2026-05-21
[release-4.22] OCPBUGS-86353: fix: scope webhook to capi namespace, remove unused webhook endpoints

damdo:
See: https://redhat.atlassian.net/browse/OCPBUGS-86353

Cherry-pick of https://github.com/openshift/cluster-capi-operator/pull/543 to release-4.22.

On unsupported platforms (e.g. IBMCloud), the openshift-capi-controllers ValidatingWebhookConfiguration is deployed by CVO but the webhook server does not start, causing Cluster creation in any namespace to fail with "connection refused". This blocks MCE/ACM which install CAPI independently and create Cluster objects in their own namespace.

Scope the remaining Cluster webhook to the openshift-cluster-api namespace via namespaceSelector so it no longer intercepts requests in other namespaces.

Remove the coreprovider, infrastructureprovider, and clusterctl provider webhook entries which had no corresponding handler code registered and were for CRDs no longer deployed by this operator.


damdo:
/verified by CI

If CI passes this can auto-merge

PR #572, opened 2026-05-29
NO-JIRA: fix shell compatibility in Makefile

simkam:
Use POSIX-compliant `=` instead of `==` and add quotes around the HOME variable in the ensure-home function.

This fixes the error "/bin/sh: 1: [: unexpected operator" that occurs on systems where /bin/sh is symlinked to dash instead of bash.

## Summary by CodeRabbit

* **Chores**
  * Enhanced build system stability and portability through improved environment configuration handling.


mdbooth:
Both fmt and lint call this function, both covered in presubmits.

/verified by CI

PR #575, opened 2026-06-03
OCPBUGS-86770: Fix OpenStack identityRef conversion and add e2e test

simkam:
https://redhat.atlassian.net/browse/OCPBUGS-86770

The MAPI-to-CAPI conversion for OpenStack was not setting the Type field on OpenStackIdentityReference. While the CRD has a kubebuilder default of "Secret", this only applies at the API server level. Programmatic struct construction yields a zero-value empty string, which differs from the stored value. This caused the differ to detect a spec change on every reconciliation, triggering an infinite delete/recreate loop of the OpenStackMachine (infra machines are immutable, so spec changes require delete+recreate). The symptom was the Synchronized condition oscillating between True and False.

The fix explicitly sets Type: "Secret" in the conversion function.

Add a platform-agnostic e2e test (machine_sync_test.go) that verifies the Synchronized condition reaches True and stays stable via Consistently, catching the oscillation without needing platform-specific assertions.

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test suite covering machine synchronization between MAPI and CAPI, verifying machines report synchronized state and linked infrastructure resources remain stable over time.

* **Bug Fixes**
  * Corrected OpenStack cloud provider identity reference so secret-based identity payloads include the proper type field.

### DIFF
--- a/e2e/machine_sync_test.go
+++ b/e2e/machine_sync_test.go
@@ -1,0 +1,85 @@
+// Copyright 2026 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
+	mapiv1beta1 "github.com/openshift/api/machine/v1beta1"
+	"github.com/openshift/cluster-capi-operator/e2e/framework"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/cluster-api/controllers/external"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+)
+
+var _ = Describe("Machine Sync", Ordered, func() {
+	BeforeAll(func() {
+		switch platform {
+		case configv1.AWSPlatformType, configv1.OpenStackPlatformType:
+			// supported
+		default:
+			Skip(fmt.Sprintf("Machine sync is not supported on %s", platform))
+		}
+
+		if !framework.IsFeatureGateEnabled(ctx, cl, features.FeatureGateMachineAPIMigration) {
+			Skip("MachineAPIMigration feature gate is not enabled")
+		}
+	})
+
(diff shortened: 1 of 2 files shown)


e2e/machine_sync_test.go
```diff
@@ -0,0 +1,75 @@
...
+			g.Expect(mapiMachine.Status.Conditions).To(ContainElement(SatisfyAll(
+				HaveField("Type", Equal(SynchronizedCondition)),
+				HaveField("Status", Equal(corev1.ConditionTrue)),
+			)))
+
+			freshInfraMachine, err := external.GetObjectFromContractVersionedRef(ctx, cl, capiMachine.Spec.InfrastructureRef, capiMachine.Namespace)
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(freshInfraMachine.GetUID()).To(Equal(infraMachineUID),
+				"infrastructure machine UID changed — was deleted and recreated")
+		}, "30s", "5s").Should(Succeed())
```
mdbooth, line 83: We can't have a 30s wait here. These add up so fast in a test suite.

Let me think about this...

simkam: Locally, without the fix, it actually fails on the first check of the synchronized condition on line 50. It might be enough to detect a problem, and we might drop the rest.

mdbooth: I think it'll flap, though, so it's non-deterministic. I didn't come up with a better idea, so lets do this.

simkam:
Rebased and the new test skipped on platforms where [MachineAPIMigration isn't supported](https://github.com/openshift/cluster-capi-operator/blob/05c113ea5edd0aaf43e34c4a534a673d457449c1/cmd/machine-api-migration/main.go#L152).

simkam:
/retest-required

```
  1. e2e-openstack-ovn-techpreview: Missing openstack-tests binary (test environment issue)
  2. e2e-aws-ovn-techpreview: Flaky timeouts in router/network/build tests (unrelated to CAPI)
  3. e2e-azure-ovn-techpreview: Flaky timeouts in deployment tests (unrelated to CAPI)
  4. e2e-aws-capi-disconnected-techpreview: Cleanup failures during deprovision (infrastructure leak issue)
```

stephenfin:
/test e2e-openstack-ovn-techpreview

@mandre has fixed those missing binary errors in https://github.com/openshift/release/pull/79254 and https://github.com/openshift/release/pull/80391.

stephenfin (commented):
/approve
/lgtm

Thanks for working on this 🙏

simkam:
/verified by "Machine Sync should synchronize a MAPI Machine to CAPI with a stable Synchronized condition"

AI analysis of failed jobs:

1. ✅ e2e-aws-capi-disconnected-techpreview - Known flake, fix available
    - Failure: MachineSet scaling conflict error ("Operation cannot be fulfilled on machinesets.machine.openshift.io: the object has been modified")
    - Root cause: Race condition in ScaleMachineSet() - no retry logic for conflicts
    - Fix: Already merged in https://github.com/openshift/cluster-api-actuator-pkg/pull/485 (adds exponential backoff retry)
    - Will be fixed by: PR #598 which vendors the fix into cluster-capi-operator
    - Not caused by this PR: This is a pre-existing flake in the test framework
  2. ✅ e2e-azure-ovn-techpreview - Infrastructure flake
    - 13 failures out of 4294 tests (0.3% failure rate)
    - All in unrelated areas: network segmentation, DRA, storage
    - Not CAPI-related, not caused by this PR
  3. ✅ e2e-openstack-ovn-techpreview - Infrastructure DNS issue
    - Failed during setup: DNS not resolving cluster API endpoint
    - Infrastructure issue, not caused by this PR

PR #577, opened 2026-06-04
OCPCLOUD-3442: Migrate four controllers to per-controller condition reporting

stefanonardo:
## Summary

- Migrate CoreCluster, InfraCluster, Kubeconfig, and SecretSync controllers from `ClusterOperatorStatusClient` (merge-patch, shared top-level conditions) to `ControllerResultGenerator` (server-side apply, per-controller prefixed conditions)
- Each controller now reports its own `Available` and `Progressing` conditions on the ClusterOperator object, which are aggregated into top-level conditions by the ClusterOperator controller
- Delete `ClusterOperatorStatusClient` and `GetClusterOperatorStatusClient` — no remaining consumers
- Add ClusterOperator watch to SecretSync for initial reconciliation at startup — on platforms where the source secret does not exist (e.g. metal3, where CAPM3 reads user data directly via a namespace-qualified `SecretReference`), report success instead of blocking the install

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/controllers/kubeconfig/kubeconfig.go
```diff
@@ -81,53 +84,38 @@ func (r *KubeconfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
 	if infra.Status.PlatformStatus == nil {
 		log.Info("No platform status exists in infrastructure object. Skipping kubeconfig reconciliation...")
-
-		if err := r.SetStatusAvailable(ctx, ""); err != nil {
-			return ctrl.Result{}, fmt.Errorf("error syncing ClusterOperatorStatus: %w", err)
-		}
-
-		return ctrl.Result{}, nil
+		return ResultGenerator.Success()
```
mdbooth: This should probably be WaitingOnExternal.

You should also check if this controller watches the Infrastructure object (check in SetupWithManager). If it doesn't we'll either need to watch it or also add a RequeueAfter here. Polling should be fine in this case if we don't already have a watch: we don't really expect this to happen.

stefanonardo: we don' watch it, I'll requeue it after 1 minute

stefanonardo:
@coderabbitai review

mdbooth (commented):
/lgtm
/approve

PR #581, opened 2026-06-05
NO-JIRA: Fix continous reconciliation of VAPs due to server-side defaulting

mdbooth:
matchConstraints has several fields which have server-side defaults. If omitted, these trigger a continuous reconciliation of all managed objects because the spec doesn't match the manifest. There is unfortunately no good way to detect server-side defaulted values, so the easiest solution is to specify them explicitly.

This is a non-functional change. All the added values are just the defaults.

## Summary by CodeRabbit

* **Configuration Updates**
  * Improved consistency and standardized behavior for cluster resource validation policies across creation and update operations, ensuring more uniform policy application across namespaces and resources.


damdo (commented):
/approve
/lgtm

PR #583, opened 2026-06-08
[wip] NO-MERGE: boxcutter: retry SSA field manager migration on 409 conflict

stbenjam:
The csaupgrade.UpgradeManagedFieldsPatch function generates a JSON patch that includes the object's resourceVersion as an optimistic lock. When the kube-apiserver's CRD controller updates a CRD's status conditions immediately after creation, the resourceVersion is bumped before the field manager migration patch arrives, causing a 409 Conflict.

Without retry logic, this conflict causes the InstallerController to enter an infinite error loop: each reconciliation re-reads from cache (which may still be stale), regenerates the same patch with the old resourceVersion, and hits the same 409.

This was observed as a deterministic bootstrap failure in payload 5.0.0-0.nightly-2026-06-06-100407 where ALL TechPreview jobs across all cloud providers failed to bootstrap. The InstallerController was stuck retrying SSA migration on CAPI CRDs and ValidatingAdmissionPolicies, preventing worker MachineSets from ever being created.

Fix: add a retry loop (max 5 attempts) that re-reads the object from cache after a conflict to pick up the latest resourceVersion before retrying the patch.

### DIFF
--- a/vendor/pkg.package-operator.run/boxcutter/machinery/objects.go
+++ b/vendor/pkg.package-operator.run/boxcutter/machinery/objects.go
@@ -2,9 +2,9 @@ package machinery
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strconv"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


mdbooth:
@stbenjam The issue is that every write operation on a CRD sent to the bootstrap KAS suffers a 20s timeout because the bootstrap KAS can't reach webhooks. It will be fixed by https://github.com/openshift/installer/pull/10344.

In the absence of a 20s timeout boxcutter normally wins this race. The addtional retry complexity is not required, as the controller already provides exponential backoff in the (normally) very unlikely event that it loses the race. The 20s timeout is a bug which we already have a fix for, so it's not worth adding code for.

The reason the race exists is that there is no k8s atomic create primitive which populates SSA metadata. The library wants to create an object which it intends to manage with SSA. It wants to return an error if the object unexpectedly already exists and does not match any given adoption criteria.

If the clanker still wants to play with it, boxcutter can be found here: https://github.com/package-operator/boxcutter.

PR #586, opened 2026-06-09
OCPCLOUD-3465: feat: implement cluster-wide proxy support via new ProxyController

tmshort:
Add a ProxyController to the capi-installer manager that watches managed
workloads opting into proxy injection via the new annotation
  capi-operator.openshift.io/inject-proxy: <container,...>
and applies HTTP_PROXY/HTTPS_PROXY/NO_PROXY environment variables to the
named containers via Server-Side Apply whenever the cluster-wide Proxy CR
changes.

Design (Option B from [OCPCLOUD-3465](https://redhat.atlassian.net/browse/OCPCLOUD-3465) Jira):
- A dedicated controller in the capi-installer manager watches the Proxy
  singleton (proxies/cluster) and all managed objects via the existing
  trackingCache, so it never creates a second informer for those objects.
- Env vars are applied with a separate SSA field manager (capi-operator-proxy)
  that owns only the three proxy env var entries per container. Boxcutter
  continues to own all other fields on the same workload object without
  conflict.
- Proxy changes propagate to running pods independently of the revision
  lifecycle, including when the installer has fallen back to an older
  revision (N-1) because the newest revision failed to roll out.
- Starting with Option B avoids a future field-ownership migration: if
  Option A (inject at render time) had been deployed first, converting to
  Option B later would require force-stealing SSA ownership from boxcutter.

The annotation is added to provider Deployments by manifests-gen (see next
commit). Providers that use different container names or need proxy injection
on additional containers should set the annotation in their upstream manifests.

SetupWithManager now returns the TrackingCache so it can be shared with
SetupProxyController without creating a second managed-object cache.

## Summary by CodeRabbit

* **New Features**
  * Automatic proxy injection: cluster proxy settings are applied to managed Deployments, DaemonSets, and StatefulSets when annotated, adding HTTP_PROXY/HTTPS_PROXY/NO_PROXY envs.
  * Dedicated proxy reconciler: operator wires a proxy controller to propagate proxy changes to tracked workloads and target specified containers.

* **Tests**
  * Test harness metrics endpoint and default TLS updated for stronger TLS (min TLS 1.2 and modern cipher) and explicit metrics binding.


tmshort:
@mdbooth this is a replacement for #517. Does it match expectations?

mdbooth:
There's a few problems here. The biggest one is that `config.openshift.io/inject-proxy` isn't handled by an admission webhook or even a controller watching annotated objects (yeah, this surprised me too). It's handled by CVO's resource builder. Therefore because these are not CVO resources, the proxy env vars won't be applied.

The other big thing is that I don't want the revision/installer controllers to contain *any* heuristics about providers. This is why for the TLS stuff I made the providers include explicit substitution points for the TLS vars. In this case the heuristic is that we only want to add proxy vars to a D/DS/SS with a container called 'manager'. This isn't true today (OpenStack already packages ORC, which doesn't follow this pattern), and I only expect the exceptions to expand over time.

The takeaways are:
* We need to implement our own mechanism for applying proxy env vars, because the existing one isn't reusable
* The new mechanism can (and should) copy the behaviour of the existing mechanism, but if it's annotation based it must use a different name to avoid confusion
* If using an annotation based mechanism, it must be applied in the provider repo

So perhaps something like an annotation `capi-operator.openshift.io/inject-proxy` which has identical semantics to `config.openshift.io/inject-proxy`, but isn't going to confuse anybody into wondering how CVO is operating on a non-CVO object. We could update manifests-gen to apply this automatically to any D/DS/SS with a container called `manager`, but the provider can also add it manually anywhere else it's required, or even override the manifests-gen generated default on provider deployments if required.

This also means we can't use the PROXY_HASH mechanism, because if we're going to apply the proxy env vars directly they need to be included in the revision in full.

I also had another option in mind. We could create a new controller which watches objects managed by the installer controller which have the new annotation, and adds the env vars after creation. This works because:
* The installer's trackingCache should be reusable by a second controller
* Boxcutter uses SSA, so a second controller can add new fields without conflict as long as it doesn't change existing fields (which is not true for TLS)
* It covers the edge case where we want to apply a new proxy configuration to revision N-1 because revision N is failing to roll out

This may or may not be simpler, not sure. It's something to consider.

I think we need to:
* Decide on a provider contract
* Implement it in manifests-gen
* Roll it out to all providers
* Implement it in capi-installer

tmshort:
Reevaluating based on @mdbooth's comments.

tmshort:
@coderabbit help

tmshort:
@coderabbitai full review

manifests-gen/customizations.go
```diff
@@ -162,12 +162,33 @@ func findWebhookServiceSecretName(objs []client.Object) map[string]string {
 	return serviceSecretNames
 }
 
+const proxyInjectAnnotation = "capi-operator.openshift.io/inject-proxy"
```
mdbooth: I'm pretty sure we've got a constant defined somewhere for `capi-operator.openshift.io`. Please can we use it here?

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+
+const (
+	proxyControllerName = "ProxyController"
+
+	// ProxyInjectAnnotation is the annotation placed on a pod template to opt
+	// into cluster-wide proxy injection. The value is a comma-separated list of
+	// container names that should receive HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+	// environment variables. The annotation is added by manifests-gen and can be
+	// overridden by provider manifests.
+	ProxyInjectAnnotation = "capi-operator.openshift.io/inject-proxy"
```
mdbooth: Perhaps use this constant from manifests-gen? I can't remember which way round we configured the dependencies (note that manifests-gen is a separate module).

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+	deployList := &appsv1.DeploymentList{}
+	if err := c.trackingCache.List(ctx, deployList); err != nil {
+		return nil, fmt.Errorf("listing managed deployments: %w", err)
+	}
+
+	for i := range deployList.Items {
+		d := &deployList.Items[i]
+		items = append(items, workloadItem{
+			apiVersion:     "apps/v1",
+			kind:           "Deployment",
```
mdbooth: micro-nit: I'd personally have pulled these from the item, too.

Don't change this unless you're particularly inspired to.

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+				"template": map[string]interface{}{
+					"spec": map[string]interface{}{
+						"containers": containerPatches,
+					},
+				},
+			},
+		},
+	}
+
+	if err := c.client.Patch(ctx, patch, util.ApplyConfigPatch(patch.Object), client.FieldOwner(proxyFieldManager), client.ForceOwnership); err != nil {
```
mdbooth: The field owner needs to be qualified (i.e. capi-operator.../thingy). The domain should again use an existing constant. Depending on where the constant is currently, we might have to move it somewhere more sensible.

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+		return nil
+	}
+
+	containerPatches := make([]map[string]interface{}, 0, len(containerNames))
+	envEntries := proxyEnvEntries(proxyVars)
+
+	for _, containerName := range containerNames {
+		containerPatches = append(containerPatches, map[string]interface{}{
+			"name": containerName,
+			"env":  envEntries,
```
mdbooth: nit: I'd bet this could be simplified to just pass `proxyVars` directly here and get rid of the `proxyEnvEntries` function.

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+
+		for _, name := range allContainers {
+			if _, ok := targetSet[name]; !ok {
+				stale = append(stale, name)
+			}
+		}
+
+		if err := c.applyProxyVars(ctx, item.apiVersion, item.kind, item.name, item.namespace, stale, nil); err != nil {
+			errs = append(errs, err)
+		}
```
mdbooth: This looks like it would always result in the proxy vars being cleared. We're submitting a second, empty SSA with the same field owner. That's just going to clear the env vars we just set in the other container, right?

Edit: ah... now I've spotted the short-circuit in applyProxyVars. This is still wrong. If you've got a pod with a manager and another container, the apply on the second one will wipe the apply on the first one. The SSA needs to be fully specified.

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+		annotation, hasAnnotation := item.podAnnotations[ProxyInjectAnnotation]
+
+		if !hasAnnotation {
+			// No annotation: clear any stale proxy env vars from all containers.
+			if err := c.applyProxyVars(ctx, item.apiVersion, item.kind, item.name, item.namespace, allContainers, nil); err != nil {
+				errs = append(errs, err)
+			}
+
+			continue
+		}
```
mdbooth: I think this is redundant. If the annotation is not set, all we need is for containerNamesFromAnnotation below to return empty targets and any env vars previously set by us will be cleared when we construct an empty SSA.

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+	ctx context.Context,
+	apiVersion, kind, name, namespace string,
+	containerNames []string,
+	proxyVars []corev1.EnvVar,
+) error {
+	log := ctrl.LoggerFrom(ctx).WithName(proxyControllerName)
+
+	if len(containerNames) == 0 {
+		return nil
+	}
```
mdbooth: This is problematic. We need to be able to submit an empty transaction.

pkg/controllers/installer/proxy_controller.go
```diff
@@ -0,0 +1,344 @@
...
+
+	return items, nil
+}
+
+// applyProxyVars SSA-applies proxy env vars to each named container on the
+// workload object. When proxyVars is empty/nil (proxy removed or no annotation),
+// it applies an empty env list, which removes any env var entries previously
+// owned by the proxy controller's SSA field manager. Only the three proxy env
+// var entries are owned; all other fields remain owned by boxcutter.
+func (c *ProxyController) applyProxyVars(
```
mdbooth: I think this needs an optimisation to prevent a potential apply storm during installation. The controller triggers on modification to any object, but for each modified object reconciles the Proxy object and iterates over all objects. We've got a bit of matrix amplification going here.

That's ok, but we're also unconditionally submitting an SSA for every workitem on every reconcile, even if it's blank. That's not very nice to KAS.

The typed apply configuration objects have an ExtractInto function which will extract all fields managed by a particular field owner. I wonder if an equivalent exists for Unstructured?

We could:
* construct `patch` as we're already doing below
* ExtractInto a new unstructured with the same field owner
* If patch deepequals existing, I think we can skip it

pkg/controllers/revision/helpers_test.go
mdbooth, line 1: Any idea why these changes were required? Doesn't look like we should be touching these.

pkg/controllers/revision/helpers_test.go
```diff
@@ -56,15 +57,15 @@ func newManagerWrapper(providerImgs []providerimages.ProviderImageManifests, tls
...
 		},
+		Metrics: metricsserver.Options{BindAddress: "0"},
 	})
 	Expect(err).NotTo(HaveOccurred())
 
 	if len(tlsOptions) == 0 {
 		tlsOptions = []func(config *tls.Config){
 			func(config *tls.Config) {
-				// Arbitrarily chosen insecure default tls configuration for tests
-				config.CipherSuites = []uint16{tls.TLS_RSA_WITH_RC4_128_SHA}
```
mdbooth, line 66: Especially shouldn't be changing these. Was this code rabbit thinking deliberately insecure test defaults are a security issue by any chance?

mdbooth (changes requested):
Nice. I think this is most of the way there. I've requested some code cleanup things like use of constants, using a fully qualified field owner, etc.

Big things:

Please can you put the new controller in its own package? This is mostly for consistency. We've got a bunch of controller, and they're all separated in their own package.

I'm pretty sure the SSA handling in the workitem loop isn't right. An SSA fully specifies everything owned by the field owner on the resource. Multiple SSAs with the same field owner will just overwrite each other. Fully construct the desired state (owned fields only) and submit a single SSA. Any which weren't specified will be removed automatically.

No tests. I think this would have highlighed the above. These should be envtests. We have a ton of examples, so Claude will write them for you in a minute.

I think the 'nothing to do' optimisation in `applyProxyVars` is important.

pkg/controllers/revision/helpers_test.go
tmshort: I believe it was. Since the test have two stages; I couldn't tell if it passed or not. :(

pkg/controllers/revision/helpers_test.go
tmshort: It was likely a CodeRabbit comment; it ought to be revertable.

tmshort: But the BindAddress was changed to explicitly be "0" so that a port can be randomly selected, and subsequently allowing more parallelization of tests.

tmshort:

Test at least one of the e2es.

tmshort:
rebased

tmshort:

Test at least one of the e2es.

pkg/controllers/proxy/proxy_controller.go
```diff
@@ -0,0 +1,384 @@
...
+	// ProxyInjectAnnotation is the annotation placed on a pod template to opt
+	// into cluster-wide proxy injection. The value is a comma-separated list of
+	// container names that should receive HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+	// environment variables. The annotation is added by manifests-gen and can be
+	// overridden by provider manifests.
+	ProxyInjectAnnotation = operatorstatus.CAPIOperatorIdentifierDomain + "/inject-proxy"
+
+	// proxyVarNames are the only env var names this controller ever owns.
+	// Used for the skip-if-unchanged optimisation.
+
```
mdbooth: Stale?

pkg/controllers/proxy/proxy_controller.go
```diff
@@ -0,0 +1,384 @@
...
+				return obj.GetName() == proxyClusterName
+			})),
+		).
+		Watches(&appsv1.Deployment{}, handler.EnqueueRequestsFromMapFunc(toProxy),
+			builder.WithPredicates(isManagedWorkload)).
+		Watches(&appsv1.DaemonSet{}, handler.EnqueueRequestsFromMapFunc(toProxy),
+			builder.WithPredicates(isManagedWorkload)).
+		Watches(&appsv1.StatefulSet{}, handler.EnqueueRequestsFromMapFunc(toProxy),
+			builder.WithPredicates(isManagedWorkload)).
+		Complete(c)
```
mdbooth: I don't think is right. I think we need to use the tracking cache here for 2 reasons:

* This will start new D/DS/SS informers rather than using the tracking cache's existing watches, so it uses more memory/KAS capacity
* This will be subject to the DefaultNamespace cache options we configured in the entrypoint, so we won't watch anything the provider installs outside of the default namespaces

On the latter point, although we expect providers to install things in the default namespace I don't think we want or need to hardcode this requirement.

pkg/controllers/proxy/proxy_controller.go
```diff
@@ -0,0 +1,384 @@
...
+	// stopping cleanly and can leave orphaned envtest processes.
+	managedReq, err := labels.NewRequirement(revisiongenerator.ManagedLabelKey, selection.Exists, nil)
+	if err != nil {
+		return fmt.Errorf("building managed label requirement: %w", err)
+	}
+
+	isManagedWorkload := predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		return labels.NewSelector().Add(*managedReq).Matches(labels.Set(obj.GetLabels()))
+	})
+
```
mdbooth: If we use the tracking cache directly I think we don't need (or want) this.

pkg/controllers/proxy/proxy_controller.go
```diff
@@ -0,0 +1,384 @@
...
+// controller reconciles for every newly-applied managed object).
+func (c *Controller) applyProxyVars(ctx context.Context, item workloadItem, proxyVars []corev1.EnvVar) error {
+	log := ctrl.LoggerFrom(ctx).WithName(controllerName)
+
+	if len(item.containers) == 0 {
+		return nil
+	}
+
+	// Build the set of containers that should receive proxy env vars.
+	annotatedSet := make(map[string]struct{})
```
mdbooth: nit: Set[string] is idiomatic in this codebase

pkg/controllers/proxy/proxy_controller.go
```diff
@@ -0,0 +1,384 @@
...
+		} else {
+			// Not annotated, or annotated but proxy was cleared: none of our var
+			// names may be present (covers both "never had proxy" and "proxy removed").
+			for _, name := range proxyVarNames() {
+				if _, ok := envByName[name]; ok {
+					return false
+				}
+			}
+		}
+	}
```
mdbooth: I don't think this is quite right, because it asserts that unannotated containers do not have proxy vars. They could have proxy vars which aren't managed by us. I think [UnstructuredExtractor](https://pkg.go.dev/k8s.io/client-go@v0.36.2/applyconfigurations/meta/v1#UnstructuredExtractor) could be the answer here. Just be careful to create the Extractor 'globally' e.g. in SetupWithManager, because it has its own discovery cache.

That said, while my preference is to fix this, it should only result in 1 additional reconcile because the resulting SSA will still be correct and will not result in further reconciles. So if the fix starts looking too complex may just add a comment here explaining that it's not quite right and why the impact of that is contained.

pkg/controllers/installer/proxy_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+			By("Waiting for env vars to be cleared after annotation removal")
+			waitForContainer(ctx, "manager", notHaveProxyEnvVars())
+
+			By("Verifying the other container also has no proxy env vars")
+			waitForContainer(ctx, "other", notHaveProxyEnvVars())
+		}, defaultNodeTimeout)
+	})
+
+	Context("SSA field ownership correctness", func() {
```
mdbooth, line 247: Please can you add another test here: does not affect proxy vars set on un-annotated containers.

So: add some proxy vars to the `other` container with custom values unrelated to the cluster configuration. Annotate the deployment to manage the `manager` container. Proxy vars should be set on `manager` and left alone (i.e. not removed) on `other`.

mdbooth (commented):
I know I asked you to move the proxy controller into a separate package same as we've done for the other controller, but I'm questioning that. I was wondering why you didn't also move the tests: it's because you're using the installer controller to create managed resources for the tests.

This wasn't what I expected, but I think I like it better than what I expected: the proxy controller explicitly only ever operates on resources managed by the installer controller. Integrating those in the integration tests absolute makes sense (to me, but I know this is not a universally popular view).

How can we make that relationship more explicit in the structure? For crdcompatibility we put related controllers in sub-packages. Suggest we also do that here? I think I'd leave the tests where they are for now.

I've asked for consideration of another SSA thing (UnstructuredExtractor): either a fix or at least a comment. And either way some test coverage of the case.

mdbooth:
I've copied an additional clanker review below. I agree with most of that, although I wonder if 'Only uppercase proxy env vars' should be major.

 ### Blocking

 None. The feature works correctly on the primary happy paths (full proxy enable/disable/delete).

 ### Major Issues

 #### 1. proxyVarsAlreadyApplied() allows stale env vars on partial proxy reconfiguration

 https://github.com/openshift/cluster-capi-operator/blob/b269f09a1bae68269e132911e56dea251c83138b/pkg/controllers/proxy
 /proxy_controller.go#L365-L380

 When a container is annotated and proxyVars is non-empty, the function only checks that the desired vars are present
 and correct. It does not check that vars removed from proxyVars are absent.

 Scenario: Admin previously had HTTPProxy, HTTPSProxy, and NoProxy all configured (three env vars injected). Admin then
 clears only NoProxy from the Proxy status. GetProxyEnvVars() returns two vars. proxyVarsAlreadyApplied() sees both
 match and returns true, skipping the apply. Stale NO_PROXY remains on the container indefinitely.

 Fix: In the annotated+non-empty branch, additionally verify that no other proxy var names from proxyVarNames() are
 present unless they appear in proxyVars:

 ```go
   desiredNames := make(map[string]struct{}, len(proxyVars))
   for _, pv := range proxyVars {
       desiredNames[pv.Name] = struct{}{}
   }
   for _, name := range proxyVarNames() {
       if _, desired := desiredNames[name]; !desired {
           if _, present := envByName[name]; present {
               return false
           }
       }
   }
 ```

 #### 2. Test coverage gaps for important behaviour paths

 The 6 integration tests cover the core flows well but miss:

 - Proxy URL update (change value, not just create/delete) — regressions in value comparison go undetected
 - Partial proxy field removal (clear one of three) — would not catch Major #1 above
 - Multi-container annotation (manager,sidecar) — supported by containerNamesFromAnnotation but untested
 - DaemonSet / StatefulSet workloads — listed in collectWorkloads but only Deployment tested

 Additionally, the haveProxyEnvVars() matcher only checks env var names, not values:

 https://github.com/openshift/cluster-capi-operator/blob/b269f09a1bae68269e132911e56dea251c83138b/pkg/controllers/insta
 ller/proxy_controller_test.go#L100-L107

 A proxy URL update test would pass even if the values are wrong. Consider matching both name and value:

 ```go
   func haveProxyEnvVarsWithURL(url string) OmegaMatcher {
       return HaveField("Env", ContainElements(
           SatisfyAll(HaveField("Name", "HTTP_PROXY"), HaveField("Value", url)),
           SatisfyAll(HaveField("Name", "HTTPS_PROXY"), HaveField("Value", url)),
           // ...
       ))
   }
 ```

 #### 3. No unit tests for SSA helpers in pkg/controllers/proxy/

 buildPatch, proxyVarsAlreadyApplied, and containerNamesFromAnnotation are non-trivial pure functions with edge cases.
 They are only reachable through slow envtests in pkg/controllers/installer/. Table-driven unit tests in
 pkg/controllers/proxy/ would:
 - Make SSA semantics easier to review and maintain
 - Run much faster than envtests
 - Cover edge cases (empty annotations, whitespace-only values, ValueFrom env vars)

 ### Minor Issues

 #### 4. Only uppercase proxy env vars

 Some Go HTTP clients (and libraries like net/http's ProxyFromEnvironment) check lowercase http_proxy first. The
 OpenShift build API documents both forms. Uppercase-only probably works for current CAPI provider controllers, but
 injecting both would be more defensive. Worth a deliberate decision comment if staying uppercase-only.

 #### 5. EnvVar.ValueFrom not handled by skip check

 In proxyVarsAlreadyApplied, only Value is indexed. If another field manager sets HTTP_PROXY via ValueFrom (e.g. a
 SecretKeyRef), the skip check sees a mismatch and applies with ForceOwnership, replacing the ValueFrom. This is
 probably fine in practice (nobody uses ValueFrom for proxy URLs), but worth a comment noting the assumption.

 #### 6. Stale doc reference in installer_controller.go

 The updated SetupWithManager godoc references SetupProxyController, but the actual function is proxy.SetupWithManager.

 #### 7. manifests-gen domain constant duplication

 Duplicated because manifests-gen is a separate Go module. The comment explaining this is good. No great alternative
 exists without creating a tiny shared module, but a CI lint check that the strings match would reduce drift risk.

 ### Nits

 - proxyVarNames() allocates on every call — could be a package-level var. Cost is negligible but the allocation is
   unnecessary.
 - Info-level logging on skip may be noisy during installation — many managed objects trigger proxy reconciliation
   during initial install. Consider V(1) for the "skipping apply" log path.

tmshort:
Rebased. Ready for review.

pkg/controllers/installer/installer_controller.go
```diff
@@ -65,30 +68,50 @@ type InstallerController struct {
...
 // SetupWithManager creates the boxcutter dependencies and sets up the installer
-// controller with the Manager. Additional sources may be provided to trigger
-// reconciliation from external events (e.g. a channel source for testing).
-func SetupWithManager(mgr ctrl.Manager, providerProfiles []providerimages.ProviderImageManifests, additionalSources ...source.Source) error {
+// controller with the Manager. The proxy sub-reconciler is wired in here so
+// that both controllers share a single trackingCache.Source() registration,
+// avoiding the double-registration shutdown issue. Additional sources may be
+// provided to trigger reconciliation from external events (e.g. a channel
+// source for testing).
+func SetupWithManager(mgr ctrl.Manager, providerProfiles []providerimages.ProviderImageManifests, additionalSources ...source.Source) (managedcache.TrackingCache, error) {
```
mdbooth: nit: the TrackingCache return is no longer required. We should probably revert this.

pkg/controllers/installer/installer_controller.go
```diff
@@ -161,10 +191,10 @@ func setupTrackingCache(mgr ctrl.Manager) (managedcache.TrackingCache, error) {
 	return trackingCache, nil
 }
 
-func setupRevisionEngine(mgr ctrl.Manager, trackingCache managedcache.TrackingCache) (*boxcutter.RevisionEngine, error) {
+func setupRevisionEngine(mgr ctrl.Manager, trackingCache managedcache.TrackingCache) (*boxcutter.RevisionEngine, discovery.DiscoveryInterface, error) {
 	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
 	if err != nil {
-		return nil, fmt.Errorf("unable to create discovery client: %w", err)
+		return nil, nil, fmt.Errorf("unable to create discovery client: %w", err)
 	}
```
mdbooth, line 196: Re-using the discovery client is nice 👌

pkg/controllers/installer/installer_controller.go
```diff
@@ -195,6 +226,10 @@ func (c *InstallerController) Reconcile(ctx context.Context, _ ctrl.Request) (ct
 		return ctrl.Result{}, fmt.Errorf("failed to write conditions: %w", err)
 	}
 
+	if err := c.proxyReconciler.Reconcile(ctx); err != nil {
+		return ctrl.Result{}, fmt.Errorf("reconciling proxy env vars: %w", err)
+	}
+
 	log.Info("Reconcile finished")
```
mdbooth, line 234: I need to think about this, because the proxy reconciler needs to report its status somewhere. I don't mind it being integrated with the installer controller's status, but we need to think about how to do that.

mdbooth (commented):
/hold until we have a solution to the status reporting problem

pkg/controllers/installer/helpers_test.go
```diff
@@ -100,6 +104,41 @@ var (
...
+// optionally, the inject-proxy annotation on the pod template.
+func proxyDeploymentYAML(withAnnotation bool) string {
+	annotations := ""
+	if withAnnotation {
+		annotations = fmt.Sprintf(`
+      annotations:
+        %s: manager`, proxycontroller.ProxyInjectAnnotation)
+	}
+
+	return fmt.Sprintf(`apiVersion: apps/v1
```
nrb, line 119: What's the reason for using string templating for YAML rather than a builder and Go struct?

tmshort:
@nrb 
@mdbooth had intended to make some API changes that would've made this a much more trivial change (something about hooks to update the deployment). This work was not completed and postponed. So, my original approach remained.

nrb (commented):
This all looks reasonable to me. One question and I don't think it's blocking. Go ahead and remove the hold when it's answered.

/approve
/hold

pkg/controllers/installer/helpers_test.go
tmshort: It's following the convention of the other test utilities, specifically in https://github.com/openshift/cluster-capi-operator/blob/main/pkg/test/provider_fixtures.go, which return YAML strings.

tmshort:
/unhold
@nrb's question answered
Still needs LGTM and verified.

mdbooth:
Just discussed this with Todd. The direction I want to go is that this should follow the same pattern as https://github.com/openshift/cluster-capi-operator/pull/648, namely the transformation should happen in the installer controller when toBoxcutterRevision is called. I don't mind which merges first, this or #648, but they're definitely going to conflict so it's best to coordinate with @theobarberbany.

The validation you're already doing on manifests should be retained, but instead of separate validation just return a user-friendly error from toBoxcutterRevision. Note that this function can't currently return an error, but #648 also adds that ability. Whether you add it or @theobarberbany does is up to you. We'll 'validate' by calling toBoxcutterRevision from the revision controller. It will discard the output, but refuse to create a revision if it can't be transformed.

As discussed, we'll also need to add the relevant watches to the installer controller.

PR #587, opened 2026-06-10
NO-JIRA: Refactor AGENTS.md to be minimal with references

mdbooth:
Also folds the test-standards skill into a reference.

This change is a slight optimisation to context usage, as well as making the test-standards reference available to agents other than Claude.

## Summary by CodeRabbit

* **Documentation**
  * Reorganized internal developer documentation with consolidated reference guides for project structure, style conventions, build tasks, and testing practices.

**Note:** This release contains internal documentation updates with no changes to user-facing features or functionality.


theobarberbany:
this is posted by claude, reviewed by me 🎉 

I've reviewed this and the instinct is right — the old AGENTS.md was 184 lines, already over the [recommended ≤150 line guideline](https://agents.md/). ETH Zurich research found that bloated context files hurt agent performance (20%+ increased inference costs with no accuracy benefit). A lean root file is the right shape.

I have a few concerns with the approach though. I've left inline comments on the specific areas.

.agents/reference/code-structure.md
```diff
@@ -0,0 +1,56 @@
+# Code Structure
```
theobarberbany, line 1: **\`.agents/\` is not a standard directory — consider hierarchical \`AGENTS.md\` files instead**

The established convention for agent instructions is [hierarchical \`AGENTS.md\` files](https://agents.md/) placed in subdirectories. Agents auto-read the nearest \`AGENTS.md\` in the directory tree — the closest one takes precedence, and each subdirectory can ship tailored instructions. The OpenAI repo has 88 of them. This is supported natively by Claude Code, Codex, Cursor, Copilot, Jules, and others, and is now governed by the [Agentic AI Foundation](https://www.infoq.com/news/2025/08/agents-md/) under the Linux Foundation.

There is a [community proposal](https://github.com/agentsmd/agents.md/issues/71) for a \`.agent\` directory (singular, different structure), but it's at the discussion stage and not adopted by any tool. The \`.agents/reference/\` directory used here doesn't match the standard, doesn't match the proposal, and won't be auto-discovered by any agent tooling.

These files will only be read if an agent actively decides to follow the links in the root AGENTS.md — which is not guaranteed, especially for quick fixes or tasks where the agent doesn't think to check conventions.

**Suggestion:** Place this content in \`AGENTS.md\` files in relevant subdirectories instead (e.g., \`pkg/AGENTS.md\`, \`e2e/AGENTS.md\`). This gets the same "thin root, detailed when needed" outcome while using the standard mechanism that tools already auto-discover.

.claude/skills/test-standards/SKILL.md
```diff
@@ -1,121 +0,0 @@
----
```
theobarberbany, line 1: **Removing this skill loses the auto-trigger on \`_test.go\` files**

This skill had an auto-trigger: it fired automatically when agents worked on \`_test.go\` files, injecting testing conventions into context at exactly the right time. Moving the content to \`.agents/reference/testing.md\` and deleting the skill removes that trigger.

Now an agent writing or modifying tests will only see the testing guidance if it independently decides to read the reference file — which isn't guaranteed, especially for quick test fixes.

The \`deep-review\` skill was updated to point at the new location, so reviews are covered. But unguided work (writing new tests, fixing test failures) loses the automatic context injection.

**Suggestion:** Either keep this skill (pointing at the new reference file location), or ensure the testing guidance ends up in a subdirectory \`AGENTS.md\` that tools auto-discover.

AGENTS.md
```diff
@@ -2,183 +2,21 @@
...
-# Build and test
-make build          # Build all binaries
-make test           # Run all tests
-make unit           # Run unit tests with coverage
-make lint           # Run linting
-make fmt            # Format code
-
 ## Project Overview
 
 The Cluster CAPI Operator manages the installation and lifecycle of Cluster API components on OpenShift clusters. It serves as a bridge between OpenShift's Machine API (MAPI) and the upstream Cluster API (CAPI), providing forward compatibility and migration capabilities.
```
theobarberbany, line 8: **These links won't be reliably followed by agents**

The root AGENTS.md is auto-loaded into context by Claude Code, Codex, Cursor, Copilot, and others. But these linked files are regular markdown files — an agent has to actively decide to read them. That decision depends on the agent judging that the reference is relevant to the current task, which won't happen consistently.

For example: with the old inline AGENTS.md, "never use \`time.Sleep()\`, use \`Eventually\`" was always in context during test work. Now it's behind a link the agent may or may not follow. "Always applied" vs "usually applied" is where regressions show up.

The standard approach here is subdirectory \`AGENTS.md\` files (see comment on \`.agents/reference/\`), which tools auto-discover without the agent needing to make a judgment call.

PR #589, opened 2026-06-10
OCPBUGS-88035: Set cluster ownership tag on AWSCluster

matzew:
(damdo) Closing in favour of: https://github.com/openshift/cluster-capi-operator/pull/594

--

Set the kubernetes.io/cluster/<infraID>=owned tag on the AWSCluster object via AdditionalTags so that CAPI-created AWS resources are visible to the installer's destroy logic. Also propagate user-defined resource tags from Infrastructure.Status.PlatformStatus.AWS.ResourceTags to match MAPI behaviour.

Without this tag, openshift-install destroy cluster cannot identify CAPI-created resources (VMs, volumes, etc.) and they are leaked.

### DIFF
--- a/pkg/controllers/infracluster/aws.go
+++ b/pkg/controllers/infracluster/aws.go
@@ -102,6 +102,16 @@ func (r *InfraClusterController) newAWSCluster(providerSpec *mapiv1beta1.AWSMach
 		return nil, fmt.Errorf("failed to extract control plane load balancer configuration: %w", err)
 	}
 
+	tags := awsv1.Tags{}
+
+	if r.Infra.Status.PlatformStatus.AWS != nil {
+		for _, t := range r.Infra.Status.PlatformStatus.AWS.ResourceTags {
+			tags[t.Key] = t.Value
+		}
+	}
+
+	tags[fmt.Sprintf("kubernetes.io/cluster/%s", r.Infra.Status.InfrastructureName)] = "owned"
+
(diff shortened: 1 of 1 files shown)


mdbooth:
/hold

Discussed this with @matzew . The missing (and unfortunately more difficult) piece here is the testing. I had it in mind that we could write something that would trigger in `aws-deprovision-verification` if it was not cleaned up, but this presents a chicken and egg problem: the deprovision verification step will not find an incorrectly tagged resource so it won't fail. I think the pragmatic solution here is to add an explicit e2e test that the created ec2 VM has the same tag the installer is looking for during destroy.

Incidentally, it's possible that the tag is already added implicitly by CAPA. It looks like CAPA adds a tag `kubernetes.io/cluster/<KubernetesClusterName> = owned`. `KubernetesClusterName` is the name of the core capi cluster which owns the machine, that the name of the core capi cluster should match the infrastructure name. Again, a test would verify this.

Therefore I think we should write the described test *without* adding the tag explicitly. Ideally the test would simply be an additional check in an existing test. Any test which already creates an AWSMachine should be sufficient.

If this passes, we're done. If not, we need to add the tag explicitly as done here.

mdbooth:
test: https://github.com/openshift/cluster-capi-operator/pull/594

damdo:
Closing in favour of: https://github.com/openshift/cluster-capi-operator/pull/594

PR #591, opened 2026-06-10
OCPBUGS-88037: Set cluster ownership label on GCPCluster

matzew:
(damdo) Closing in favour of: https://github.com/openshift/cluster-capi-operator/pull/594

-- 

Set the kubernetes-io-cluster-<infraID>=owned label on the GCPCluster object via AdditionalLabels so that CAPI-created GCP resources are visible to the installer's destroy logic. GCP label keys only permit lowercase letters, hyphens, and underscores. Also propagate user-defined resource labels from Infrastructure.Status.PlatformStatus.GCP.ResourceLabels and resource manager tags from GCP.ResourceTags to match MAPI behaviour.

Without this label, openshift-install destroy cluster cannot identify CAPI-created resources and they are leaked.


damdo:
Closing in favour of: https://github.com/openshift/cluster-capi-operator/pull/594

PR #592, opened 2026-06-10
CORS-4512: aws: allow privateDNSName and assignPrimaryIPv6 field on CAPI machines in dual-stack clusters

tthvo:
## Description

Remove `privateDnsName` from the unsupported AWS spec fields admission policy and CAPI-to-MAPI conversion validation. This unblocks dualstack installs that require configuring instance hostname DNS records (AAAA/A) via the CAPA PrivateDNSName field. Additionally, `assignPrimaryIPv6` is also handled the same way.

**Note**: MAPI has no equivalent fields and the infra CR already holds these information via ipFamily, so it is silently dropped during conversion.

See also https://github.com/openshift/cluster-capi-operator/pull/593

## Summary by CodeRabbit

* **Bug Fixes**
  * Relaxed admission restrictions by no longer rejecting `privateDnsName` in AWS machine policy checks.
  * Added admission enforcement to deny requests when ignition `proxy` or `tls` settings are present.
* **New Features**
  * Improved AWS machine conversion to support dual-stack configurations by populating `privateDnsName` and `assignPrimaryIPv6` based on the infrastructure IP family.
* **Tests**
  * Updated and expanded conversion tests to reflect the new dual-stack and validation behavior.

### DIFF
--- a/pkg/conversion/capi2mapi/aws.go
+++ b/pkg/conversion/capi2mapi/aws.go
@@ -788,10 +788,11 @@ func handleUnsupportedAWSMachineFields(fldPath *field.Path, spec awsv1.AWSMachin
 		errs = append(errs, field.Invalid(fldPath.Child("cloudInit"), spec.CloudInit, "cloudInit is not supported"))
 	}
 
-	if spec.PrivateDNSName != nil {
-		// Not required for our use case.
-		errs = append(errs, field.Invalid(fldPath.Child("privateDNSName"), spec.PrivateDNSName, "privateDNSName is not supported"))
-	}
+	// privateDNSName is not checked here because the relevant information is available on the
+	// infrastructure CR, so nothing additional needs to be added to the MAPI machine spec.
+
+	// assignPrimaryIPv6 is not checked here because the relevant information is available on the
+	// infrastructure CR, so nothing additional needs to be added to the MAPI machine spec.
 
 	if spec.Ignition != nil {
 		if spec.Ignition.Proxy != nil {
--- a/pkg/conversion/mapi2capi/aws.go
+++ b/pkg/conversion/mapi2capi/aws.go
@@ -284,6 +284,8 @@ func (m *awsMachineAndInfra) toAWSMachine(providerSpec mapiv1beta1.AWSMachinePro
 		spec.CapacityReservationID = &providerSpec.CapacityReservationID
 	}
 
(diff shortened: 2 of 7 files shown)


pkg/conversion/capi2mapi/aws.go
```diff
@@ -788,11 +788,6 @@ func handleUnsupportedAWSMachineFields(fldPath *field.Path, spec awsv1.AWSMachin
 		errs = append(errs, field.Invalid(fldPath.Child("cloudInit"), spec.CloudInit, "cloudInit is not supported"))
 	}
 
-	if spec.PrivateDNSName != nil {
-		// Not required for our use case.
-		errs = append(errs, field.Invalid(fldPath.Child("privateDNSName"), spec.PrivateDNSName, "privateDNSName is not supported"))
-	}
```
nrb, line 794: Bringing the discussion here from Slack, I think we should update the VAPs to allow CAPI resources to be created with the field, since it's required for installation. This check, however, should stay in order to reinforce the fact that dual stack clusters are a CAPI-only feature, and thus aren't convertible.

tthvo:
/retitle: [CORS-4512](https://redhat.atlassian.net/browse/CORS-4512): aws: allow privateDnsName field on AWS CAPI machines for dualstack

pkg/conversion/capi2mapi/aws.go
tthvo: Following up (from slack), the direction I went with:

- MAPI -> CAPI:  if dual-stack, read the infrastructure CR to populate the `privateDNSName` and `assignPrimaryIPv6` for CAPI machines.
- CAPI -> MAPI: The info is already on the infrastructure CR so nothing to convert.

pkg/conversion/mapi2capi/aws.go
```diff
@@ -284,6 +284,8 @@ func (m *awsMachineAndInfra) toAWSMachine(providerSpec mapiv1beta1.AWSMachinePro
 		spec.CapacityReservationID = &providerSpec.CapacityReservationID
 	}
 
+	convertAWSDualStackFieldsToCAPI(m.infrastructure, &spec)
```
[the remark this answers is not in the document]
tthvo: It's ok. We access the infrastructure elsewhere in the same way. Check line 301

PR #597, opened 2026-06-16
OCPCLOUD-3009: Bootstrap OTE framework

pmeida:
## Summary

- Introduces the [OpenShift Tests Extension](https://docs.google.com/document/d/1cFZj9QdzW8hbHc3H0Nce-2xrJMtpDJrwAse9H7hLiWk/edit?tab=t.0) (OTE) binary scaffold registering the component as `openshift:payload:cluster-capi-operator`
- Adds `openshift-tests-extension/` as a **separate Go module** to keep OTE dependencies (ginkgo fork, test framework) isolated from the main module's dependency graph, following the [cluster-control-plane-machine-set-operator](https://github.com/openshift/cluster-control-plane-machine-set-operator) convention
- 3 test suites configured with `Parallelism` and `ClusterStability` execution parameters and wired to the standard conformance hierarchy via `Parents`
- Embeds the gzipped binary in the operator image at `/usr/bin/cluster-capi-operator-tests-ext.gz`
- Wires up ginkgo integration so that any future test written as a regular `.go` file (not `_test.go`) is automatically discovered by `openshift-tests` with no changes to the extension
- `e2e/machine_migration_vap_tests.go` → `e2e/machine_migration_vap_test.go` - corrects an accidental naming error that would have caused those tests to be picked up by OTE unintentionally

Similar approaches:
- https://github.com/openshift/cluster-etcd-operator/tree/main
- https://github.com/openshift/cluster-cloud-controller-manager-operator/tree/main

## Test Suites

| Suite | Qualifier | Parent | Parallelism | Timeout |
|---|---|---|---|---|
| `capio/parallel` | not Serial, not Disruptive | `openshift/conformance/parallel` | - | 30m |
| `capio/serial` | Serial, not Disruptive | `openshift/conformance/serial` | 1 | 30m |
| `capio/disruptive` | Disruptive | `openshift/disruptive-longrunning` | 1 | 90m |

The `Parents` field is what causes tests to be automatically included in the nightly conformance runs and feed Sippy for promotion gate evaluation.

## Test plan

- [x] `make bin/cluster-capi-operator-tests-ext` builds successfully
- [x] `bin/cluster-capi-operator-tests-ext info` returns correct component metadata (`openshift:payload:cluster-capi-operator`)
- [x] `bin/cluster-capi-operator-tests-ext list tests` returns empty (zero tests registered, as intended)
- [x] All existing unit tests pass (`make unit`)
- [x] `make vendor` runs cleanly
- [x] check if `e2e presubmit test` are not affected by the change
- [x] use `/payload-job-with-prs` to check if the binary is detected (and with 0 tests)
- [x] test if a test without `_test.go` runs with `make e2e`

## Local outputs

```bash
# for up to date results the binary has to be rebuild every time you change tests
make bin/cluster-capi-operator-tests-ext
...
bin/cluster-capi-operator-tests-ext info
{
    "apiVersion": "v1.1",
    "source": {
        "commit": "",
        "build_date": "",
        "git_tree_state": ""
    },
    "component": {
        "product": "openshift",
        "type": "payload",
        "name": "cluster-capi-operator"
    },
    "suites": [
        {
            "name": "capio/parallel",
            "description": "",
            "parents": [
                "openshift/conformance/parallel"
            ],
            "qualifiers": [
                "(source == \"openshift:payload:cluster-capi-operator\") \u0026\u0026 (!labels.exists(l, l == \"Serial\") \u0026\u0026 !labels.exists(l, l == \"Disruptive\"))"
            ]
        },
...
```
No tests are yet configured with OTE (that is for future tasks)
```bash
bin/cluster-capi-operator-tests-ext list tests
null
```

Validation with PR #599 tests (renaming `crd_compatibility_test.go` → `crd_compatibility.go`):
```bash
bin/cluster-capi-operator-tests-ext list tests | python3 -c "import sys,json; d=json.load(sys.stdin); print(f'{len(d)} tests discovered')"
10 tests discovered
```

## Next steps

- [OCPCLOUD-3542](https://redhat.atlassian.net/browse/OCPCLOUD-3542): Implement actual e2e tests using the `et.ExtensionTestSpec` API

blocks https://github.com/openshift/origin/pull/31307

## Summary by CodeRabbit

* **New Features**
  * Introduced OpenShift Tests Extension (OTE) for extended e2e test integration and execution

* **Documentation**
  * Added comprehensive guide for OTE setup, test discovery, configuration requirements, and integration with nightly test runs and feature gates


openshift-tests-extension/go.mod
```diff
@@ -0,0 +1,22 @@
...
+	github.com/google/cel-go v0.27.0 // indirect
+	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/spf13/pflag v1.0.10 // indirect
+	github.com/stretchr/testify v1.11.1 // indirect
+	golang.org/x/exp v0.0.0-20260112195511-716be5621a96 // indirect
+	golang.org/x/text v0.35.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20260209200024-4cfbd4190f57 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20260209200024-4cfbd4190f57 // indirect
+	google.golang.org/protobuf v1.36.11 // indirect
+)
```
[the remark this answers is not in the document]
pmeida: These are all packages that don't publish tagged releases. Hash pinning via go.sum is in place.

pmeida:
## How to add a test to OTE

The extension uses ginkgo auto-discovery. Whether a test is visible to OTE depends entirely on the file suffix:

| File | `make e2e` | OTE (`list tests`) |
|---|---|---|
| `e2e/my_test.go` | ✅ | ❌ |
| `e2e/my.go` | ✅ | ✅ |
| `e2e/my.go` + `//go:build !e2e` | ❌ | ✅ |

**To add a test to OTE:** write it in a regular `.go` file (not `_test.go`). No changes to the extension are needed — it is picked up automatically on the next build.

**To keep a test out of OTE:** use the standard `_test.go` suffix. It still runs via `make e2e`.

This was validated locally using the tests from #599 — renaming `crd_compatibility_test.go` → `crd_compatibility.go` caused all 10 CRD Compatibility Checker tests to appear in `list tests` with no other changes.

openshift-tests-extension/cmd/main.go
```diff
@@ -0,0 +1,90 @@
...
+
+	ext.AddSuite(e.Suite{
+		Name:       "capio/periodic",
+		Qualifiers: []string{`labels.exists(l, l == "Periodic")`},
+	})
+
+	ext.AddSuite(e.Suite{
+		Name:        "capio/e2e",
+		Description: "All Cluster CAPI Operator tests",
+	})
```
damdo: Let's chat about these.
What does OTE enforce WRT to what suites we define?
Do we want to define all these?

pmeida: OTE doesn't enforce any particular suite structure - suites are optional metadata. The two that actually matter for promotion are `capio/parallel` and `capio/serial` via their Parents field, which is how tests get included in the nightly conformance runs and feed Sippy for promotion gate evaluation.

The `presubmit/periodic` suites were added for future-use, but they are not used currently. Happy to drop them if you'd prefer to start minimal and add as needed. The idea behind those 2 was to do explicit label-based filtering, controlling which tests run in presubmit/periodics. For this to be possible we would have to change the jobs configuration to something like:

```yaml
test:
- ref: openshift-e2e-test
  env:
  - name: TEST_SUITE
    value: capio/presubmit
```

pmeida: On reflection, they're harder to use in presubmits than they appear - a job using `openshift-e2e-test` with `TEST_SUITE: capio/presubmit` would pull the binary from the ART/stable image, not the PR's freshly-built one, so new tests in a PR wouldn't be picked up. For presubmit validation of new tests `make e2e` from src remains the right approach.

Lets drop those.

pmeida: Kept only `capio/parallel`, `capio/serial`, and `capio/disruptive` — these are the only suites that map to suites actually run in nightly builds via their Parents field (`openshift/conformance/parallel`, `openshift/conformance/serial`, `openshift/disruptive-longrunning`).

If we want to increase testing frequency with an explicit periodic job in the future, we can reference the parent or the child suites individually in the job.

pmeida:
## Local OTE flow walk-through

To answer the question about local invocation - here's the full flow:

> **Note:** `run-suite` and `run-test` require a reachable OCP cluster because `InitCommonVariables()` fetches the `Infrastructure` object. `list tests` and `list suites` work without one.

---

### Added a demo file used (`e2e/demo.go` - regular `.go`, not `_test.go`, so it is discoverable by OTE)

```go
package e2e

import (
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
	g "github.com/openshift-eng/openshift-tests-extension/pkg/ginkgo"
)

var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManagementAWS] Cluster API OTE demo", func() {
	It("should be discoverable by OTE as a parallel test", func() {
		By("This test has no labels, so it lands in capio/parallel")
	})

	It("should be a serial test", Label("Serial"), func() {
		By("This test has Label(Serial), so it lands in capio/serial")
	})

	It("should be a disruptive test", Label("Disruptive"), func() {
		By("This test has Label(Disruptive), so it lands in capio/disruptive")
	})

	It("should be an informing test that fails", g.Informing(), func() {
		By("This test is informing - its failure should not fail the suite")
		Expect("actual").To(Equal("expected"))
	})
})
```

---

### `capio/parallel` - includes the 2 unlabeled tests

```bash
# using openshift/conformance/parallel would result in the exact same (as it is the parent)
~ ./bin/cluster-capi-operator-tests-ext run-suite capio/parallel
[
  {
    "name": "[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManagementAWS] Cluster API OTE demo should be an informing test that fails",
    "lifecycle": "informing",
    "duration": 37,
    "startTime": "2026-06-17 10:31:00.994642 UTC",
    "endTime": "2026-06-17 10:31:00.995154 UTC",
    "result": "failed",
    "output": "  STEP: This test is informing - its failure should not fail the suite @ 06/17/26 11:31:00.994\n  [FAILED] in [It] - /Users/pealmeid/Projects/cluster-capi-operator/e2e/demo.go:43 @ 06/17/26 11:31:00.995\n",
    "error": "fail [/Users/pealmeid/Projects/cluster-capi-operator/e2e/demo.go:43]: Expected\n    \u003cstring\u003e: actual\nto equal\n    \u003cstring\u003e: expected"
  },
  {
    "name": "[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManagementAWS] Cluster API OTE demo should be discoverable by OTE as a parallel test",
    "lifecycle": "blocking",
    "duration": 38,
    "startTime": "2026-06-17 10:31:00.997223 UTC",
    "endTime": "2026-06-17 10:31:00.997424 UTC",
    "result": "passed",
    "output": "  STEP: This test has no labels, so it lands in capio/parallel @ 06/17/26 11:31:00.997\n"
  }
]
```

The informing test failure is **visible in the JSON output** but the suite exits with **code 0** - it does not gate anything. This is consistent with how `openshift-tests` treats informing failures: surfaced as a signal in Sippy/dashboards, never a blocker.
Note: `make e2e` uses plain ginkgo which has no concept of "informing"; it just sees a failing spec and fails the run. The informing lifecycle is only meaningful when the binary is invoked through OTE (`run-suite`).

---

### `capio/serial` - only picks up tests with `Label("Serial")`

```bash
# using openshift/conformance/serial would result in the exact same (as it is the parent)
~ ./bin/cluster-capi-operator-tests-ext run-suite capio/serial
[
  {
    "name": "[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManagementAWS] Cluster API OTE demo should be a serial test",
    "lifecycle": "blocking",
    "duration": 0,
    "startTime": "2026-06-17 10:24:42.741528 UTC",
    "endTime": "2026-06-17 10:24:42.741754 UTC",
    "result": "passed",
    "output": "  STEP: This test has Label(Serial), so it lands in capio/serial @ 06/17/26 11:24:42.741\n"
  }
]
```

Suite qualifiers handle the filtering - no need to pass labels manually.

---

### `run-test` by exact name - runs one test regardless of suite

```bash
~ ./bin/cluster-capi-operator-tests-ext run-test "[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManagementAWS] Cluster API OTE demo should be a disruptive test"
[
  {
    "name": "[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManagementAWS] Cluster API OTE demo should be a disruptive test",
    "lifecycle": "blocking",
    "duration": 0,
    "startTime": "2026-06-17 10:27:05.302156 UTC",
    "endTime": "2026-06-17 10:27:05.302426 UTC",
    "result": "passed",
    "output": "  STEP: This test has Label(Disruptive), so it lands in capio/disruptive @ 06/17/26 11:27:05.302\n"
  }
]
```

openshift-tests-extension/go.mod
```diff
@@ -0,0 +1,153 @@
...
+	github.com/openshift-eng/openshift-tests-extension v0.0.0-20260612102633-2fd5b2fa4221
+	github.com/openshift/cluster-capi-operator/e2e v0.0.0-00010101000000-000000000000
+	github.com/spf13/cobra v1.10.2
+)
+
+replace (
+	github.com/onsi/ginkgo/v2 => github.com/openshift/onsi-ginkgo/v2 v2.6.1-0.20241205171354-8006f302fd12
+	github.com/openshift/cluster-api-actuator-pkg/testutils => github.com/openshift/cluster-api-actuator-pkg/testutils v0.0.0-20260310144400-bec013a007a8
+	github.com/openshift/cluster-capi-operator/e2e => ../e2e
+)
```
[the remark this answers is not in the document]
pmeida: Neither finding passes the "verify and fix only still-valid issues with minimal changes" bar here — one is a behavior change requiring re-vendoring, the other is a false positive for this build model. No changes made.

pmeida:
## End-to-end binary verification (temporary test — now reverted)

This PR adds/moves no test to OTE scope, so to confirm the OTE binary is correctly embedded in the image **and** discovered by `openshift-tests`, a temporary no-op test was added (commit `67716b2e`, now removed via force-push) with the following content:

```go
// e2e/ote_binary_verification.go
package e2e

import . "github.com/onsi/ginkgo/v2"

var _ = Describe("[sig-cluster-lifecycle] Cluster CAPI Operator OTE binary", func() {
	It("should be discoverable by openshift-tests", Label("Serial"), func() {})
})
```

No `[OCPFeatureGate:]` filter so it runs on any cluster. `Label("Serial")` so it lands in `capio/serial → openshift/conformance/serial`.

Two runs were used for verification:

**1. Presubmit `e2e-aws-capi-techpreview` ([run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/597/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/2067269562750996480))**
The test appeared and passed - confirming that regular `.go` files (non `_test.go`) are picked up by `make e2e` as expected.

**2. Payload job with openshift/cluster-capi-operator#597 + openshift/origin#31307 ([run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-cluster-capi-operator-597-openshift-origin-31307-nightly-5.0-e2e-aws-ovn-single-node-techpreview-serial/2067241748672811008))**
Triggered via `/payload-job-with-prs` substituting both PRs into a `nightly-5.0-e2e-aws-ovn-single-node-techpreview-serial` payload. This exercises the full OTE path: `openshift-tests` extracts the binary from the `cluster-capi-operator` image, invokes `list tests`, and runs matching tests via `capio/serial`.

openshift-tests-extension/cmd/main.go
```diff
@@ -0,0 +1,85 @@
...
+	e2e "github.com/openshift/cluster-capi-operator/e2e"
+)
+
+func main() {
+	extensionRegistry := e.NewRegistry()
+
+	ext := e.NewExtension("openshift", "payload", "cluster-capi-operator")
+
+	defaultTimeout := 30 * time.Minute
+	disruptiveTimeout := 90 * time.Minute
```
damdo, line 38: Are these mandatory? Is there a default value if not set? Is the timeout per test, per suite or per overall parent (including the new suite)

ATM we have overall suite tests in the "local" e2es, but I guess those won't apply for these as the tests are cherry-picked via filters and ran in a different suite of its own?

AGENTS.md
```diff
@@ -182,3 +182,46 @@ FContext("context name", func() { /* tests */ })
...
+
+### Feature gate labeling
+
+Every e2e test must carry the appropriate `[OCPFeatureGate:FeatureName]` tag in its `Describe` block name. This tells `openshift-tests` to skip the test automatically on clusters where the feature gate is not enabled:
+
+```go
+var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManagementAWS] Cluster API AWS", func() {
+    It("should run a machine", func() { ... })
+})
+```
```
damdo: Would it be possible to put this in the README.md of the OTE module itself and then reference it here?

damdo (commented):
Thanks a lot Pedro, for the detailed investigation.
Here is a first pass review

openshift-tests-extension/cmd/main.go
pmeida: Good questions.

> Are these mandatory? Is there a default value if not set?

No. TestTimeout is an optional field on Suite. If omitted, OTE defaults to [90 minutes](https://github.com/openshift-eng/openshift-tests-extension/blob/main/pkg/extension/extensiontests/types.go#L63) per test.

> Per test, per suite, or per overall parent? 

Per individual test spec. OTE spawns each test as its own child process via SpawnProcessToRunTest - so the timeout controls how long that single process lives before being killed. So the `30m` is intentional, it's a tighter bound suited to what CAPI tests actually do. The disruptive `90m` makes sense too since disruptive scenarios can involve cluster restarts or longer reconciliation loops.

On your last point - yes, you're right. The make e2e ginkgo `--timeout=120m` is a suite-level timeout controlling the entire local run as a single process. It has no bearing on OTE, because OTE doesn't call ginkgo directly - it cherry-picks individual tests and spawns each one in isolation. The `120m` timeout never enters the picture.

I have seen implementations that rely on the defaults and others that overwrite it like we did. This has probably to be adapted with time and learnings.

AGENTS.md
pmeida: make sense. done 👍🏼

openshift-tests-extension/cmd/main.go
damdo: Ok, SGTM then, they'll probably need adjusting overtime, for now it is ok, TY

damdo:
@pmeida Looks like something merged, we'll need rebasing here

damdo (commented):
/approve

vr4manta:
LGTM

damdo:
/verified by @pmeida


 https://github.com/openshift/cluster-capi-operator/pull/597#issuecomment-4734553355

pmeida:
had to rebase onto latest upstream main (post PR [#598](https://github.com/openshift/cluster-capi-operator/pull/598)). The ovn presubmit builds were failing because openshift-tests-extension/go.mod still pinned cluster-api-actuator-pkg/testutils to the old version while root and e2e go.mod had already been bumped by [#598](https://github.com/openshift/cluster-capi-operator/pull/598). With go.work present, the Docker build runs in workspace mode and requires all modules to agree on replace directives - the conflict caused go build to abort before compiling anything. Aligned all three modules and re-vendored.

some presubmits went through some didnt, just bad timing

PR #599, opened 2026-06-16
OCPCLOUD-3542: Add CRD Compatibility Checker OTE e2e tests

stefanonardo:
## Summary
Add 10 e2e tests covering all three webhook paths of the CRD Compatibility Checker: CRD schema validation, object validation, and object pruning

## Summary by CodeRabbit

* **Tests**
  * Added a Ginkgo/Gomega end-to-end suite for CRD compatibility behavior, gated by the feature setting.
  * Validates controller Deployment health, required validating webhooks, and expected replica/topology alignment.
  * Covers CRD admission and update compatibility (compatible updates succeed; incompatible changes are eventually rejected with compatibility-related errors).
  * Adds object schema validation coverage for deny behavior, including missing required-field rejection messaging.
  * Extended the shared e2e framework with create-and-cleanup and delete-and-wait helpers, and expanded the test scheme with OpenShift API types.


e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+
+	Eventually(func() error {
+		err := cl.Get(ctx, client.ObjectKey{Name: compatibilityRequirementCRDName}, crd)
+		if apierrors.IsNotFound(err) {
+			Skip("CompatibilityRequirement CRD is not installed (feature gate CRDCompatibilityRequirementOperator not enabled). Skipping CRD compatibility tests.")
+		}
+
+		return err
+	}).Should(Succeed(), "failed to check for CompatibilityRequirement CRD")
+}
```
mdbooth: I feel like this should gate on the feature gate rather than the CRD. Can anybody more familiar with such things confirm? Perhaps @simkam or @JoelSpeed ?

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+		}
+
+		return err
+	}).WithTimeout(framework.WaitShort).Should(Succeed(), "failed to delete test CRD")
+
+	Eventually(func() bool {
+		err := cl.Get(ctx, key, &apiextensionsv1.CustomResourceDefinition{})
+		return apierrors.IsNotFound(err)
+	}).WithTimeout(framework.WaitMedium).WithPolling(framework.RetryMedium).Should(BeTrue(), "test CRD was not fully deleted")
+}
```
mdbooth: I think we already have a generic cleanupResource helper (or similar) which would cover both of these cases? Or perhaps that's only in the integration tests...

Can you look harder for one, anyway? If it doesn't exist we should write it. It's identical to both of these except it takes a `client.Object` instead of a specific type. You can use an Unstructured for the get if you don't want to get fancy.

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+		fresh := &apiextensionsv1alpha1.CompatibilityRequirement{}
+		g.Expect(cl.Get(ctx, client.ObjectKeyFromObject(requirement), fresh)).To(Succeed())
+
+		g.Expect(fresh.Status.Conditions).To(SatisfyAll(
+			test.HaveCondition(apiextensionsv1alpha1.CompatibilityRequirementAdmitted).
+				WithStatus(metav1.ConditionTrue),
+			test.HaveCondition(apiextensionsv1alpha1.CompatibilityRequirementProgressing).
+				WithStatus(metav1.ConditionFalse).
+				WithReason(apiextensionsv1alpha1.CompatibilityRequirementUpToDateReason),
+		))
```
mdbooth: We also need to check that status.ObservedCRD has the expected UID and Generation to avoid races.

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+	trackResource(crd)
+
+	By("Creating the CompatibilityRequirement")
+
+	Expect(cl.Create(ctx, requirement)).To(Succeed())
+	trackResource(requirement)
+
+	DeferCleanup(func() {
+		cleanupCompatibilityRequirement(requirement)
+		cleanupTestCRD(crd)
```
mdbooth: Ideally you want to put the DeferCleanup immediately after the create (so 2 DeferCleanups here). Otherwise a failure of the second Create means the first object will not be cleaned up.

Although, is there a `createTestResource` helper which both creates the object and also schedules its cleanup?

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+
+			for _, pod := range pods.Items {
+				Expect(pod.Status.Phase).To(Equal(corev1.PodRunning), fmt.Sprintf("pod %s is not Running", pod.Name))
+			}
+
+			By("Checking the CRD validation ValidatingWebhookConfiguration exists")
+			Expect(cl.Get(ctx, client.ObjectKey{Name: staticVWCCRDValidation}, &admissionregistrationv1.ValidatingWebhookConfiguration{})).To(Succeed())
+
+			By("Checking the CompatibilityRequirement validation ValidatingWebhookConfiguration exists")
+			Expect(cl.Get(ctx, client.ObjectKey{Name: staticVWCCRValidation}, &admissionregistrationv1.ValidatingWebhookConfiguration{})).To(Succeed())
```
mdbooth: IIRC these are created dynamically by the controller itself, so we should put these in an Eventually for robustness.

Also, top Ginkgo tip: the logs are better structured if you pass the actions associated with By to the By call as a lambda, e.g.:

```go
			By("Checking the CompatibilityRequirement validation ValidatingWebhookConfiguration exists", func() {
			  Eventually(cl.Get(ctx, client.ObjectKey{Name: staticVWCCRValidation}, &admissionregistrationv1.ValidatingWebhookConfiguration{})).Should(Succeed())
            })
```

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+			}).WithTimeout(framework.WaitLong).WithPolling(framework.RetryMedium).Should(Succeed())
+
+			By("Checking all pods are Running")
+
+			pods := &corev1.PodList{}
+			Expect(cl.List(ctx, pods,
+				client.InNamespace(framework.CompatibilityRequirementsNamespace),
+				client.MatchingLabels{"k8s-app": crdCompatibilityDeploymentName},
+			)).To(Succeed())
+			Expect(pods.Items).ToNot(BeEmpty(), "expected at least one pod")
```
mdbooth: Eventually

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+			Expect(cl.Get(ctx, client.ObjectKey{Name: staticVWCCRValidation}, &admissionregistrationv1.ValidatingWebhookConfiguration{})).To(Succeed())
+		})
+	})
+
+	Context("CRD schema validation with Deny action", func() {
+		It("should reject incompatible CRD updates", func() {
+			testCRD := test.GenerateTestCRD()
+			requirement := test.GenerateTestCompatibilityRequirement(testCRD)
+			createTestCRDAndRequirement(testCRD, requirement)
+			waitForRequirementReady(requirement)
```
mdbooth: There's a lot of duplicated preamble here. We'd typically try to structure this nested. Perhaps:

```go
  Context("When a CRD has a valid CompatibilityRequirement", func() {
    var (
      testCRD *CRD...
      required *CompatReq...
    )

    BeforeEach(func() {
      testCRD = test.GenerateTestCRD()
      requirement = test.GenerateTestCompatibilityRequirement(testCRD)    
	  createTestCRDAndRequirement(testCRD, requirement)
	  waitForRequirementReady(requirement)
    })

    It("should reject incompatible CRD updates", func() {
        ...
    })

    It("should reject CRD deletion while a CompatibilityRequirement references it", func() {
      ...
    })
  })
```

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,557 @@
...
+				if err == nil {
+					return StopTrying("CRD update succeeded but should have been rejected")
+				}
+
+				if apierrors.IsForbidden(err) {
+					return nil
+				}
+
+				return err
+			}).WithTimeout(framework.WaitShort).WithPolling(framework.RetryShort).Should(Succeed(), "incompatible CRD update should be rejected with Forbidden")
```
mdbooth: Here we should match the returned error. So keep the StopTrying return if the error is nil, but after that just return the error. Then instead of Succeed() as the matcher, it should be "MatchError(...some specific portion of the error message...)"

mdbooth (commented):
I've paused reviewing for the moment. Posted discussion point in https://redhat-internal.slack.com/archives/C05KZA3NVU6/p1781619084762889

e2e/crd_compatibility_test.go
simkam: It looks like our existing tests check the feature gate instead of CRD presence.

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+			switch infra.Status.ControlPlaneTopology {
+			case configv1.SingleReplicaTopologyMode:
+				Expect(*deployment.Spec.Replicas).To(Equal(int32(1)),
+					"SNO cluster should have 1 replica")
+			case configv1.HighlyAvailableTopologyMode:
+				Expect(*deployment.Spec.Replicas).To(BeNumerically(">", int32(1)),
+					"HA cluster should have more than 1 replica")
+			default:
+				Fail(fmt.Sprintf("unknown control plane topology %q, update this test to handle it", infra.Status.ControlPlaneTopology))
```
simkam: it might change after review but currently all topologies, except SingleReplica, have 2 replicas.

https://github.com/openshift/cluster-capi-operator/pull/600/changes#diff-a496892d68f33b5513cd57332ceb7211e5522620f72cc64ae3ec28b4af04c67cR34

stefanonardo: will fix once #600 gets merged

e2e/crd_compatibility_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+func testCRDGVK(crd *apiextensionsv1.CustomResourceDefinition) schema.GroupVersionKind {
+	return schema.GroupVersionKind{
+		Group:   crd.Spec.Group,
+		Version: crd.Spec.Versions[0].Name,
+		Kind:    crd.Spec.Names.Kind,
+	}
+}
+
+var _ = Describe("CRD Compatibility Checker [sig-cluster-lifecycle][OCPFeatureGate:CRDCompatibilityRequirementOperator]", Ordered, func() {
```
pmeida: I am not sure if there is a convention, but I do think it matters to follow this order:
```
[<sig-group>] [<feature-group>] <description>
````

Worst case we keep it consistent across openshift and on this specific repo.

**Comment by pmeida** [comment]
Just want to leave the note that the OTE framework is now fully integrated ([origin#31307](https://github.com/openshift/origin/pull/31307), [cluster-capi-operator#597](https://github.com/openshift/cluster-capi-operator/pull/597)).

Once your test is integrated with OTE (e.g. by removing the `_test` suffix from the filename), it will run automatically in these presubmits via `openshift-e2e-test` step with `TEST_SUITE=openshift/conformance/parallel`:
- e2e-aws-ovn-techpreview
- e2e-gcp-ovn-techpreview
- e2e-azure-ovn-techpreview
- e2e-openstack-ovn-techpreview

These are the presubmits that use `TPNU` and the conformance suite. So you can quickly understand the outcome of the integrated tests, although the data will not feed sippy for feature promotion as they are presubmits.
Another option is run run `/pj-rehearse` with a top-level release periodic.

**Comment by simkam** [comment]
/testwith openshift/cluster-capi-operator/main/e2e-aws-capi-techpreview openshift/cluster-capi-operator/#600

**Comment by simkam** [comment]
/testwith openshift/cluster-capi-operator/main/e2e-aws-capi-techpreview openshift/cluster-capi-operator#600

File: e2e/crd_compatibility_test.go (hunk 9 above)
Comment (simkam) [reply e2e/crd_compatibility_test.go]: @stefanonardo FYI https://github.com/openshift/cluster-capi-operator/pull/600 has been merged

File: e2e/crd_compatibility.go [hunk 11]
```diff
@@ -0,0 +1,296 @@
...
+					available := false
+					for _, c := range deployment.Status.Conditions {
+						if c.Type == appsv1.DeploymentAvailable && c.Status == corev1.ConditionTrue {
+							available = true
+
+							break
+						}
+					}
+
+					g.Expect(available).To(BeTrue(), "deployment should have Available=True condition")
```
simkam: test helper should be better than manual check

```
g.Expect(deployment.Status.Conditions).To(ContainElement(SatisfyAll(
    HaveField("Type", Equal(appsv1.DeploymentAvailable)),
    HaveField("Status", Equal(corev1.ConditionTrue)),
)))
```

e2e/crd_compatibility.go
```diff
@@ -0,0 +1,296 @@
...
+					pods := &corev1.PodList{}
+					g.Expect(cl.List(ctx, pods,
+						client.InNamespace(framework.CompatibilityRequirementsNamespace),
+						client.MatchingLabels{"k8s-app": crdCompatibilityDeploymentName},
+					)).To(Succeed())
+					g.Expect(pods.Items).ToNot(BeEmpty(), "expected at least one pod")
+
+					for _, pod := range pods.Items {
+						g.Expect(pod.Status.Phase).To(Equal(corev1.PodRunning), fmt.Sprintf("pod %s is not Running", pod.Name))
+					}
```
simkam: It could be simplified with

```
g.Expect(pods.Items).To(HaveEach(HaveField("Status.Phase", Equal(corev1.PodRunning))))
```

e2e/crd_compatibility.go
```diff
@@ -0,0 +1,296 @@
...
+				DeferCleanup(func() {
+					_ = cl.Delete(ctx, validCR)
+				})
+			})
+		})
+
+		It("should reject CRs that do not conform to the compatibility schema", func() {
+			By("Submitting a CR missing the required appName field", func() {
+				invalidCR := test.NewTestObject(testCRDGVK(testCRD)).
+					WithSpec(map[string]any{}).
```
simkam, line 272: Add the object name. The valid test above has a name
```
WithName("invalid-cr-test").
```

simkam (changes requested):

mdbooth (commented):
/approve
/lgtm

mdbooth:
All tests observed running successfully in e2e-aws-ovn-techpreview

/verified by CI

stefanonardo:
I'm investigating the flakes separately

PR #600, opened 2026-06-17
OCPCLOUD-3539: add topology-aware operator for CRD Compatibility Checker

simkam:
https://redhat.atlassian.net/browse/OCPCLOUD-3539

Implement [OCPCLOUD-3539](https://redhat.atlassian.net/browse/OCPCLOUD-3539) by converting the CRD Compatibility Checker from a static CVO-deployed Deployment to a Manager/Operand pattern. This fixes SNO conformance by making the operand topology-aware.

The new crd-compatibility-checker-operator:
- Watches Infrastructure.status.controlPlaneTopology
- Configures operand with 1 replica (no PDB) for SingleReplica topology
- Configures operand with 2 replicas + PDB for HighlyAvailable topologies
- Corrects drift when operand is manually modified
- Uses separate ServiceAccounts (operator vs operand) for least privilege

Changes:
- New controller in pkg/controllers/crdcompatibilityoperator/
- New binary cmd/crd-compatibility-checker-operator/
- Integration tests covering topology transitions and drift correction
- Operator manifests with RBAC for managing Deployments and PDBs
- Updated build pipeline (Makefile, Dockerfile.rhel)
- Renumbered CRD checker manifests to eliminate gaps

The operand Deployment is now managed dynamically by the operator instead of statically by CVO, allowing proper HA configuration without breaking SNO.

## Release Notes

* **New Features**
  * Added the CRD Compatibility Checker Operator for topology-aware management of the operand workload, including replica handling and high-availability behavior (PDB present/absent).

* **Documentation**
  * Documented the operator’s watched resources and topology-driven reconciliation/drift correction behavior.

* **Chores**
  * Updated operator packaging to ship the operator binary.
  * Added operator ServiceAccount, RBAC, metrics Service, and deployment/secret wiring.
  * Expanded egress NetworkPolicy selection for the operator.

* **Tests**
  * Added integration tests covering topology changes and drift correction.


simkam:
@coderabbitai review

manifests/0000_21_crd-compatibility-checker-operator_00_service-account.yaml
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+  name: compatibility-requirements-operator-token
+  namespace: openshift-compatibility-requirements-operator
+  annotations:
+    kubernetes.io/service-account.name: compatibility-requirements-operator
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+    release.openshift.io/feature-gate: "CRDCompatibilityRequirementOperator"
+type: Opaque
```
mdbooth: I think we don't want this secret. Is this copypasta from capi-controllers where we do need it for an unusual reason?

manifests/0000_21_crd-compatibility-checker-operator_01_rbac-roles.yaml
```diff
@@ -0,0 +1,79 @@
...
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-gate: "CRDCompatibilityRequirementOperator"
+  name: system:openshift:openshift-compatibility-requirements-operator:compatibility-requirements-operator
+rules:
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - infrastructures
+  - apiservers
```
mdbooth, line 16: reviewers note: apiservers required by tls gubbins.

pkg/controllers/crdcompatibilityoperator/controller.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+}
+
+// SetupWithManager registers the controller with the manager.
+func (r *CRDCompatibilityOperatorController) SetupWithManager(mgr ctrl.Manager) error {
+	// Single fixed reconcile key for all events
+	toFixedKey := func(ctx context.Context, _ client.Object) []reconcile.Request {
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: "fixed"}}}
+	}
+
```
mdbooth, line 70: Don't think I've seen this before: we normally seem to reconcile some arbitrary object. Can't think of any reason not to do this, though.

cmd/crd-compatibility-checker-operator/main.go
```diff
@@ -0,0 +1,141 @@
...
+		log.Error(err, "unable to initialize operator config")
+		os.Exit(1)
+	}
+
+	// Limit cache to operator namespace only
+	mgrOpts.Cache = cache.Options{
+		DefaultNamespaces: map[string]cache.Config{
+			controllers.DefaultCRDCompatibilityCheckerNamespace: {},
+		},
+		SyncPeriod: ptr.To(10 * time.Minute),
```
mdbooth: Please can we omit SyncPeriod? The default is 10 hours, which is fine. We don't actually need a sync at all.

pkg/controllers/crdcompatibilityoperator/controller.go
```diff
@@ -0,0 +1,346 @@
...
+			Name:      operandDeploymentName,
+			Namespace: r.namespace,
+		},
+	}
+
+	_, err := controllerutil.CreateOrPatch(ctx, r.client, deployment, func() error {
+		deployment.Spec.Replicas = &replicas
+		deployment.Spec.Selector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"k8s-app": operandLabel,
```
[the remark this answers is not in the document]
mdbooth: meh

pkg/controllers/crdcompatibilityoperator/controller.go
```diff
@@ -0,0 +1,358 @@
...
+		log.Error(err, "Failed to reconcile PodDisruptionBudget")
+		return ctrl.Result{}, err
+	}
+
+	log.Info("Successfully reconciled CRD Compatibility Checker operand")
+
+	return ctrl.Result{}, nil
+}
+
+//nolint:funlen
```
mdbooth: I'll call this a nit because it's not critical, but:

I personally find it really hard to k8s resources specified in full in Go like this. And it also makes the function really long, hence the linter warning. In the other places in capio where we've done similar we've used embed.FS to embed a yaml from a separate file, then we parse it before using it. It's safe and idiomatic to panic on a parse error in this case, btw.

i.e. Parse the embedded yaml, then apply topology modifications.

See https://github.com/openshift/cluster-capi-operator/blob/f2f0de3ba179137e33563644535e753d9be25d90/pkg/controllers/crdcompatibility/bindata/assets.go#L23-L24

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+		_, err = startManager(testCtx, cfg, cl)
+		Expect(err).NotTo(HaveOccurred(), "failed to start controller manager")
+	})
+
+	AfterAll(func() {
+		By("Tearing down test environment")
+		testCtxCancel()
+		Expect(testEnv.Stop()).To(Succeed(), "failed to stop test environment")
+	})
```
mdbooth: I prefer DeferCleanup for this stuff, but this can go entirely because we don't need this testEnv at all.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+		testCtx, testCtxCancel = context.WithCancel(ctx)
+
+		By("Starting fresh test environment")
+
+		testEnv = &envtest.Environment{}
+
+		var err error
+
+		cfg, cl, err = test.StartEnvTest(testEnv)
+		Expect(err).NotTo(HaveOccurred(), "failed to start test environment")
```
mdbooth: This is starting a duplicate testEnv. We already started one in suite_test.go. We should just use that.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/openshift/cluster-capi-operator/pkg/controllers"
+	"github.com/openshift/cluster-capi-operator/pkg/test"
+)
+
+const (
+	timeout  = 10 * time.Second
+	interval = 100 * time.Millisecond
+)
+
+var _ = Describe("CRDCompatibilityOperatorController", Ordered, func() {
```
mdbooth: This does not need to be Ordered. The setup in BeforeAll is very cheap and can be duplicated across parallel processes (once the redundant envtest is removed).

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+		infraObj = &configv1.Infrastructure{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: controllers.InfrastructureResourceName,
+			},
+			Spec: configv1.InfrastructureSpec{},
+			Status: configv1.InfrastructureStatus{
+				ControlPlaneTopology: configv1.HighlyAvailableTopologyMode,
+			},
+		}
+		Expect(cl.Create(testCtx, infraObj)).To(Succeed(), "failed to create Infrastructure CR")
```
mdbooth: You can move this into a BeforeEach in the appropriate Context below (i.e. 'when cluster topology is HighlyAvailable'). Because this is a cluster-scoped you should add a DeferCleanup() here which calls CleanupAndWait so the Create call doesn't fail on the next invocation.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+		Expect(err).NotTo(HaveOccurred(), "failed to start test environment")
+
+		By("Creating test namespace")
+
+		ns := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: testNamespace,
+			},
+		}
+		Expect(cl.Create(testCtx, ns)).To(Succeed(), "failed to create test namespace")
```
mdbooth: The usual pattern is to create a new namespace for each test, i.e. move this to a BeforeEach. Generate it with `generateName` so it's guaranteed to be unique. Store the Name of created object so you can use it for the Deployment and PDB below.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+		}
+		pdbKey = types.NamespacedName{
+			Name:      operandPDBName,
+			Namespace: testNamespace,
+		}
+
+		By("Starting controller manager")
+
+		_, err = startManager(testCtx, cfg, cl)
+		Expect(err).NotTo(HaveOccurred(), "failed to start controller manager")
```
mdbooth: If the tests can run safely using the same manager throughout, which is likely, you can move this into the suite initialisation. If not you can start it in BeforeEach, but you need to ensure you add a DeferCleanup which stops it and waits for it to stop before continuing.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+				return cl.Get(testCtx, pdbKey, pdb)
+			}, timeout, interval).Should(Succeed(), "PDB should be created")
+
+			Expect(pdb.Spec.MinAvailable).NotTo(BeNil(), "PDB minAvailable should not be nil")
+			Expect(pdb.Spec.MinAvailable.IntVal).To(Equal(int32(1)), "PDB should have minAvailable=1")
+			Expect(pdb.Spec.Selector.MatchLabels["k8s-app"]).To(Equal(operandLabel), "PDB selector should match operand label")
+		})
+	})
+
+	Context("when cluster topology changes to SingleReplica", func() {
```
mdbooth: I'd recommend moving this Context into the context above, so this Context becomes: `when cluster topology is HighlyAvailable when cluster topology changes to SingleReplica`

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+		BeforeEach(func() {
+			By("Updating Infrastructure to SingleReplica topology")
+			Eventually(func() error {
+				if err := cl.Get(testCtx, types.NamespacedName{Name: controllers.InfrastructureResourceName}, infraObj); err != nil {
+					return err
+				}
```
mdbooth: If you declare 'infrastructure' as a context-local variable in the parent context and populate it on Create you don't need to do this: you can just use the once you've already got.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+				infraObj.Status.ControlPlaneTopology = configv1.SingleReplicaTopologyMode
+
+				return cl.Status().Update(testCtx, infraObj)
```
mdbooth: In order for this test to be valid, before doing this you need to wait for the previous configuration to have rolled out this could be a simple:

Eventually(Get(dep, testNamespace) && Get(pdb, testNamespace) both succeed)

Then update infrastructure object and assert below that they're updated accordingly. Otherwise this is a race between the test setup and the controller.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+				if err := cl.Get(testCtx, deploymentKey, deployment); err != nil {
+					return -1
+				}
+
+				if deployment.Spec.Replicas == nil {
+					return -1
+				}
+
+				return *deployment.Spec.Replicas
+			}, timeout, interval).Should(Equal(int32(1)), "Deployment should be scaled down to 1 replica")
```
mdbooth: nit: This is ok, but my preference here is that the eventually returns the whole deployment object, and Should() asserts `HaveField("Spec.Replicas", EquivalentTo(1))`. This just means the error message will contain the whole deployment object if it fails, which is usually more useful.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+			pdb := &policyv1.PodDisruptionBudget{}
+
+			Eventually(func() bool {
+				err := cl.Get(testCtx, pdbKey, pdb)
+				return apierrors.IsNotFound(err)
+			}, timeout, interval).Should(BeTrue(), "PDB should be deleted")
+		})
+	})
+
+	Context("when cluster topology changes back to HighlyAvailable", func() {
```
mdbooth: I'd start again here with a top-level `when cluster topology is SingleReplica`.

I'd recommend writing 2 helper functions which assert each of the 2 states we're interested in. Something like:

```go
func assertHA() {
    It("should ensure Deployment exists with 2 replicas", ...)

    It("should ensure PDB exists", ...)
}

func assertSingleReplica() {
    It("should ensure Deployment exists with 1 replica", ...)

    It("should ensure PDB does not exist", ...)
}
```

You can then reuse these:

```go
Context("when cluster topology is HighlyAvailable", func() {
    BeforeEach(func() {
        // Create infrastructure object, DeferCleanup
    })

    assertHA()

    Context("and changes to SingleReplica", func() {
        BeforeEach(func() {
            // wait for HA rollout
            // update infrastructure to SingleReplica
        })

        assertSingleReplica()
    })

    Context("and dep is manually modified", func() {
        BeforeEach(func() {
            // wait for HA rollout
            // modify dep
        })

        assertHA()
    })

    Context("and dep is manually deleted", ...)
    Context("and pdb is manually modified", ...)
    Context("and pdb is manually deleted", ...)
})
```

mdbooth (changes requested):
The non-test code looks good. The only change there which is truly necessary is removing the extraneous secret from the manifests. I'd like the resource object definition cleanups, but I wouldn't insist on this.

The tests need a bit of work. Let me know if I can help with this.

manifests/0000_21_crd-compatibility-checker-operator_00_service-account.yaml
simkam: removed

cmd/crd-compatibility-checker-operator/main.go
simkam: removed

pkg/controllers/crdcompatibilityoperator/controller_test.go
simkam: updated

pkg/controllers/crdcompatibilityoperator/controller_test.go
simkam: contexts refactored to 

```
  CRDCompatibilityOperatorController (Serial)
  │
  ├── when cluster topology is HighlyAvailable
  │   ├── ✓ should create Deployment with 2 replicas and PDB
  │   ├── and changes to SingleReplica
  │   │   ├── ✓ should update to 1 replica and delete PDB
  │   │   └── and changes back to HighlyAvailable
  │   │       └── ✓ should restore 2 replicas and recreate PDB
  │   ├── and Deployment is manually modified
  │   │   └── ✓ should correct drift back to 2 replicas
  │   └── and PDB is manually deleted
  │       └── ✓ should recreate the PDB
  │
  ├── when cluster topology is SingleReplica
  │   └── ✓ should create Deployment with 1 replica and no PDB
  │
  └── when using other topology modes
      └── should create Deployment with 2 replicas and PDB (DescribeTable)
          ├── ✓ DualReplica topology
          ├── ✓ HighlyAvailableArbiter topology
          └── ✓ External topology
```

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,285 @@
...
+		})
+
+		It("should create Deployment with 2 replicas and PDB", func() {
+			assertHA()
+		})
+
+		Context("and changes to SingleReplica", func() {
+			BeforeEach(func() {
+				By("Waiting for HA state to be fully rolled out")
+				assertHA()
```
mdbooth, line 135: nit: for future reference, ginkgo's reporting is slightly nicer if you express this as:

```go
    By("Waiting for HA state to be fully rolled out", func() {
        assertHA()
    })
```

Everything in the block is associated explicitly with the By() in the output, including execution time. This works especially well for operations which may take arbitrarily long, like this one waiting for a condition to be fulfilled. Don't do this now unless you're really inspired to for some reason. This is fine.

pkg/controllers/crdcompatibilityoperator/controller_test.go
```diff
@@ -0,0 +1,285 @@
...
+					deployment.Spec.Replicas = &wrongReplicas
+
+					return k8sClient.Update(ctx, deployment)
+				}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+			})
+
+			It("should correct drift back to 2 replicas", func() {
+				assertHA()
+			})
+		})
```
mdbooth, line 200: nit: I'd just have written this as a single It(), tbh. Same below.

mdbooth (approved):
/approve
/lgtm

simkam:
/verified by included integration tests and manually on HA and SNO clusters. E2E tests will be added in https://github.com/openshift/cluster-capi-operator/pull/599

PR #603, opened 2026-06-18
OCPCLOUD-3541: Use feature gates consistently

mdbooth:
This change removes various hard-coded assumptions in the code about which
platforms are supported and uses feature gates exclusively for feature
enablement. Specifically:

* When ClusterAPIMachineMigration is enabled we now always run the operator,
  regardless of whether the platform has support.
* Consequently we always install core CAPI, even if platform support is not yet
  enabled.
* The revision controller now checks the per-platform feature gate before
  enabling any platform-specific profiles.
* The machine-api-migration entrypoint now uses the MachineAPIMigration
  featuregate and the per-platform MachineAPIMigration instead of hard-coding
  AWS and OpenStack.

## Summary by CodeRabbit

- **New Features**
  - Platform-specific components are now included only when their corresponding feature gates are enabled.
  - Machine API migration components now start only when migration is enabled for the detected platform.

- **Bug Fixes**
  - Unsupported platforms no longer trigger unnecessary deployment removal or misleading availability status.
  - Controller startup now handles disabled features and unsupported platforms cleanly without waiting indefinitely.

- **Refactor**
  - Platform detection and feature-gate handling have been streamlined for more consistent controller behavior.

### DIFF
--- a/pkg/conversion/test/fuzz/fuzz.go
+++ b/pkg/conversion/test/fuzz/fuzz.go
@@ -810,6 +810,7 @@ func MAPIMachineSetFuzzerFuncs() fuzzer.FuzzerFuncs {
 			func(m *mapiv1beta1.MachineSetStatus, c randfill.Continue) {
 				c.FillNoCustom(m)
 
+				m.LabelSelector = ""         // Ignore this field. Both MAPI and CAPI (field is called 'selector') set this from spec.selector.
 				m.ObservedGeneration = 0     // Ignore, this field as it shouldn't match between CAPI and MAPI.
 				m.AuthoritativeAPI = ""      // Ignore, this field as it is not present in CAPI.
 				m.SynchronizedGeneration = 0 // Ignore, this field as it is not present in CAPI.
--- a/pkg/util/featuregates.go
+++ b/pkg/util/featuregates.go
@@ -110,3 +110,26 @@ func IsCAPIEnabledForPlatform(currentFeatureGates featuregates.FeatureGate, plat
 		return false
 	}
 }
+
+// IsMAPIMigrationEnabledForPlatform returns true if MAPI migration is enabled for the given
+// platform in the given feature gates.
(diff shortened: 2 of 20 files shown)


pkg/util/featuregates.go
```diff
@@ -110,3 +110,26 @@ func IsCAPIEnabledForPlatform(currentFeatureGates featuregates.FeatureGate, plat
 		return false
 	}
 }
+
+// IsMAPIMigrationEnabledForPlatform returns true if MAPI migration is enabled for the given
+// platform in the given feature gates.
+func IsMAPIMigrationEnabledForPlatform(platform configv1.PlatformType, featureGates featuregates.FeatureGate) bool {
```
pmeida: nit:

`IsCAPIEnabledForPlatform` takes `(featureGates, platform)`, but the new `IsMAPIMigrationEnabledForPlatform` takes `(platform, featureGates)`. We can make the order consistent.

pkg/conversion/test/fuzz/fuzz.go
```diff
@@ -810,6 +810,7 @@ func MAPIMachineSetFuzzerFuncs() fuzzer.FuzzerFuncs {
 			func(m *mapiv1beta1.MachineSetStatus, c randfill.Continue) {
 				c.FillNoCustom(m)
 
+				m.LabelSelector = ""         // Ignore, this field. Both MAPI and CAPI (field is called 'selector') set this from spec.selector.
```
pmeida: nit:

The comma here creates an awkward read

pmeida (commented):
Clean refactor.
Just minor nits.

pkg/conversion/test/fuzz/fuzz.go
mdbooth: I can't believe a clanker would have written that, but I can't believe I would have written that either 😅 Gonna blame a clanker.

pkg/util/featuregates.go
mdbooth: Just about worth a respin, I think.

PR #605, opened 2026-06-19
NO-JIRA: Fix a race in MachineSet e2es

mdbooth:
We were using getAWSMachineTemplateByPrefix, which assumes that only 1 template exists with the given prefix. But this relies on the previous template having been garbage collected before we check. Instead we just use the new name directly as we already have it.

## Summary by CodeRabbit

* **Tests**
  * Improved the MachineSets update end-to-end test to reliably validate that a newly created InfraTemplate results in the expected AWSMachineTemplate `InstanceType` update, using wait-for-condition polling to prevent flaky timing issues.


damdo (commented):
/approve
/lgtm

PR #611, opened 2026-06-26
OCPBUGS-92817: delete MAPI MachineSets before CAPI in e2e cleanup

pmeida:
## Summary

Fixes a deadlock in `cleanupMachineSetTestResources` that causes `e2e-aws-capi-techpreview` to fail with a 15-minute timeout.

When a test creates both a CAPI MachineSet and a MAPI MachineSet with the same name and `authoritativeAPI: ClusterAPI`, deleting CAPI first causes the sync controller to loop in `reconcileCAPItoMAPIMachineSetDeletionNormal` - it waits for the CAPI-specific finalizer (`cluster.x-k8s.io/machineset`) to be removed, but its own constant requeues conflict with the CAPI controller's finalizer removal patch, deadlocking cleanup.

Deleting MAPI first instead triggers `reconcileCAPItoMAPIMachineSetDeletionCAPINotDeleting`, which removes the sync finalizer from CAPI immediately. The CAPI MachineSet can then be deleted cleanly with no sync interference.

## Test plan

- [x] `e2e-aws-capi-techpreview` passes without the `Should have deleted MachineSet openshift-cluster-api/capi-ms-auth-capi-*` timeout

Fixes: https://issues.redhat.com/browse/OCPBUGS-92817

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted test cleanup order for machine set migrations so managed resources are removed in the correct sequence.
  * Improved cleanup reliability by deleting MAPI machine sets before CAPI machine sets, helping avoid leftover resources during end-to-end runs.

### DIFF
--- a/e2e/machineset_migration_helpers.go
+++ b/e2e/machineset_migration_helpers.go
@@ -443,22 +443,12 @@ func updateCAPIMachineSetInfraTemplate(capiMachineSet *clusterv1.MachineSet, new
 	)
 }
 
-// cleanupMachineSetTestResources deletes CAPI MachineSets, MAPI MachineSets, and AWSMachineTemplates created during tests.
+// cleanupMachineSetTestResources deletes MAPI MachineSets, CAPI MachineSets, and AWSMachineTemplates created during tests.
 func cleanupMachineSetTestResources(ctx context.Context, cl client.Client, capiMachineSets []*clusterv1.MachineSet, awsMachineTemplates []*awsv1.AWSMachineTemplate, mapiMachineSets []*mapiv1beta1.MachineSet) {
 	GinkgoHelper()
 
 	cleanupCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
 	defer cancel()
-
-	for _, ms := range capiMachineSets {
-		if ms == nil {
-			continue
-		}
-
-		By(fmt.Sprintf("Deleting CAPI MachineSet %s", ms.Name))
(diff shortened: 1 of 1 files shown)


pmeida:
This fixes the test's cleanup order to use the supported deletion path (MAPI first). When MAPI is deleted first, `reconcileCAPItoMAPIMachineSetDeletionCAPINotDeleting` removes the sync finalizer from CAPI immediately and steps aside, allowing clean deletion.

The root cause of the deadlock is in `reconcileCAPItoMAPIMachineSetDeletionNormal`: when CAPI is deleted first, the sync controller requeues indefinitely waiting for the CAPI-specific finalizer to be removed, and those requeues can conflict with the CAPI controller's own finalizer removal patch. 
This fix avoids triggering that path in tests but it doesnt solve the core issue.

theobarberbany:
/lgtm
/approve

nice one :)

pmeida:
/testwith openshift/cluster-capi-operator/main/e2e-aws-capi-techpreview-post-install openshift/cloud-credential-operator#1053
/testwith openshift/cluster-capi-operator/main/e2e-aws-capi-techpreview openshift/cloud-credential-operator#1053

pmeida:
/testwith openshift/cluster-capi-operator/main/e2e-aws-capi-techpreview https://github.com/openshift/cloud-credential-operator/pull/1053

damdo (commented):
/lgtm

PR #614, opened 2026-07-02
OCPBUGS-84321: ClusterAPIMachineManagement: capi-controllers: excessive restarts of during cluster installation

sg00dwin:
## Summary

- Move the capi-controllers Deployment from CVO static manifests to the operator's manifests-gen pipeline at installOrder 30
- This ensures all provider CRDs are installed and confirmed Established before capi-controllers starts, eliminating ~10 crash-restart cycles during cluster installation
- Use `adopt-existing: always` annotation for seamless ownership transfer on upgrade — no tombstone, no downtime

## Problem

During cluster installation, CVO deploys the capi-controllers Deployment simultaneously with the capi-operator installer. The installer needs time to install provider CRDs (e.g. `AWSCluster`, `Cluster`), but capi-controllers starts immediately and crashes because those CRDs don't exist yet. Both containers (`capi-controllers` and `machine-api-migration`) exit with `os.Exit(1)` and restart ~10 times before the CRDs are eventually available.

## Solution

The operator's Boxcutter installer uses probe-gated sequencing — components at a higher installOrder only deploy after earlier components are confirmed ready. By moving the Deployment to installOrder 30 (after core CRDs at 10 and providers at 20), the controllers only start once all required CRDs exist.

No changes to controller logic. The fix is purely deployment ordering.

### Upgrade path (adopt-existing)

The Deployment uses the `capi-operator.openshift.io/adopt-existing: "always"` annotation, which tells Boxcutter to adopt the existing CVO-managed Deployment in-place via Server-Side Apply with ForceOwnership. No tombstone is needed — CVO simply stops managing the Deployment when the file is removed from `manifests/`, and Boxcutter patches it in-place. No delete, no gap, no pod restarts during ownership transfer.

On fresh install, the annotation is a no-op — Boxcutter creates the Deployment normally.

### Self-image-ref

The `--self-image-ref` flag is needed because release tooling rewrites image references in CVO manifests (`manifests/`), but not in manifests embedded inside the operator image (`capi-operator-manifests/`). Without it, the Deployment would be created with the build-time CI placeholder image, which doesn't exist in production. At runtime, the operator replaces the placeholder with the actual image digest from the pod spec.

### Test utility change

The `LoadAdmissionPolicyProfiles()` function now includes the Deployment in its kustomize output. A type filter was added to only keep `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` objects, so the Deployment doesn't interfere with admission policy tests.

## Changes

| File | Change |
|---|---|
| `ocp-manifests-input/default/capi-controllers-deployment.yaml` | New — Deployment YAML with adopt-existing annotation |
| `ocp-manifests-input/default/kustomization.yaml` | New — aggregates admission-policies + deployment for manifests-gen |
| `admission-policies/default/kustomization.yaml` | Removed the deployment resource (VAPs only now) |
| `Makefile` | Point `--kustomize-dir` at `ocp-manifests-input/default`; added `--self-image-ref` |
| `capi-operator-manifests/default/manifests.yaml` | Regenerated — now includes the Deployment |
| `capi-operator-manifests/default/metadata.yaml` | Regenerated — now includes selfImageRef |
| `manifests/0000_30_cluster-api_17_deployment.yaml` | Deleted — CVO no longer manages this |
| `pkg/admissionpolicy/testutils/util.go` | Added type filter to skip non-policy objects |

---

Assisted-by: Claude (Opus 4.6)

## Summary by CodeRabbit

- **New Features**
  - Added the Cluster API controllers Deployment to the default manifests, including required SCC/management settings and control-plane scheduling.
  - Added explicit liveness/readiness probes and diagnostics/health endpoints, plus webhook/mTLS secret mounts.
  - Added a default Kustomize composition that includes admission policies and the controllers Deployment, along with controller self-image metadata and updated adoption behavior.

- **Bug Fixes**
  - Admission policy profile loading now retains only recognized ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding objects.

- **Chores**
  - Updated default manifest generation to use the new input configuration and controller self image reference.


capi-operator-manifests/default/manifests.yaml
```diff
@@ -945,3 +945,105 @@ spec:
...
+      volumes:
+      - name: cert
+        secret:
+          defaultMode: 420
+          secretName: capi-controllers-webhook-service-cert
+      - name: metrics-cert
+        secret:
+          defaultMode: 420
+          secretName: capi-controllers-metrics-tls
+status: {}
```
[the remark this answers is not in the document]
sg00dwin: These settings don't exist in the original Deployment on `main` (`manifests/0000_30_cluster-api_17_deployment.yaml`) — no securityContext, no resource limits, no probes. This PR moves that Deployment between deployment mechanisms without changing its spec. Adding hardening would be a separate effort beyond the scope of this bug fix.

On limits specifically — `customizeDeployment()` in manifests-gen explicitly strips resource limits from all operator-managed Deployments, so adding them to the source manifest would have no effect on the generated output.

capi-operator-manifests/default/manifests.yaml
```diff
@@ -945,3 +945,105 @@ spec:
...
+          protocol: TCP
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+        terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - mountPath: /tmp/k8s-metrics-server/serving-certs
+          name: metrics-cert
+          readOnly: true
```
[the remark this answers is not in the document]
sg00dwin: **Probes:** The healthz ports (9440, 9441) are declared in the Deployment manifest but neither binary actually configures a health endpoint — there's no `HealthProbeBindAddress` in `cmd/capi-controllers/main.go` or `cmd/machine-api-migration/main.go`. Adding probes against those ports would cause the pods to fail health checks and crash-loop. The original Deployment on `main` has no probes for the same reason.

**securityContext:** Same as the other comment — the original Deployment on `main` has no securityContext. This PR moves the Deployment between deployment mechanisms without changing its spec. Hardening would be a separate effort.

capi-operator-manifests/default/manifests.yaml
```diff
@@ -945,3 +945,105 @@ spec:
...
+        - containerPort: 8443
+          name: diagnostics-o
+          protocol: TCP
+        - containerPort: 9440
+          name: healthz-o
+          protocol: TCP
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
```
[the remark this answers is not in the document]
sg00dwin: Omitting resource limits is intentional — it's an OpenShift convention. The manifests-gen pipeline [explicitly strips limits](https://github.com/openshift/cluster-capi-operator/blob/main/manifests-gen/customizations.go#L178) from all operator-managed Deployments per the [OpenShift enhancement conventions](https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits). Adding them to the source manifest would have no effect on the generated output.

pkg/admissionpolicy/testutils/util.go
```diff
@@ -274,7 +274,10 @@ func LoadAdmissionPolicyProfiles() map[string][]client.Object {
 			o, _, err := scheme.Codecs.UniversalDeserializer().Decode(r.Raw, nil, nil)
 			Expect(err).WithOffset(1).NotTo(HaveOccurred())
 
-			if co, ok := o.(client.Object); ok {
+			switch co := o.(type) {
+			case *admissionregistrationv1.ValidatingAdmissionPolicy:
```
theobarberbany, line 278: What's the motivation here ? Missing any explanation in the PR description :)

admission-policies/default/kustomization.yaml
```diff
@@ -14,3 +14,4 @@ resources:
   - prevent-migration-when-machine-updating.yaml
   - provide-warning-when-not-synchronized.yaml
   - authoritative-api-transition-requires-capi-infrastructure-ready.yaml
+  - capi-controllers-deployment.yaml
```
theobarberbany: Hmm - this isn't an admission policy? why are we in the admission policies folder / kustomization?

manifests/0000_30_cluster-api_00_tombstones-5.0.yaml
```diff
@@ -0,0 +1,16 @@
+# Tombstones for objects moved from CVO to operator-managed during 5.0.
```
theobarberbany: Do we need/want a CVO tombstone here?  I don't understand the motivation.

If we delete the CAPI controllers, we will have a period of time where there are no CAPI controllers running before they're recreated by the CAPI installer. 

I think we need to ensure the ordering is right here - will having no CAPI controllers cause the CVO to stall?

We may be able to use the `adopt-existing: always` annotation on the Deployment so boxcutter adopts the existing one with no downtime?

Makefile
```diff
@@ -49,7 +49,8 @@ ocp-manifests: manifests-gen ## Generate admission policy profiles for image emb
 		--profile-name default \
 		--kustomize-dir ./admission-policies/default \
 		--name cluster-capi-operator \
-		--install-order 30
+		--install-order 30 \
+		--self-image-ref registry.ci.openshift.org/openshift:cluster-capi-operator
```
theobarberbany, line 53: Why have we added this here?  I'm not sure we need it, but I could be wrong.

I *think* we would end up doing a replace of registry.ci.openshift.org/openshift:cluster-capi-operator with registry.ci.openshift.org/openshift:cluster-capi-operator ? 

The manifests that get shipped get rewritten by release tooling, but im not sure. One way to check would be running an e2e with this omitted.

theobarberbany:
@sg00dwin  I think this is a good start, but think the approach needs refining. 

We shouldn't be putting our deployment in the admission policies folder. 

The CVO only installs annotated manifests, so we can probably keep the deployment in the `manifests` directory, and add an additional customise.yaml pointing to the deployment and the admission policy folders

capi-operator-manifests/default/manifests.yaml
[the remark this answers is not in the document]
theobarberbany: @sg00dwin 

 Both binaries use commonoptions.go which registers:
  - /healthz (ping) on the health-addr port
  - /readyz (ping) on the same port

Default is :9440, and the machine-api-migration container overrides it to :9441 via --health-addr=:9441 in the Deployment args.

See  pkg/commoncmdoptions/commonoptions.go

It's arguably out of scope for this, but may be worth adding in if you've got time :)

pkg/admissionpolicy/testutils/util.go
sg00dwin: The Deployment is now part of the generated manifests (via kustomize), so `LoadAdmissionPolicyProfiles()` picks it up alongside the VAPs. This filter ensures only `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` objects get loaded — the Deployment is skipped so it doesn't interfere with admission policy tests.

Added an explanation to the PR description.

admission-policies/default/kustomization.yaml
sg00dwin: Moved it out. The Deployment now lives in a new `ocp-manifests-input/default/` directory with an aggregating kustomization that references both `admission-policies/default` and the Deployment. The Makefile points `--kustomize-dir` there instead. Admission policies directory is VAPs-only again.

manifests/0000_30_cluster-api_00_tombstones-5.0.yaml
sg00dwin: Dropped the tombstone entirely. The Deployment now has `capi-operator.openshift.io/adopt-existing: "always"`, so Boxcutter adopts the existing CVO-managed Deployment in-place via SSA. No delete, no gap, no pod restarts during upgrade. On fresh install, the annotation is a no-op.

Makefile
sg00dwin: It's needed because release tooling rewrites images in CVO manifests (`manifests/`), but not in manifests embedded inside the operator image (`capi-operator-manifests/`). Without `--self-image-ref`, the Deployment would keep the build-time CI placeholder image, which doesn't exist in production. The operator uses the selfImageRef mechanism to replace it at runtime with the actual image digest from the pod spec.

sg00dwin:
## Updated per review

Changes since last review:

1. **Moved Deployment out of `admission-policies/`** — now in `ocp-manifests-input/default/` with an aggregating kustomization that references both the admission-policies directory and the Deployment. Admission policies directory is VAPs-only again.

2. **Dropped the tombstone, using `adopt-existing: always` instead** — Boxcutter adopts the existing CVO-managed Deployment in-place via SSA. No delete, no gap, no pod restarts during upgrade. On fresh install, the annotation is a no-op.

3. **Kept `--self-image-ref`** — release tooling rewrites images in CVO manifests but not in manifests embedded inside the operator image. Without it, the Deployment would use the build-time CI placeholder image. The operator replaces it at runtime with the actual image digest.

4. **Updated PR description** — added sections explaining the upgrade path, self-image-ref rationale, and test utility change.

All tests pass (27 suites), build and lint clean.

theobarberbany:
@sg00dwin -  CI looks sad, your changes look good.

/approve
/lgtm

Please ensure those two jobs pass once CI is happy again :)

capi-operator-manifests/default/manifests.yaml
```diff
@@ -945,3 +945,107 @@ spec:
...
+        ports:
+        - containerPort: 9443
+          name: webhook-server
+          protocol: TCP
+        - containerPort: 8443
+          name: diagnostics-o
+          protocol: TCP
+        - containerPort: 9440
+          name: healthz-o
+          protocol: TCP
```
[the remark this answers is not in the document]
sg00dwin: Adding liveness and readiness probes to both containers in the capi-controllers deployment — `capi-controllers` on port 9440 and `machine-api-migration` on port 9441, matching the pattern used by the `capi-operator` and `capi-installer` deployments.

pkg/admissionpolicy/testutils/util.go
stefanonardo: maybe we can add a comment here or log the skipped type

stefanonardo:
@sg00dwin the Azure/GCP/Metal3/vSphere tests failed because we do not start the health servers on these platforms. A possible solution is to start the manager even on unsupported platforms without registering controllers

simkam:
/test e2e-azure-ovn-techpreview e2e-aws-ovn-techpreview e2e-gcp-ovn-techpreview

https://github.com/openshift/ovn-kubernetes/pull/3298 has been merged so CI should improve

PR #618, opened 2026-07-03
OCPBUGS-85337: fix(manifests-gen): set imagePullPolicy IfNotPresent on provider workloads

damdo:
## Summary

- In disconnected clusters using pinned image sets, provider pods (`capi-controller-manager`, `capa-controller-manager`, etc.) crash with `ImagePullBackOff` because their containers have `imagePullPolicy: Always`. The images are already present on the node via pinning, so pulling is unnecessary and fails without registry access.
- Adds a kustomize replacement in the `manifests-gen` component that sets `imagePullPolicy: IfNotPresent` on all containers in all provider workloads at build time.
- Renames the local-config ConfigMap from `webhook-namespace-selector` to `kustomize-values` since it now holds more than just the webhook namespace value.

## Usage

Example usage in a provider repo: https://github.com/openshift/cluster-api/pull/302

## Test plan

- [ ] Vendor this change into a provider repo (e.g. `openshift/cluster-api`) and rebuild its manifests to verify `imagePullPolicy: IfNotPresent` appears on all workloads containers in the generated output.
- [ ] Verify the generated manifests are otherwise unchanged apart from the new field.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring image pull policy through cluster values.
  * Updated workload manifests to apply the configured image pull policy across supported resource types.

* **Bug Fixes**
  * Improved namespace scoping for webhook-related resources by sourcing the namespace from the shared values config.

* **Chores**
  * Consolidated manifest configuration into a single values source for easier maintenance.

### DIFF
--- a/manifests-gen/kustomization.yaml
+++ b/manifests-gen/kustomization.yaml
@@ -4,7 +4,7 @@ kind: Component
 namespace: openshift-cluster-api
 
 resources:
-  - webhook-namespace-selector.yaml
+  - kustomize-values.yaml
 
 # Scope all provider webhooks to the openshift-cluster-api namespace.
 # This prevents webhooks from intercepting requests in other namespaces,
@@ -12,7 +12,7 @@ resources:
 replacements:
   - source:
       kind: ConfigMap
-      name: webhook-namespace-selector
+      name: kustomize-values
       fieldPath: data.namespace
     targets:
       - select:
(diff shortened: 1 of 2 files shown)


damdo:
We don't need and want to run e2es for manifests gen changes as they are not exercised directly here

damdo:
/verified by @damdo 

via https://github.com/openshift/cluster-api/pull/302

PR #619, opened 2026-07-06
no-jira: capi: enable MachineTaintPropagation feature gate on CAPI controllers

tthvo:
## Description

Set `EXP_MACHINE_TAINT_PROPAGATION=true` in envsubst substitutions so the upstream CAPI controller-manager deploys with `MachineTaintPropagation` enabled.

This allows taint propagation from CAPI machines to nodes, which is a requirement for edge nodes (i.e. AWS local and wavelength zones). See https://github.com/openshift/installer/pull/10625

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed environment-based feature gate substitution so `MachineTaintPropagation` now resolves to `true` as expected.
  * Preserved existing feature-gate values while applying the new default substitution behavior.

* **Tests**
  * Added coverage for the `MachineTaintPropagation` feature gate to verify the generated feature-gates string includes the correct value.

### DIFF
--- a/pkg/revisiongenerator/transform.go
+++ b/pkg/revisiongenerator/transform.go
@@ -36,6 +36,8 @@ func envSubstSubstitutions(key string) string {
 	// provider metadata.
 	case "EXP_BOOTSTRAP_FORMAT_IGNITION":
 		return "true"
+	case "EXP_MACHINE_TAINT_PROPAGATION":
+		return "true"
 	default:
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


tthvo:
/close 

There is a [downstream patch](https://github.com/openshift/cluster-api/blob/4917d86210f4decb58a7added337c209335d1314/openshift/patches/enable-metadata-syncing.yaml#L15) for CAPI controller deployment that we can update to enable the feature gate by default. Thus, this change is no longer necessary.

PR #621, opened 2026-07-09
OCPCLOUD-3538,OCPCLOUD-3556: manifests-gen: enable KRM plugins and add exclude/rename-resources transformers

damdo:
## Summary

- Enable kustomize KRM function plugins in manifests-gen so that providers can reference container-based transformers (e.g. starlark) in their kustomization.yaml.
- Set `AsCurrentUser` to run KRM containers as the current user instead of `nobody`, avoiding `crun: setgroups: Invalid argument` in nested rootless podman (OpenShift CI). Kustomize hardcodes `--user nobody` by default, which triggers a `setgroups()` call that fails in nested user namespaces.
- Add two generic KRM starlark transformers that providers can use:
  - `exclude-resources`: excludes CRDs and filters associated webhook and RBAC entries. Providers supply config via an `exclude-resources-config` ConfigMap.
  - `rename-resources`: renames resources by apiGroup/Kind/name and automatically updates known references (ClusterRoleBinding/RoleBinding roleRef, binding subjects for ServiceAccounts, webhook clientConfig for Services). Providers supply config via a `rename-resources-config` ConfigMap.

## Context

See https://github.com/openshift/cluster-api-provider-aws/pull/618 and https://github.com/openshift/cluster-api/pull/304 for demonstrations of this framework in use. Those PRs currently vendor these changes locally. Once this PR merges, they will re-vendor and drop the local overrides.

## Test plan

- [ ] `cd manifests-gen && go build ./...` compiles
- [ ] Existing providers that do not reference the transformers are unaffected (no-op when not referenced in kustomization.yaml)
- [ ] CAPA PR #618 and core CAPI PR #304 validate the full end-to-end flow

## Summary by CodeRabbit

* **New Features**
  * Manifest generation now supports additional provider-defined KRM function plugins during output creation.
  * Added new config-driven resource transformers:
    * Exclude transformer to remove selected resources, including CRDs, and to prune related webhook and RBAC rules.
    * Rename transformer to rename resources and update all dependent references in bindings and webhook service targets.

### DIFF
--- a/manifests-gen/generate.go
+++ b/manifests-gen/generate.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/kustomize/api/krusty"
+	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/yaml"
 )
@@ -66,7 +67,16 @@ func generateKustomizeResources(kustomizeDir string) ([]client.Object, error) {
 	// Compile assets using kustomize.
 	fmt.Printf("> Generating OpenShift manifests based on kustomize.yaml from %q\n", kustomizeDir)
 
-	k := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+	opts := krusty.MakeDefaultOptions()
+	// EnabledPluginConfig lifts plugin restrictions (PluginRestrictionsNone) so
+	// non-builtin plugins such as container-based KRM functions (e.g. starlark
+	// transformers) are allowed. BploUseStaticallyLinked tells kustomize to
+	// resolve builtin plugins from compiled-in code rather than the filesystem.
+	opts.PluginConfig = types.EnabledPluginConfig(types.BploUseStaticallyLinked)
+	// Run KRM containers as the current user instead of "nobody" to avoid
+	// "crun: setgroups: Invalid argument" in nested rootless podman (OpenShift CI).
+	opts.PluginConfig.FnpLoadingOptions.AsCurrentUser = true
+	k := krusty.MakeKustomizer(opts)
 	fSys := filesys.MakeFsOnDisk()
 
 	res, err := k.Run(fSys, kustomizeDir)
--- a/manifests-gen/transformers/exclude-resources.yaml
+++ b/manifests-gen/transformers/exclude-resources.yaml
@@ -1,0 +1,180 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: exclude-resources
+  annotations:
+    config.kubernetes.io/function: |
+      container:
+        image: ghcr.io/kptdev/krm-functions-catalog/starlark@sha256:902f801cd887ec88cc89275672e98811dc3456dcbede2027325ac45d60c84867
+data:
+  source: |
+    # exclude-resources: removes CRDs and their associated resources from the
+    # kustomize output.
+    #
+    # Config: a ConfigMap named "exclude-resources-config" with an "excludeCRDs"
+    # key in data. Each line of its value is of the form: plural.group (e.g. machinepools.cluster.x-k8s.io)
+    #
+    # For each excluded CRD, the transformer cascades removal to:
+    #   - MutatingWebhookConfiguration/ValidatingWebhookConfiguration entries
+    #     whose rules reference the excluded resource
+    #   - ClusterRole RBAC rules that grant access to the excluded resource
+    #     (including sub-resources like /status, /finalizers)
+    #
+    # If all webhook entries in a configuration are excluded, the entire
+    # webhook configuration resource is removed.
+
+    def find_config(items):
+      """Find the exclude-resources-config ConfigMap from the resource list."""
+      for item in items:
+        if item.get("kind") == "ConfigMap" and item.get("metadata", {}).get("name") == "exclude-resources-config":
+          return item.get("data", {})
+      return None
+
+    def parse_exclude_crds(config):
+      """Parse CRD names into {group: set(plurals)} and a set of full CRD names."""
+      by_group = {}
+      crd_names = []
+      raw = config.get("excludeCRDs", "").strip()
+      if not raw:
+        return by_group, crd_names
+      for line in raw.split("\n"):
+        crd_name = line.strip()
+        if not crd_name:
+          continue
+        parts = crd_name.split(".", 1)
+        if len(parts) != 2:
+          fail("invalid CRD name (expected plural.group): " + crd_name)
+        plural = parts[0]
+        group = parts[1]
+        if group not in by_group:
+          by_group[group] = []
+        by_group[group].append(plural)
+        crd_names.append(crd_name)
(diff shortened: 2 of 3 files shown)


manifests-gen/transformers/exclude-resources.yaml
```diff
@@ -0,0 +1,149 @@
...
+          # 3. Filter RBAC rules + rename ClusterRole
+          if kind == "ClusterRole":
+            rules = resource.get("rules", [])
+            filtered_rules = []
+            for rule in rules:
+              if not rule_has_excluded_group(rule, by_group):
+                filtered_rules.append(rule)
+                continue
+              remaining = filter_rule_resources(rule, by_group)
+              if remaining:
```
theobarberbany, line 173: We don't exclude the `exclude-resources` configmap, this file? 

Dont think this is a problem though?

manifests-gen/transformers/exclude-resources.yaml
```diff
@@ -0,0 +1,149 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: exclude-resources
+  annotations:
+    config.kubernetes.io/function: |
+      container:
+        image: ghcr.io/kptdev/krm-functions-catalog/starlark:v0.5.0
```
theobarberbany: pin to digest not tag?

manifests-gen/transformers/exclude-resources.yaml
```diff
@@ -0,0 +1,149 @@
...
+          fail("invalid CRD name (expected plural.group): " + crd_name)
+        plural = parts[0]
+        group = parts[1]
+        if group not in by_group:
+          by_group[group] = []
+        by_group[group].append(plural)
+        crd_names.append(crd_name)
+      return by_group, crd_names
+
+    def parse_rename_roles(config):
```
theobarberbany: nit: we don't check new exists?

manifests-gen/generate.go
```diff
@@ -66,7 +67,14 @@ func generateKustomizeResources(kustomizeDir string) ([]client.Object, error) {
 	// Compile assets using kustomize.
 	fmt.Printf("> Generating OpenShift manifests based on kustomize.yaml from %q\n", kustomizeDir)
 
-	k := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+	opts := krusty.MakeDefaultOptions()
+	// Enable KRM function plugins (e.g. starlark container transformers)
+	// so that providers can use them in their kustomization.yaml.
+	opts.PluginConfig = types.EnabledPluginConfig(types.BploUseStaticallyLinked)
```
theobarberbany, line 75: This is a little confusing to me with what it's actually doing, but I think the comment covers it. I'm also not an expert, so this may be wrong.

nit: might explain how `BploUseStaticallyLinked` actually enables the star lark plugin?

theobarberbany:
A few nits.

We also have no testing of the starlark logic, so if it breaks we will only catch it once we revendor changes into a provider. Not sure if that's worth doing at this point though.   

/lgtm

manifests-gen/transformers/exclude-resources.yaml
damdo: Not a problem. The `exclude-resources-config` ConfigMap (supplied by the provider) is marked with `config.kubernetes.io/local-config: "true"`, so kustomize suppresses it from output. The `exclude-resources` transformer ConfigMap itself (this file) is also not emitted because kustomize strips function-annotated ConfigMaps automatically.

manifests-gen/transformers/exclude-resources.yaml
damdo: Switched to the mirrored image at `quay-proxy.ci.openshift.org/openshift/ci:ci_krm-functions-catalog-starlark_v0.5.0`. So I think we don't need to pin to digest.

manifests-gen/transformers/exclude-resources.yaml
damdo: Thanks done

manifests-gen/generate.go
damdo: Added comment to better explain

manifests-gen/transformers/exclude-resources.yaml
```diff
@@ -0,0 +1,180 @@
...
+    config.kubernetes.io/function: |
+      container:
+        image: ghcr.io/kptdev/krm-functions-catalog/starlark@sha256:902f801cd887ec88cc89275672e98811dc3456dcbede2027325ac45d60c84867
+data:
+  source: |
+    # exclude-resources: removes CRDs and their associated resources from the
+    # kustomize output.
+    #
+    # Config: a ConfigMap named "exclude-resources-config" with an "excludeCRDs"
+    # field. Each line: plural.group (e.g. machinepools.cluster.x-k8s.io)
```
hongkailiu: nit:

https://pkg.go.dev/k8s.io/api/core/v1#ConfigMap

```suggestion
    # key in data. Each line of its value is of the form: plural.group (e.g. machinepools.cluster.x-k8s.io)
```

manifests-gen/transformers/rename-resources.yaml
```diff
@@ -0,0 +1,120 @@
...
+  annotations:
+    config.kubernetes.io/function: |
+      container:
+        image: ghcr.io/kptdev/krm-functions-catalog/starlark@sha256:902f801cd887ec88cc89275672e98811dc3456dcbede2027325ac45d60c84867
+data:
+  source: |
+    # rename-resources: renames Kubernetes resources by apiGroup/Kind/name.
+    #
+    # Config: a ConfigMap named "rename-resources-config" with a "renameResources"
+    # field. Each line: apiGroup/Kind/oldName=newName
```
hongkailiu: nit: Same here.

```suggestion
    # key in data. Each line of its value is of the form: apiGroup/Kind/oldName=newName
```

hongkailiu (commented):
Only two NITs. Free to ignore.

manifests-gen/transformers/exclude-resources.yaml
```diff
@@ -0,0 +1,180 @@
...
+      return matched
+
+    def is_excluded_resource(resource_name, group, by_group):
+      """Check if a resource name (possibly with sub-resource) is excluded.
+
+      A wildcard ("*") in resources matches all excluded plurals in the group.
+      """
+      if group not in by_group:
+        return False
+      if resource_name == "*":
```
theobarberbany, line 76: Is this saying - if we have wildcards anywhere, we want to drop them / exclude? So enforcing more strict RBAC / role writing?

damdo: It's coming out of this comment: https://github.com/openshift/cluster-capi-operator/pull/621#discussion_r3674620741

damdo: The wildcard handling is scoped to excluded groups only so if a ClusterRole rule has resources: * but its apiGroups don't overlap with any excluded group, the rule is kept

e.g. upstream CAPI has a rule like apiGroups: addons.cluster.x-k8s.io, resources: *

Since we exclude all CRDs in that group, that wildcard rule is granting access to nothing we ship, so the transformer now drops it

theobarberbany: Ahh ok gotcha. Makes sense!

PR #623, opened 2026-07-10
OCPBUGS-98329: Add aggregated cluster role for extended per-provider RBAC

mdbooth:
Adds `openshift-capi-installer:provider-permissions` ClusterRole and binds it to capi-installer. This is an aggregated cluster rule which uses the label `capi-operator.openshift.io/aggregate-to-installer`. If a provider's manifests include any object which the default CAPIO RBAC does not cover they can specify their additional requirements in their own repo as a separate ClusterRole which aggregates to this one.

Also fixes a bug initialising boxcutter's tracking cache which turns missing RBAC into a panic.

## Summary by CodeRabbit

- **New Features**
  - Added aggregated installer provider RBAC: a new cluster-scoped `ClusterRole` and a corresponding `ClusterRoleBinding` to grant `capi-installer` the default provider-permissions via label-driven aggregation.
- **Bug Fixes**
  - Improved installer managed tracking by wiring the correct REST mapping for cluster-scoped resources.
- **Documentation**
  - Expanded the provider contract with a new “capi-installer permissions” section, including an example of extending permissions through the aggregated `ClusterRole`.
- **Tests**
  - Increased test logging verbosity and refined machine sync readiness/sentinel workflow assertions.

### DIFF
--- a/pkg/controllers/machinesync/machine_sync_controller_test.go
+++ b/pkg/controllers/machinesync/machine_sync_controller_test.go
@@ -1439,44 +1439,45 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
 
 		Context("cluster api vap tests", func() {
 			BeforeEach(func() {
-				By("Waiting for VAP to be ready")
+				By("Waiting for VAP to be ready", func() {
+					machineVap = &admissionregistrationv1.ValidatingAdmissionPolicy{}
+					machineVap.Name = clusterAPIMachineVAPName
+					Eventually(k.Update(machineVap, func() {
+						admissiontestutils.AddSentinelValidation(machineVap)
+					})).Should(Succeed())
 
-				machineVap = &admissionregistrationv1.ValidatingAdmissionPolicy{}
-				Eventually(k8sClient.Get(ctx, client.ObjectKey{Name: clusterAPIMachineVAPName}, machineVap), setupTimeout).Should(Succeed())
-				Eventually(k.Update(machineVap, func() {
-					admissiontestutils.AddSentinelValidation(machineVap)
-				})).Should(Succeed())
+					sentinelGeneration := machineVap.Generation
+					Eventually(k.Object(machineVap), setupTimeout).Should(
+						HaveField("Status.ObservedGeneration", BeNumerically(">=", sentinelGeneration)),
+					)
+				})
 
-				Eventually(k.Object(machineVap), setupTimeout).Should(
-					HaveField("Status.ObservedGeneration", BeNumerically(">=", 2)),
-				)
+				By("Updating the VAP binding", func() {
+					policyBinding = &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}
+					Eventually(k8sClient.Get(ctx, client.ObjectKey{
+						Name: clusterAPIMachineVAPName}, policyBinding), setupTimeout).Should(Succeed())
 
-				By("Updating the VAP binding")
+					Eventually(k.Update(policyBinding, func() {
+						admissiontestutils.UpdateVAPBindingNamespaces(policyBinding, mapiNamespace.GetName(), capiNamespace.GetName())
+					}), timeout).Should(Succeed())
 
-				policyBinding = &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}
-				Eventually(k8sClient.Get(ctx, client.ObjectKey{
-					Name: clusterAPIMachineVAPName}, policyBinding), setupTimeout).Should(Succeed())
+					// Wait until the binding shows the patched values
+					Eventually(k.Object(policyBinding), setupTimeout).Should(
+						SatisfyAll(
+							HaveField("Spec.ParamRef.Namespace",
+								Equal(mapiNamespace.GetName())),
 
-				Eventually(k.Update(policyBinding, func() {
-					admissiontestutils.UpdateVAPBindingNamespaces(policyBinding, mapiNamespace.GetName(), capiNamespace.GetName())
-				}), timeout).Should(Succeed())
+							HaveField("Spec.MatchResources.NamespaceSelector.MatchLabels",
+								HaveKeyWithValue("kubernetes.io/metadata.name",
+									capiNamespace.GetName())),
+						),
+					)
+				})
 
-				// Wait until the binding shows the patched values
-				Eventually(k.Object(policyBinding), setupTimeout).Should(
-					SatisfyAll(
-						HaveField("Spec.ParamRef.Namespace",
-							Equal(mapiNamespace.GetName())),
-
-						HaveField("Spec.MatchResources.NamespaceSelector.MatchLabels",
-							HaveKeyWithValue("kubernetes.io/metadata.name",
-								capiNamespace.GetName())),
-					),
-				)
-
-				By("Creating the sentinel CAPI infra machine")
-
-				capaMachine = capaMachineBuilder.WithName("sentinel-machine").Build()
-				Eventually(k8sClient.Create(ctx, capaMachine)).Should(Succeed(), "capa machine should be able to be created")
+				By("Creating the sentinel CAPI infra machine", func() {
+					capaMachine = capaMachineBuilder.WithName("sentinel-machine").Build()
+					Eventually(k8sClient.Create(ctx, capaMachine)).Should(Succeed(), "capa machine should be able to be created")
+				})
 
 				capaMachineRef := clusterv1.ContractVersionedObjectReference{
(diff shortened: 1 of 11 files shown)


mdbooth:
/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

mdbooth:
Failed due to a typo in the featuregate name

/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

mdbooth:
/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

mdbooth:
/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

mdbooth:
For anybody keeping count: these are all infra capacity issues. Prow has not yet managed to execute this job at all. It has not failed.

/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

mdbooth:
/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

mdbooth:
Looks like another flake. I'll investigate it separately.

/test unit

mdbooth:
/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -1439,92 +1439,95 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
-							Kind:               machineKind,
-							APIVersion:         clusterv1.GroupVersion.String(),
-							Name:               testMachine.Name,
-							UID:                testMachine.UID,
-							BlockOwnerDeletion: ptr.To(true),
-							Controller:         ptr.To(false),
-						},
-					})
-				})).Should(Succeed())
+				By("Creating a sentinel CAPI machine", func() {
```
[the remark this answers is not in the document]
mdbooth: The By() block populates the testMachine. I prefer them to be visually linked.

mdbooth:
/testwith openshift/cluster-capi-operator/main/e2e-metal3-capi-techpreview openshift/cluster-api-provider-metal3#86

mdbooth:
The metal3 job cannot pass without both this PR and https://github.com/openshift/cluster-api-provider-metal3/pull/86

The multi-pr test validates this.

/override ci/prow/e2e-metal3-capi-techpreview

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -1439,92 +1439,97 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+					Eventually(k8sClient.Create(ctx, testMachine), timeout).Should(Succeed())
 
-				By("Creating a sentinel MAPI Machine")
+					By("setting the owner reference on the sentinel CAPI infra machine")
```
pmeida: Looking at the trend.
This `By` has no callback so it's just a progress annotation. This can be a sibling step with its own callback instead + capitalize the `S`.

This might be irrelevant since most of the `By` calls in the file do not use a callback.

mdbooth: Hmm, yes. I've fat-fingered this. I don't think I intended to do this.

However, fortunately I don't think the output is confusing because "setting the owner reference on the sentinel CAPI infra machine" can reasonably be accounted to "Creating a sentinel CAPI machine". It's trivial to fix, so I have to respin this for any reason I'll fix it. But for the sake of more CI runs I'll leave it for now.

pmeida: Agree. Definitely trivial.

pkg/controllers/machinesync/machine_sync_controller_test.go
```diff
@@ -1439,92 +1439,97 @@ var _ = Describe("With a running MachineSync Reconciler", func() {
...
+					machineVap.Name = clusterAPIMachineVAPName
+					Eventually(k.Update(machineVap, func() {
+						admissiontestutils.AddSentinelValidation(machineVap)
+					})).Should(Succeed())
 
-				policyBinding = &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}
-				Eventually(k8sClient.Get(ctx, client.ObjectKey{
-					Name: clusterAPIMachineVAPName}, policyBinding), setupTimeout).Should(Succeed())
+					Eventually(k.Object(machineVap), setupTimeout).Should(
+						HaveField("Status.ObservedGeneration", BeNumerically("==", machineVap.Generation)),
```
theobarberbany: This was `=>2` to avoid races, it's basically a proxy for 'this has been acknowledged by the api server so hopefully is enforced now'. This is because there's no status on VAPs / bindings  saying something to this effect :( 

So  >= 2 = "the API server has processed at least the version with our sentinel in it." 

I think making it strictly equal what we've observed is less good. If something else changes something on the VAP we may flake.

In this case, the something else is the VAP status controller which we deliberately run here.

pmeida:
> The metal3 job cannot pass without both this PR and https://github.com/openshift/cluster-api-provider-metal3/pull/86

AFAIK the `e2e-metal3-capi-techpreview` presubmit is failing due to an OpenSSL 3.5.7 update in CentOS Stream 9 breaking the metal baremetal cluster bootstrap (this is already acked in internal discussions) - unrelated to what this PR addresses. The override is correct for that reason.

That said, the multi-PR presubmit only covers a fresh install but the definitive proof that the crash loop is resolved would come from a payload test combining both PRs against the failing job.

pkg/controllers/machinesync/machine_sync_controller_test.go
mdbooth: @theobarberbany It's ok, because `k.Object` updates `machineVap` every time it's called. This means that '==' here just means 'up to date with its current state'. It doesn't matter what that state is. This doesn't flake if something else modifies the object.

mdbooth:
> > The metal3 job cannot pass without both this PR and [openshift/cluster-api-provider-metal3#86](https://github.com/openshift/cluster-api-provider-metal3/pull/86)
> 
> AFAIK the `e2e-metal3-capi-techpreview` presubmit is failing due to an OpenSSL 3.5.7 update in CentOS Stream 9 breaking the metal baremetal cluster bootstrap (this is already acked in internal discussions) - unrelated to what this PR addresses. The override is correct for that reason.
>
> That said, the multi-PR presubmit only covers a fresh install but the definitive proof that the crash loop is resolved would come from a payload test combining both PRs against the failing job.

The multi-PR job passed: https://github.com/openshift/cluster-capi-operator/pull/623/checks?check_run_id=87110897024

pkg/controllers/machinesync/machine_sync_controller_test.go
mdbooth: OH, NO IT'S NOT! /panto voice

Because machineVap.Generation as an argument to BeNumerically is only evaluated once, before the call to Eventually. I wanted to get rid of the hard-coded `2` here, and I still do. Changing this back to '>=' would be correct here, but it would rely on the subtle behaviour that I just missed, so I'm just going to capture it explicitly in a separate variable, which is what I was going to do before I thought of this (wrong thing).

pkg/controllers/machinesync/machine_sync_controller_test.go
mdbooth: I had to touch it anyway, so I fixed this.

pkg/controllers/machinesync/machine_sync_controller_test.go
mdbooth: Fixed 👍

theobarberbany:
/lgtm
/approve

CI is a bloodbath, but the changes look sensible. Leaving verification to you

mdbooth:
This is verified by the multi-pr runs, e.g. https://github.com/openshift/cluster-capi-operator/pull/623/checks?check_run_id=87110897024

/verified by CI

mdbooth:
This is a component readiness fix.

mdbooth:
This cannot pass, yet.

/override ci/prow/e2e-metal3-capi-techpreview

damdo (commented):
Nice

/lgtm

PR #613, opened 2026-07-01
OCPCLOUD-3571: Add OTE e2e tests for ClusterAPIMachineManagementAWS feature gate

pmeida:
## Summary

Implements 6 OTE e2e tests for the `ClusterAPIMachineManagementAWS` feature gate, plus a generic platform Walk in the OTE extension registration.

## Tests

| # | Test | What it proves | Gated by |
|---|---|---|---|
| 1 | `capa-controller-manager` deployment available in `openshift-cluster-api` | AWS provider controller deployed by capi-installer | `ClusterAPIMachineManagementAWS` only |
| 2 | `awsclusters.infrastructure.cluster.x-k8s.io` CRD established | CAPA CRD bundle installed by capi-installer | `ClusterAPIMachineManagementAWS` only |
| 3 | `awsmachines.infrastructure.cluster.x-k8s.io` CRD established | CAPA CRD bundle installed by capi-installer | `ClusterAPIMachineManagementAWS` only |
| 4 | `awsmachinetemplates.infrastructure.cluster.x-k8s.io` CRD established | CAPA CRD bundle installed by capi-installer | `ClusterAPIMachineManagementAWS` only |
| 5 | `awsclustercontrolleridentities.infrastructure.cluster.x-k8s.io` CRD established | CAPA CRD bundle installed by capi-installer | `ClusterAPIMachineManagementAWS` only |
| 6 | `AWSCluster.Status.Ready=True` | AWSCluster was created and is fully ready (VPC, subnets, security groups, load balancer all reconciled) | `ClusterAPIMachineManagementAWS` only |

Tests 1-5 check that capi-installer deployed the CAPA CRD bundle. Test 6 directly checks the `AWSCluster` object - the AWS-specific infra resource this feature gate controls. `Status.Ready=True` is set by CAPA only after all sub-conditions (VPC, subnets, security groups, load balancer, etc.) succeed, making it the canonical readiness signal for the AWSCluster.

## Platform restriction (two-layer)

**Layer 1 - OTE environmentSelector (periodic jobs):**
A generic Walk reads `Label("platform:<name>")` from each spec and adds `spec.Include(PlatformEquals(platform))`. No per-platform clauses in `main.go` - adding GCP/Azure tests only requires a `Label` on the Describe block.

**Layer 2 - `BeforeEach` guard (`make e2e` presubmits):**
Platform check runs first, before any `config.openshift.io\ API calls, so the test skips cleanly on non-AWS clusters (including MicroShift). Feature gate check follows.

[...]

PR #606, opened 2026-06-23
OCPCLOUD-3507: Add OTE e2e tests for ClusterAPIMachineManagement feature gate

pmeida:
## Summary

Adds 8 non-disruptive health-check e2e tests registered with the OTE binary that validate the Cluster API stack is correctly installed and operational when the `ClusterAPIMachineManagement` feature gate is enabled. These tests provide the Sippy signal required for feature gate promotion.

## Tests

All tests are `blocking`, `capio/parallel → openshift/conformance/parallel`:

| Context | Test |
|---|---|
| Operator & controller deployments | capi-operator deployment available |
| Operator & controller deployments | cluster-api ClusterOperator reporting healthy |
| Operator & controller deployments | capi-controllers deployment available |
| Operator & controller deployments | capi-installer deployment available |
| CRD & API readiness | clusters CRD installed and established |
| CRD & API readiness | machines CRD installed and established |
| CRD & API readiness | machinesets CRD installed and established |
| Management cluster resources | management cluster kubeconfig Secret present |

## Design notes

- Validated against a real AWS TechPreview cluster — all 8 tests pass both via `bin/cluster-capi-operator-tests-ext run-suite capio/parallel` and via `make e2e`

## Dependencies

- Requires openshift/origin#31307 to be merged for OTE nightlies to discover and invoke the binary
- Can be merged independently; nightlies pick up tests once both PRs land

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for Cluster API machine management.
  * Verifies operator and controller deployments become available.
  * Confirms the Cluster API operator reports healthy status.
  * Validates core Cluster API resources are established and ready.
  * Confirms the management cluster kubeconfig is created successfully.
  * Tests run only when the corresponding feature is enabled.

PR #627, opened 2026-07-15
NO-JIRA: port OTP CAPI tests to OTE framework

stefanonardo:
## Summary
- Port 6 tests from openshift-tests-private `clusterinfrastructure/capi` to the existing OTE framework
- Tests rewritten to use controller-runtime client patterns matching existing e2e tests (no new dependencies)
- All tests tagged with `[OTP]`, `[Jira:OCPCLOUD]`, `[OCPFeatureGate:ClusterAPIMachineManagement]`, `[sig-cluster-lifecycle]` and start as `Informing`

### Files
- `e2e/operator.go` — operator configuration checks (deployment annotations, CRD installation, terminationMessagePolicy, secret sync)
- `e2e/webhook_validation.go` — webhook enforcement (infra cluster deletion denied, invalid cluster kind patch rejected, cluster deletion denied)

### Tests ported
| OTP ID | Suite | Description |
|--------|-------|-------------|
| 61980 | parallel | Workload management annotations on deployments |
| 71913 | parallel | IPAM CRDs installed |
| 73620 | parallel | terminationMessagePolicy = FallbackToLogsOnError |
| 51141 | disruptive | worker-user-data secret re-syncs after deletion |
| 76078 | disruptive | Infrastructure cluster deletion denied by webhook |
| 55205 | disruptive | Webhook rejects invalid cluster kind patch & deletion |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added ordered end-to-end validations for Cluster API infrastructure on OpenShift: verify workload management annotations on Deployment pod templates, ensure IPAM CRDs are installed, and confirm all containers use the expected termination message policy.
  * Added platform-scoped coverage to ensure worker user-data Secrets are removed and then automatically re-created.
  * Added webhook enforcement tests to confirm dry-run deletes are denied and invalid infrastructure references are rejected.
* **Chores**
  * Updated end-to-end test dependencies to support the new validations.


e2e/operator.go
```diff
@@ -0,0 +1,147 @@
...
+	framework "github.com/openshift/cluster-capi-operator/e2e/framework"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("[OTP][Jira:OCPCLOUD][OCPFeatureGate:ClusterAPIInstall][sig-cluster-lifecycle] Cluster_Infrastructure CAPI", Ordered, func() {
```
simkam: do we keep `[OTP]` tag if it is no longer in OTP? What about `[Jira:OCPCLOUD]` tag?

e2e/operator.go
```diff
@@ -0,0 +1,147 @@
+package e2e
```
simkam: do we want `otp_` filename prefix?

e2e/operator.go
stefanonardo: it's required by the OTE integration guide for tracking purposes

e2e/operator.go
stefanonardo: not really, I renamed the tests

e2e/go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/metal3-io/cluster-api-provider-metal3/api v1.11.2
 	github.com/onsi/ginkgo/v2 v2.28.1
 	github.com/onsi/gomega v1.39.1
+	github.com/openshift-eng/openshift-tests-extension v0.0.0-20260612102633-2fd5b2fa4221
```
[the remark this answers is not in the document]
stefanonardo: it's out of scope

e2e/operator.go
```diff
@@ -0,0 +1,96 @@
...
+	oteg "github.com/openshift-eng/openshift-tests-extension/pkg/ginkgo"
+	configv1 "github.com/openshift/api/config/v1"
+	framework "github.com/openshift/cluster-capi-operator/e2e/framework"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("[OTP][Jira:OCPCLOUD][OCPFeatureGate:ClusterAPIMachineManagement][sig-cluster-lifecycle] Cluster_Infrastructure CAPI", Ordered, func() {
```
pmeida: Do we need `Ordered`? Also for the other Describe block.

e2e/go.mod
pmeida: Not major but.
The openshift-tests-extension submodule was intentionally kept as a separate Go module to isolate its dependencies from the main module.

We could switch 
```diff
- oteg.Informing()
+ Label("Lifecycle:informing")
```

to eliminate the dependency since Ginkgo is already a main module dependency.

e2e/operator.go
```diff
@@ -0,0 +1,96 @@
...
+
```
pmeida: When tests are invoked via make e2e (bypassing OTE tag filtering), they'll run regardless of gate state and fail on a non-TechPreview cluster. Needs a BeforeEach FG guard.
```python
  BeforeEach(func() {
      if !framework.IsFeatureGateEnabled(ctx, cl, features.FeatureGateClusterAPIMachineManagement) {
          Skip("Feature gate ClusterAPIMachineManagement is not enabled.")
      }
  })
```

e2e/go.mod
pmeida: The other repos are not following this practise, so we might keep it as it is.

e2e/operator.go
```diff
@@ -0,0 +1,96 @@
+package e2e
```
pmeida: Both files are missing the Apache 2.0 license header that every other file in this package carries

e2e/webhook_validation.go
```diff
@@ -0,0 +1,97 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("[OTP][Jira:OCPCLOUD][OCPFeatureGate:ClusterAPIMachineManagement][sig-cluster-lifecycle] Cluster API Webhook Validation", Ordered, func() {
+
+	It("should deny deletion of infrastructure cluster resources", Label("Disruptive"), oteg.Informing(), func() {
+		switch platform {
```
pmeida: This is a nice addon for skiping when using `make e2e` and it also skips the test at runtime when running with `openshift-tests`.

However, #613 introduces `Label("platform:<name>")` with OTE suite-level wiring that filters at **scheduling** time. 
Adding the label avoids OTE spinning up test infrastructure for these tests on unsupported platforms. 
This might be a follow up, since the convention is only introduced at #613

e2e/go.mod
stefanonardo: I believe the OTE migration skill did this, I'll replace it with `Label`

pmeida: maybe it was coming from here: https://github.com/openshift/cluster-capi-operator/blob/main/openshift-tests-extension/README.md?plain=1#L90

If your moving forward with it, can you also update the README?

e2e/webhook_validation.go
```diff
@@ -0,0 +1,116 @@
...
+	case configv1.GCPPlatformType:
+		return "GCPCluster"
+	case configv1.VSpherePlatformType:
+		return "VSphereCluster"
+	case configv1.AzurePlatformType:
+		return "AzureCluster"
+	default:
+		return ""
+	}
+}
```
pmeida: The codebase already has `pkg/util.GetCAPITypesForInfrastructure()` which maps platform to infra cluster kind. See `webhook_test.go` for usage.

e2e/operator.go
```diff
@@ -0,0 +1,115 @@
...
+		By("Waiting for secret to be re-created with a new UID")
+		Eventually(func() bool {
+			recreated := &corev1.Secret{}
+			if err := cl.Get(ctx, secretKey, recreated); err != nil {
+				return false
+			}
+			return recreated.UID != originalUID
+		}, framework.WaitMedium, framework.RetryMedium).Should(BeTrue(),
+			"worker-user-data secret should be re-created from %s", framework.MAPINamespace)
+	})
```
pmeida: I see your not using komega here, and other places.
```go
# example with komega
recreated := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: secretKey.Name, Namespace: secretKey.Namespace}}
Eventually(komega.Object(recreated)).WithTimeout(framework.WaitMedium).WithPolling(framework.RetryMedium).Should(
    HaveField("UID", Not(Equal(originalUID))),
)
```
This is the most impactful one - it fixes the swallowed-error problem and the missing-komega problem in a single change.
The other calls would also benefit from retry on transient API errors. The simple Gets are lower priority but still worth aligning for consistency.

pmeida (commented):
PR #606 introduces tests for the foundational deployment layer - highly deterministic, fast checks. 
This PR ports deeper operational/functional behavior tests that happen to share the same gate.
That said, most tests here are also checking static state (annotations exist, CRDs exist, pod spec fields) - equally deterministic. 
The secret re-sync test is the exception: it deletes a secret and waits for the controller to recreate it, making it a reconciliation/timing test and inherently more fragile. Also worth noting: the secret sync controller activates with the platform-specific gate (e.g. ClusterAPIMachineManagementAWS), so this test may be gated on the wrong feature gate.

I don't see direct overlaps between the two PRs, but #606 established a convention where tests for a specific feature gate live in a single file (e.g. `cluster_api_machine_management.go`, `cluster_api_machine_management_aws.go`). These PR breaks the pattern.

pmeida:
Since #606 has been merged we can already (but I am postponing to apply in one go):
- Move `operator.go` and `webhook_validation.go` `[OCPFeatureGate:ClusterAPIMachineManagement]` tests into `cluster_api_machine_management.go`
- Move `operator.go` and `webhook_validation.go` `[OCPFeatureGate:ClusterAPIMachineManagement*]` tests into `cluster_api_machine_management_*.go` - this are new files

Lets just wait for #613
after that we can:
- Move `operator.go` and `webhook_validation.go` `[OCPFeatureGate:ClusterAPIMachineManagementAWS]` tests into `cluster_api_machine_management_aws.go`
- Also apply `Label("platform:aws")` to  `[OCPFeatureGate:ClusterAPIMachineManagementAWS]` and so on for other platforms - this allows filtering at OTE level

pmeida:
I was starting to implement the above comment but stumbled with the 2 concerns.
Example test:
```
// port=maybe - 90.2% pass rate (246 runs last 60 days)
g.It("Author:zhsun-NonHyperShiftHOST-Medium-55205-Webhook validations for CAPI [Disruptive]", func() {
```

1)
>  We recommend only porting tests with port=yes (95%+ pass rate), but others can be ported if a team feels they are worthwhile
[source](https://docs.google.com/document/d/1cFZj9QdzW8hbHc3H0Nce-2xrJMtpDJrwAse9H7hLiWk/edit?tab=t.0)

The following ported tests have pass rates bellow it:
- 51141 - 89,4%
- 76078 - 89%
- 55205 - 90,2%

Maybe we shouldn't port them.

2) Most of the ported tests have `NonHyperShiftHOST` in the name. This means they be excluded from Hypershift cluster runs. We can achieve the same with a label introduced by #638

stefanonardo:
> I was starting to implement the above comment but stumbled with the 2 concerns. Example test:
> 
> ```
> // port=maybe - 90.2% pass rate (246 runs last 60 days)
> g.It("Author:zhsun-NonHyperShiftHOST-Medium-55205-Webhook validations for CAPI [Disruptive]", func() {
> ```
> 
> 
> > We recommend only porting tests with port=yes (95%+ pass rate), but others can be ported if a team feels they are worthwhile
> > [source](https://docs.google.com/document/d/1cFZj9QdzW8hbHc3H0Nce-2xrJMtpDJrwAse9H7hLiWk/edit?tab=t.0)
> 
> The following ported tests have pass rates bellow it:
> 
>     * 51141 - 89,4%
> 
>     * 76078 - 89%
> 
>     * 55205 - 90,2%
> 
> 
> Maybe we shouldn't port them.
> 
>     2. Most of the ported tests have `NonHyperShiftHOST` in the name. This means they be excluded from Hypershift cluster runs. We can achieve the same with a label introduced by [OCPBUGS-100143: e2e: skip CRD Compatibility Checker tests on External topology clusters #638](https://github.com/openshift/cluster-capi-operator/pull/638)

1. our rewrite might make these tests more stable. I'd keep them as `Informing` for a sprint and let the data speak. If they do not become stable, we drop them.

2. sounds good!

pmeida:
Moved tests from `operator.go` and `webhook_validation.go` into feature-gate-aligned files: `cluster_api_machine_management.go` for common tests, and `cluster_api_machine_management_{aws,gcp,vsphere}.go` for platform-specific secret sync tests.

Added `Label("skip-topology:External") + BeforeEach` skip guards to the two OTP Describe blocks whose original tests had `NonHyperShiftHOST` in openshift-tests-private (OTP IDs: 61980, 71913, 73620, 76078, 55205). Secret sync tests (OTP 51141) are left without the label since the original had no `NonHyperShiftHOST` marker.

e2e/cluster_api_machine_management.go
```diff
@@ -89,3 +94,140 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManage
...
+			for _, deploy := range deploys.Items {
+				annotations := deploy.Spec.Template.Annotations
+				Expect(annotations).To(HaveKeyWithValue(
+					"target.workload.openshift.io/management",
+					`{"effect": "PreferredDuringScheduling"}`,
+				), "deployment %s is missing the workload management annotation", deploy.Name)
+			}
+		})
+
+		It("should have IPAM CRDs installed", Label("Lifecycle:informing"), func() {
```
mdbooth: This test has some history.

The IPAM CRDs are required by MAPV (VSphere), so they are *always* installed, not just when the ClusterAPIMachineManagement FG is enabled. There's a special case to handle this, but it applies only when the feature gate is *not* enabled. See:

https://github.com/openshift/cluster-api/blob/303d9786a5017d299b6e7fc702bb92f5cb4550cf/openshift/manifests/0000_30_cluster-api_04_crd.core-cluster-api.yaml#L9

We should move this test to a separate block which runs unconditionally, i.e. with or without the FG. We should also add a comment explaining that the test can and should be removed some time after the FG is promoted and this special case has been removed.

pmeida: Done.

mdbooth (commented):
/approve
/lgtm

PR #633, opened 2026-07-28
OCPCLOUD-3647: Consolidate install-time object processing in toBoxcutterRevision

mdbooth:
This refactor forms the basis of a simplification to enable several other
features:
* Integration of CRD CompatibilityRequirement generation into capi-installer
* Consolidation of install-time transformations (e.g. TLS/Proxy Env vars) in
  one place
* Simplification of RenderedRevision

## Summary by CodeRabbit

* **Improvements**
  * Installer revisions now reconcile more reliably and report resource-resolution issues clearly.
  * Rendered components present resources consistently, with CRDs grouped into dedicated installation phases when applicable.
  * Existing-resource adoption annotations are validated and honored correctly, including collision protection for “always” adoption.
* **Refactor**
  * Simplified resource handling across rendered components and installation workflows.
* **Tests**
  * Expanded coverage for phase creation, resource grouping, repeatability, and adoption behavior.

### DIFF
--- a/pkg/controllers/installer/boxcutter.go
+++ b/pkg/controllers/installer/boxcutter.go
@@ -17,6 +17,8 @@ limitations under the License.
 package installer
 
 import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"pkg.package-operator.run/boxcutter"
 	"pkg.package-operator.run/boxcutter/probing"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -26,6 +28,16 @@ import (
 )
 
 // toBoxcutterRevision converts an InstallerRevision to a boxcutter.Revision.
+// Each component in an InstallerRevision contains a parsed collection of
+// objects with no further ordering. toBoxcutterRevision creates a Revision
+// containing the following phases for each of these components:
+// - a phase for the component's CRDs, if any
+// - a phase for the component's remaining objects, if any
+//
+// Although this is not yet the case, toBoxcutterRevision is intended to capture
+// all processing required only at installation time. Ideally InstallerRevision
+// will contain only parsed versions of exactly what was in provider manifests
+// without any further processing. That processing should be done here.
 func toBoxcutterRevision(installerRevision revisiongenerator.InstallerRevision) boxcutter.Revision {
 	probeOpts := util.SliceMap(allProbes(), func(p *probing.GroupKindSelector) boxcutter.PhaseReconcileOption {
 		return boxcutter.WithProbe(boxcutter.ProgressProbeType, p)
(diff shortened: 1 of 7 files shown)


pkg/revisiongenerator/revision_test.go
```diff
@@ -310,8 +310,7 @@ func TestForInstall(t *testing.T) {
 
 		components := installer.Components()
 		g.Expect(components).To(HaveLen(1))
-		g.Expect(components[0].CRDs()).To(HaveLen(1))
-		g.Expect(components[0].Objects()).To(HaveLen(1))
+		g.Expect(components[0].Objects()).To(HaveLen(2))
```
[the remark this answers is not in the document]
mdbooth: This is because CRDs no longer have special handling. I don't have a separate test for Leases, ConfigMaps, EndpointSlices, or ReplicaSets, either.

mdbooth:
Infra flakes.

/retest-required

mdbooth:
Filed OCPBUGS-100155 for the flake.

mdbooth:
Infra flake

/test regression-clusterinfra-aws-ipi-techpreview-capi

damdo (commented):
/approve

pkg/controllers/installer/revision_reconciler.go
```diff
@@ -122,53 +112,43 @@ func newRevisionReconciler(installerController *InstallerController, log logr.Lo
...
 
-		return convertedRevision{
-			revision:      rev,
-			conversionErr: err,
-		}
-	})
+	isComplete, messages, errs := r.reconcileRevisions(ctx, apiRevisions)
 
 	// Resolve collected objects to final ObjectReferences with correct plurals
 	if err := r.resolveCollectedObjects(); err != nil {
```
theobarberbany, line 135: Was the change from early-return to append intentional? Previously a `resolveCollectedObjects` failure was terminal, the reconciler bailed immediately, returned no completed revision name, and discarded reconciliation results. 

Now it appends to `errs` and may return `(revisionName=X, errs=[resolveError])` where it previously returned `(nil, nil, [resolveError])`. If intentional, might be worth a comment explaining why resolution failure should no longer block.

(I think this is because we don't collect and pre compute all revisions up front any longer, and may happen mid reconcile)

pkg/controllers/installer/boxcutter.go
```diff
@@ -25,56 +27,53 @@ import (
...
-}
-
-// GetRevisionNumber returns the revision number of the revision.
-func (r boxcutterRevision) GetRevisionNumber() int64 {
-	return r.revision.RevisionIndex()
-}
-
-// GetPhases returns the phases of the revision.
-func (r boxcutterRevision) GetPhases() []boxcutter.Phase {
+// toBoxcutterRevision converts an InstallerRevision to a boxcutter.Revision.
```
theobarberbany, line 30: nit: comment restates the function name. The interesting behaviour is that it splits CRDs into a dedicated `-crds` phase per component, attaches probes and adopt-existing options, and calls `collectObjects` for GVK/relatedObject registration.

I might also include why we've done this all here too - so what the motivation was for the design ending up like this.

PR #640, opened 2026-07-31
OCPBUGS-100316: Change failurePolicy to Ignore on MAPI-guarding VAPs …

miyadav:
…with CAPI paramKind.

 looks good to me as immediate fix for unblocking tp installs on vsphere 
  When the CAPI CRDs (cluster.x-k8s.io/v1beta2) are not yet established during install, the apiserver cannot resolve the paramKind and the VAP is "misconfigured". With `failurePolicy: Fail`, this denies every MAPI machine update — including status patches from machine-api-controllers, which are normally exempted via matchConditions but never reach that
evaluation. Workers clone successfully in vCenter but the machinecontroller can never persist the provider task ID, leaving all workers stuck in Provisioning and failing the install.

  Affected files:
  - authoritative-api-transition-requires-capi-infrastructure-ready.yaml
  - machine-api-machine-vap.yaml
  - machine-api-machine-set-vap.yaml
  - only-create-mapi-machine-if-authoritative-api-capi.yaml
  - prevent-authoritative-mapi-machineset-create-when-capi-exists.yaml
  
  generated by *claude*
  
  I understand there might be a reason we set them as `Fail` earlier , so we may have to dig more for a way to establish that CRDs are established and then we change it to fail ( that might need more in depth analysis ) 

## Summary by CodeRabbit

* **Behavior Changes**
  * Updated several admission policies to continue processing requests when policy evaluation cannot be completed.
  * Applies to machine and machine set validation, MAPI creation safeguards, and authoritative API transition checks.
  * These checks now fail open instead of blocking requests when an evaluation error occurs.

### DIFF
--- a/admission-policies/default/authoritative-api-transition-requires-capi-infrastructure-ready.yaml
+++ b/admission-policies/default/authoritative-api-transition-requires-capi-infrastructure-ready.yaml
@@ -25,7 +25,7 @@ kind: ValidatingAdmissionPolicy
 metadata:
   name: openshift-mapi-authoritative-api-transition-requires-capi-infrastructure-ready-and-not-deleting
 spec:
-  failurePolicy: Fail
+  failurePolicy: Ignore
 
   paramKind:
     apiVersion: cluster.x-k8s.io/v1beta2
--- a/admission-policies/default/machine-api-machine-set-vap.yaml
+++ b/admission-policies/default/machine-api-machine-set-vap.yaml
@@ -25,7 +25,7 @@ kind: ValidatingAdmissionPolicy
 metadata:
   name: machine-api-machine-set-vap
 spec:
-  failurePolicy: Fail
+  failurePolicy: Ignore
 
   paramKind:
     apiVersion: cluster.x-k8s.io/v1beta2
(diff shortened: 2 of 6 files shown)


theobarberbany (commented):
This changes these VAPs from fail-closed to fail-open permanently to solve what is a transient install-time ordering problem. Once CAPI CRDs are established, paramKind resolution works fine — the issue only exists during the window before the CRDs are created.

The risk with `failurePolicy: Ignore` is that these guardrails silently disappear any time paramKind can't be resolved — not just during install, but during upgrades, CRD issues, or any other disruption. These policies exist specifically to hard-block dangerous state transitions (MAPI mutations conflicting with CAPI, creating MAPI resources when CAPI is authoritative, etc.), so failing open defeats their purpose.

The right fix here is an ordering/dependency one — ensure the CAPI CRDs are established before these VAPs are applied. The `installOrder` mechanism already exists in this repo's manifests infrastructure and should be usable for this.

theobarberbany:
/close 

We already have ordering built in, the ordering is correct,  core CRDs 10, providers 20, VAPs 30. The logs provided in the CI run on the bug prove this. 

It looks like this ran at the same time the apiserver was rolling out, we had DNS issues, and the VAPs were created only about 10s before. In units 10s is enough time for the VAP controllers to warm up and enforce,  but in this instance it was not. 

The logs also show this self healed in 2s, and CAPV failed to retry after that error, hardening that logic is the actual fix here IMO.

see https://redhat.atlassian.net/browse/OCPBUGS-100316?focusedCommentId=17832993 for explanation @miyadav

PR #642, opened 2026-08-04
NO-JIRA: resolve placeholder version for k8s.io/autoscaler APIs

simkam:
Add replace directive to `go.work` to pin `k8s.io/autoscaler/cluster-autoscaler/apis` to a real version.

The cluster-autoscaler-operator dependency uses a Go placeholder version (`v0.0.0-00010101000000-000000000000`) with a [replace](https://github.com/openshift/cluster-autoscaler-operator/blob/main/go.mod#L8) in its own `go.mod`, but Go workspace mode does not propagate replace directives from transitive dependencies — only `go.work` replaces apply globally.

fixes https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/82669/rehearse-82669-pull-ci-openshift-cluster-capi-operator-main-verify-deps/2084559394309148672 from https://github.com/openshift/release/pull/82669


damdo (commented):
/approve
/lgtm

PR #644, opened 2026-08-08
OCPBUGS-104496: aws: use capa-controller-manager as service account

tthvo:
## Description

The service account for cluster-api-provider-aws (CAPA) is `capa-controller-manager` instead of `capi-controllers`.

The service account is used to construct the condition for IAM role trust policy. A mismatch will reject the pod's request to assume IAM role via AssumeRoleWithWebIdentity.

For reference, see CAPA openshift manifest: https://github.com/openshift/cluster-api-provider-aws/blob/1f5b17cc6969d6013a1abc5c6c62f00add809ceb/openshift/capi-operator-manifests/default/manifests.yaml#L201

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated AWS credential access to use the correct controller, improving cluster provisioning reliability.

### DIFF
--- a/manifests/0000_30_cluster-api_01_credentials-request.yaml
+++ b/manifests/0000_30_cluster-api_01_credentials-request.yaml
@@ -10,7 +10,7 @@ metadata:
     release.openshift.io/feature-gate: "ClusterAPIMachineManagement,ClusterAPIMachineManagementAWS"
 spec:
   serviceAccountNames:
-    - capi-controllers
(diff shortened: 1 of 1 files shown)


tthvo:
/verified by @tthvo and e2es

See also: https://github.com/openshift/cluster-api-provider-aws/pull/624#issuecomment-5225530413

damdo (commented):
/approve

PR #647, opened 2026-08-18
OCPBUGS-100155: fix flaky MachineSet sync test by atomically verifying InfraTemplate

simkam:
https://redhat.atlassian.net/browse/OCPBUGS-100155

The test that verifies `AWSMachineTemplate` rotation after a MAPI MachineSet `providerSpec` update was racy: it captured the new `InfrastructureRef` name in one `Eventually` block, then verified the template content in a separate block. Between these two steps, the sync controller could reconcile again and replace the intermediate template, leaving the test looking for a deleted object.

Fix by merging the two `Eventually` blocks into a single atomic poll that reads the current CAPI MachineSet, captures the `InfrastructureRef` name, and verifies the referenced template's `InstanceType` in the same iteration.

Also refresh the MAPI MachineSet from the API server before patching its `providerSpec` to avoid stale `ResourceVersion` conflicts.

Add an explicit `RequeueAfter: time.Second` when the `machinesetsync` controller is waiting for outdated `InfraMachineTemplates` to be deleted, providing a safety-net requeue instead of relying solely on watch events under CI load.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved machine set synchronization when replacing infrastructure templates.
  * Added a short delay while outdated templates are being deleted to prevent premature reconciliation.
  * Improved reliability when updating machine set provider specifications by refreshing the latest resource state first.
* **Tests**
  * Expanded migration validation to confirm the replacement infrastructure template and instance type.

### DIFF
--- a/e2e/machineset_migration_helpers.go
+++ b/e2e/machineset_migration_helpers.go
@@ -321,6 +321,11 @@ func updateAWSMachineSetProviderSpec(ctx context.Context, cl client.Client, mapi
 	GinkgoHelper()
 
 	By(fmt.Sprintf("Updating MachineSet %s providerSpec", mapiMachineSet.Name))
+
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(mapiMachineSet), mapiMachineSet); err != nil {
+		return fmt.Errorf("failed to refresh MachineSet %s: %w", mapiMachineSet.Name, err)
+	}
+
 	providerSpec := getAWSProviderSpecFromMachineSet(mapiMachineSet)
 	Expect(providerSpec).ToNot(BeNil(), "failed to extract AWS ProviderSpec from MachineSet %s", mapiMachineSet.Name)
 
--- a/e2e/machineset_migration_mapi_authoritative_test.go
+++ b/e2e/machineset_migration_mapi_authoritative_test.go
@@ -27,6 +27,7 @@ import (
 	capiframework "github.com/openshift/cluster-capi-operator/e2e/framework"
 	awsv1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
 )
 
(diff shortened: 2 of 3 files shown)


pkg/controllers/machinesetsync/machineset_sync_controller.go
```diff
@@ -361,7 +362,7 @@ func (r *MachineSetSyncReconciler) reconcileMAPIMachineSetToCAPIMachineSet(ctx c
 		return ctrl.Result{}, fmt.Errorf("unable to delete outdated Cluster API infrastructure machine templates: %w", err)
 	} else if shouldRequeue {
 		logger.Info("Waiting for Cluster API infrastructure machine templates to be deleted")
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: time.Second}, nil
```
mdbooth: Not a fan of this. There's no reason we should ever miss a watch event, even until intense load, even if the watch is cancelled and restarted at any point. This just hides bugs.

e2e/machineset_migration_helpers.go
```diff
@@ -321,6 +321,11 @@ func updateAWSMachineSetProviderSpec(ctx context.Context, cl client.Client, mapi
 	GinkgoHelper()
 
 	By(fmt.Sprintf("Updating MachineSet %s providerSpec", mapiMachineSet.Name))
+
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(mapiMachineSet), mapiMachineSet); err != nil {
+		return fmt.Errorf("failed to refresh MachineSet %s: %w", mapiMachineSet.Name, err)
+	}
```
mdbooth: Was this strictly required? If not I prefer to remove it. I think this could make tests harder to debug if the helper might not actually be using the same object state it was passed.

e2e/machineset_migration_mapi_authoritative_test.go
```diff
@@ -324,20 +325,23 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] Ma
...
-
-					Eventually(k.Object(newAWSMachineTemplate), capiframework.WaitShort, capiframework.RetryShort).Should(
-						HaveField("Spec.Template.Spec.InstanceType", Equal(newInstanceType)),
-					)
-				})
+					g.Expect(cl.Get(ctx, client.ObjectKeyFromObject(newAWSMachineTemplate), newAWSMachineTemplate)).To(Succeed())
+					g.Expect(newAWSMachineTemplate.Spec.Template.Spec.InstanceType).To(Equal(newInstanceType),
+						"AWSMachineTemplate %s should have InstanceType %s", currentTemplateName, newInstanceType)
+				}, capiframework.WaitMedium, capiframework.RetryMedium).Should(Succeed(),
+					"Should have a new InfraTemplate with InstanceType %s", newInstanceType)
```
mdbooth, line 344: 👍

mdbooth (changes requested):

pkg/controllers/machinesetsync/machineset_sync_controller.go
simkam: removed

e2e/machineset_migration_helpers.go
simkam: Right, it is not required, removed.

damdo:

Known to be facing issues (failure is at install time and unrelated)

PR #648, opened 2026-08-18
OCPCLOUD-3318: Integrate CompatibilityRequirements into the CCAPIO Insaller

theobarberbany:
## Summary by CodeRabbit

* **New Features**
  * Added compatibility checks for selected unmanaged custom resources before installing or updating revisions.
  * Installation now waits for required resources to be admitted and confirmed compatible.
  * Added permissions and lifecycle handling for compatibility requirements.

* **Bug Fixes**
  * Prevents incompatible revisions from replacing the currently active revision.
  * Reports errors when configured unmanaged resources are missing or cannot be converted.

* **Tests**
  * Added coverage for compatibility gating, filtering, failure handling, and successful installation scenarios.

* **Documentation**
  * Added guidance for writing synchronous and asynchronous Kubernetes tests.

### DIFF
--- a/.claude/skills/test-standards/SKILL.md
+++ b/.claude/skills/test-standards/SKILL.md
@@ -115,7 +115,62 @@ time.Sleep(5 * time.Second)
 Expect(machine.Status.Phase).To(Equal("Running"))
 ```
 
+## Ginkgo/Gomega patterns
+
+Use built-in features over custom implementations:
+- Use `DescribeTable` with `Entry` for table-driven tests instead of manual loops
+- Use `HaveField`, `HaveValue`, `HaveKey` for struct/map assertions instead of manual field checks
+- Use `ConsistOf` for unordered slice matching instead of sorting + `Equal`
+- Use `BeNumerically` for numeric comparisons instead of manual range checks
+
+### Async assertions with Komega
+
+Use Komega for Kubernetes object assertions:
+```go
+Eventually(k.Object(myResource)).Should(HaveField("ObjectMeta.ResourceVersion", Equal(expectedRV)))
+
+Eventually(k.UpdateStatus(myResource, func() {
+    myResource.Status.SomeField = "value"
+})).Should(Succeed())
+```
+
+### Resource management
+
+Use testutils resource builders for creating test objects:
+```go
+mapiMachine = mapiMachineBuilder.
+    WithNamespace(namespace).
+    WithName("foo").
+    WithAuthoritativeAPI(machinev1beta1.MachineAuthorityMachineAPI).
+    Build()
+```
+
+Standard cleanup in AfterEach:
+```go
+testutils.CleanupResources(Default, ctx, cfg, k8sClient, namespace,
+    &machinev1beta1.Machine{},
+    &clusterv1.Machine{},
+)
+```
+
+### Complex assertions
+
+Combine matchers with `SatisfyAll`:
+```go
+Eventually(komega.Object(resource)).Should(SatisfyAll(
+    HaveField("Status.AuthoritativeAPI", Equal(expected)),
+    HaveField("Status.SynchronizedGeneration", BeZero()),
+))
+```
+
+Nested field checks:
+```go
+HaveField("Status.Conditions", ContainElement(SatisfyAll(
+    HaveField("Type", Equal("Paused")),
+    HaveField("Status", Equal(corev1.ConditionTrue)),
+)))
+```
+
 ## General
 
-- Follow patterns from CLAUDE.md (Komega, DescribeTable, resource builders, testutils cleanup).
 - Remove `FIt`/`FContext` before committing.
--- a/pkg/controllers/installer/boxcutter.go
+++ b/pkg/controllers/installer/boxcutter.go
@@ -17,8 +17,11 @@ limitations under the License.
 package installer
 
 import (
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"pkg.package-operator.run/boxcutter"
 	"pkg.package-operator.run/boxcutter/probing"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -38,7 +41,11 @@ import (
 // all processing required only at installation time. Ideally InstallerRevision
 // will contain only parsed versions of exactly what was in provider manifests
 // without any further processing. That processing should be done here.
-func toBoxcutterRevision(installerRevision revisiongenerator.InstallerRevision, collectObjects func(obj *unstructured.Unstructured)) boxcutter.Revision {
+func toBoxcutterRevision(
+	installerRevision revisiongenerator.InstallerRevision,
+	collectObjects func(obj *unstructured.Unstructured),
+	unmanagedCRDs []string,
+) (boxcutter.Revision, error) {
 	probeOpts := util.SliceMap(allProbes(), func(p *probing.GroupKindSelector) boxcutter.PhaseReconcileOption {
 		return boxcutter.WithProbe(boxcutter.ProgressProbeType, p)
 	})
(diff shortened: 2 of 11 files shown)


.claude/skills/test-standards/SKILL.md
```diff
@@ -115,7 +115,62 @@ time.Sleep(5 * time.Second)
...
+    Build()
+```
+
+Standard cleanup in AfterEach:
+```go
+testutils.CleanupResources(Default, ctx, cfg, k8sClient, namespace,
+    &machinev1beta1.Machine{},
+    &clusterv1.Machine{},
+)
+```
```
mdbooth, line 154: FWIW I've been treating AfterEach as an anti-pattern generally. Prefer DeferCleanup where possible, which is almost always. DeferCleanup should run as soon as possible after the action which requires cleanup to avoid having to consider error paths, etc.

.claude/skills/test-standards/SKILL.md
mdbooth, line 1: Nice set of additional rules, just one note about AfterEach.

However, I don't get the benefit of this because I don't use Claude. I would still very much like to see this moved somewhere it can be referenced:
* By tools other than Claude
* By agents not using this skill

I would like to see all of this moved somewhere else which is referenced by AGENTS.md. The skill can retain its front-matter etc and also require the agent to immediately read the reference file when the skill fires. AGENTS.md can use the same front-matter when referencing the doc, tbh.

pkg/controllers/installer/boxcutter.go
```diff
@@ -57,14 +64,47 @@ func toBoxcutterRevision(installerRevision revisiongenerator.InstallerRevision,
...
+	var compatObjects []*unstructured.Unstructured
+
 	for _, component := range installerRevision.Components() {
-		var crds, objects []*unstructured.Unstructured
+		// Step 1: Transform — replace unmanaged CRDs with CompatibilityRequirements.
+		// Future transformations (proxy env vars, etc.) chain here before the split.
+		var transformed []*unstructured.Unstructured
 
 		for _, obj := range component.Objects() {
 			collectObjects(obj)
```
mdbooth, line 76: I was going to say this isn't correct because we might not emit the object. However, on reflection I do think this is correct in a way which isn't obvious and therefore probably needs a comment:

IIRC (but lets at least get a bot to check):
* collectObjects of a CRD results in the CRD being added to relatedObjects
* a CRD in relatedObjects causes must-gather to fetch all objects of that kind

We want to must-gather to collect all objects for our CRDs, even if we don't manage them.

HOWEVER, thinking about that even further, that coupling is nasty. We could comment it here, but I think the cleaner solution in code would be to:
* Only ever call collectObjects(obj) for objects which are actually omitted
* Teach the relatedObjects collector that CompatibilityRequirements should be handled like CRDs

I think this has a much lower chance of a future innocuous cleanup breaking relatedObjects.

This is my preference, but otherwise this should have a comment and a warning label attached.

pkg/controllers/installer/boxcutter.go
```diff
@@ -57,14 +64,47 @@ func toBoxcutterRevision(installerRevision revisiongenerator.InstallerRevision,
...
+					return nil, fmt.Errorf("building CompatibilityRequirement for CRD %s: %w", obj.GetName(), err)
+				}
+
+				labels := cr.GetLabels()
+				if labels == nil {
+					labels = make(map[string]string)
+				}
+
+				labels[revisiongenerator.ManagedLabelKey] = "compatibility-requirements"
+				cr.SetLabels(labels)
```
mdbooth, line 92: 👍This is nasty but required until we move that transformation here.

pkg/controllers/installer/boxcutter.go
```diff
@@ -75,11 +115,26 @@ func toBoxcutterRevision(installerRevision revisiongenerator.InstallerRevision,
 		addPhase(component.Name(), objects)
 	}
 
+	if unmanagedSet.Len() > 0 {
+		return nil, fmt.Errorf("unmanaged CRDs not found in any component: %v", sets.List(unmanagedSet))
+	}
```
mdbooth, line 120: We MUST remove this check. It is not only legitimate to have an unmangedCRD entry for a CRD which isn't present, it's a documented use case[^1] when preparing to update to a version which managed a CRD that previously didn't.

i.e. Version X+1 installs Foo, version X does not. I am running version X and have my own Foo. I don't want to deal with a conflict error on upgrade.
* While still running version X, add Foo to unmanaged CRDS <-- this would break
* Upgrade to version X+1 never attempts to manage Foo, no errors

[^1]: or at least I'm pretty sure it is 😅 If it's not, it needs to be.

pkg/controllers/installer/boxcutter.go
```diff
@@ -57,14 +64,47 @@ func toBoxcutterRevision(installerRevision revisiongenerator.InstallerRevision,
...
+				}
+
+				labels[revisiongenerator.ManagedLabelKey] = "compatibility-requirements"
+				cr.SetLabels(labels)
+
+				collectObjects(cr)
+				compatObjects = append(compatObjects, cr)
+
+				// Unmanaged CRD intentionally dropped — not installed, only checked for compatibility.
+				continue
```
mdbooth, line 98: As mentioned before I'm intending to expand the things that go here, e.g. adding the managed label, proxy env, adoption annotation, etc, etc. As mentioned above, I also think it's cleaner and safer to only call collectObjects once for objects which we will actually emit. Therefore, I think I'd restructure this a bit something like:

```
  for component in Components
    var allTransformed

    for object in component.Objects
      transformed := object
      if isCRD(object)
        ... transformed = CRD compat
        ... add managed label until this transformation is moved below
     
      ... insert other tranformations here ...
      collectObject(transformed)

      allTransformed += transformed

    for obj in allTransformed
      if isCompat
        ...
      else if isCRD
        ...
      else
        ...
```

pkg/controllers/installer/compatibility.go
```diff
@@ -0,0 +1,83 @@
...
+			},
+			ObjectSchemaValidation: apiextensionsv1alpha1.ObjectSchemaValidation{
+				Action: apiextensionsv1alpha1.CRDAdmitActionDeny,
+				NamespaceSelector: metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"kubernetes.io/metadata.name": capiNamespace,
+					},
+				},
+			},
+		},
```
mdbooth, line 74: 👍

pkg/controllers/installer/installer_controller.go
```diff
@@ -222,6 +222,11 @@ func (c *InstallerController) reconcile(ctx context.Context, log logr.Logger) op
 	}
 
 	revisionReconciler := newRevisionReconciler(c, log)
+
+	if clusterAPI.Spec != nil {
+		revisionReconciler.unmanagedCRDs = clusterAPI.Spec.UnmanagedCustomResourceDefinitions
+	}
```
mdbooth, line 228: This isn't correct. A revision's unmanagedCRDs are specified on the revision itself. It is the responsibility of the revision controller to add them to the revision and create a new revision if they change. Different revisions can therefore have different unmanagedCRDs, depending on when they were created.

pkg/controllers/installer/probes.go
```diff
@@ -31,6 +31,8 @@ func allProbes() []*probing.GroupKindSelector {
 	return []*probing.GroupKindSelector{
 		crdEstablishedProbe(),
 		deploymentAvailableProbe(),
+		compatibilityRequirementAdmittedProbe(),
+		compatibilityRequirementCompatibleProbe(),
```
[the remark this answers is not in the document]
mdbooth: I would have missed this. Feels like an opportunity for future subtle bugs, too. My preference would be to update probeSucceededPredicate instead to handle multiple probes for the same GVK.

pkg/controllers/installer/revision_reconciler.go
```diff
@@ -95,6 +95,7 @@ type collectedObjectRef struct {
 type revisionReconciler struct {
 	*InstallerController
 	log                   logr.Logger
+	unmanagedCRDs         []string
```
mdbooth, line 98: As mentioned elsewhere, this doesn't live here. It lives in RenderedRevision. We populate it from the APIRevision during conversion, and it comes *directly* from the specific API revision which is being converted, not from the spec.

If the revision controller isn't currently populating that we'll need to update the revision controller.

mdbooth (commented):
There are a couple of major things to address here, but despite that I think this is mostly there in terms of the work required.

I don't want to get hung up on my comments about the test-standards skill in this PR so lets handle that separately. I'm personally happy to approve this PR with or without the skill changes, and with or without my comments addressed.

The missing piece is an E2E. I'm happy to discuss whether we should add that in this PR or separately. My preference is this PR, otherwise we need to verify manually.

I think the E2E has to be disruptive. Also, because we disallow removing unmanagedCRDs it's *only* for ephemeral test clusters. That's not great, tbh. I'd like to see us:
* Mark a core CRD (e.g. Cluster) as unmanaged
* Fail to modify it in an incompatible way
* Successfully modify it in a compatible way

@JoelSpeed interested in your thoughts on irreversible e2e tests.

JoelSpeed:
> Also, because we disallow removing unmanagedCRDs it's only for ephemeral test clusters.

Do we actually prevent a user from deleting the entire resource? What happens if you `oc delete clusterapi cluster`? (is that the right resource?)

mdbooth:
> Do we actually prevent a user from deleting the entire resource? What happens if you oc delete clusterapi cluster? (is that the right resource?)

That's an interesting idea. I don't think so, but I can think of at least 1 way that it would break: the boxcutter revision number would reset, but existing resources would be tagged with a higher revision so wouldn't be updated. With some careful surgery it might be possible to fudge something, though. I'm thinking:

* Patch CVO to unmanage capi-installer
* Scale capi-installer to zero so revision controller doesn't stomp on us
* Delete CAPI config
* Create new CAPI object with status containing the most recent revision, but empty unmanagedCRDs in spec
* Patch CVO to manage capi-installer again
* Manually scale capi-installer to its previous value (to avoid waiting 10 minutes for CVO to do it)
* Wait for CAPI config to create a new revision and become stable

That's moderately nasty, but it would allow us to run an e2e.

pkg/controllers/installer/boxcutter.go
theobarberbany: Descoping the restructure to a follow-up PR and taking the comment-and-warning-label option here for now.

pkg/controllers/installer/boxcutter.go
theobarberbany: Descoping the restructure to a follow-up PR and taking the comment-and-warning-label option here for now.

pkg/revisiongenerator/revision.go
```diff
@@ -150,36 +156,65 @@ func substitutionsFromMap(m map[string]string) []operatorv1alpha1.ClusterAPIInst
...
+	// substitutions continued to be listed in the current revision.
+	if data, err := json.Marshal(r.substitutions); err == nil {
+		h.Write(data)
+	} else {
+		return "", fmt.Errorf("error marshalling substitutions: %w", err)
+	}
+
+	// Also include unmanagedCRDs for the same reasons.
+	if data, err := json.Marshal(r.unmanagedCRDs); err == nil {
+		h.Write(data)
```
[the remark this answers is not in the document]
theobarberbany: this is all techpreview, so IMO this is fine for now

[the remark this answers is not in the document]
theobarberbany: Not doing this.

Folding `unmanagedCRDs` into the content ID is deliberate, and for exactly the reason stated in the comment two lines above for `substitutions`: a revision whose unmanaged CRD set differs renders differently, so it must hash differently. A `len() > 0` guard would make two genuinely different recipes claim the same content ID.

The compatibility concern doesn't apply. `substitutions` already folds in unconditionally and has the same property, and this is gated behind `TechPreviewNoUpgrade`, so there is no upgrade path across the change. A revision whose recomputed content ID no longer matches the stored one is rejected by design and superseded by the next generated revision.

PR #656, opened 2026-08-25
OCPBUGS-115197: machinesync: preserve Synchronized condition lastTransitionTime in CAPI-to-MAPI sync

simkam:
https://redhat.atlassian.net/browse/OCPBUGS-115197

In the CAPI-to-MAPI success path, `convertedMAPIMachine` was passed to `applySynchronizedConditionWithPatch`. In steady-state reconciliations (no status or spec changes detected), `ensureMAPIMachineStatusUpdated` returns early before copying existing conditions into `convertedMAPIMachine`, leaving its `Status.Conditions` empty. This caused `SetLastTransitionTime` to treat the Synchronized condition as new and reset it to `Now()` on every sync.

Fix by using `existingMAPIMachine` (which carries the live conditions from the cluster) when it is non-nil, falling back to `convertedMAPIMachine` only for newly created machines where no prior condition exists.

Add integration tests for both MAPI-to-CAPI and CAPI-to-MAPI directions that confirm `lastTransitionTime` is preserved across successive successful reconciliations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved machine synchronization so successful reconciliations preserve the existing synchronization condition timestamp.
  * Ensured consistent behavior when either Machine API or Cluster API is authoritative.
  * New machines continue to receive the synchronization status as expected.

* **Tests**
  * Added coverage for preserving synchronization timestamps across repeated reconciliations.

### DIFF
--- a/pkg/controllers/machinesync/machine_sync_controller.go
+++ b/pkg/controllers/machinesync/machine_sync_controller.go
@@ -401,7 +401,16 @@ func (r *MachineSyncReconciler) reconcileCAPIMachinetoMAPIMachine(ctx context.Co
 		return result, createUpdateErr
 	}
 
-	return ctrl.Result{}, r.applySynchronizedConditionWithPatch(ctx, convertedMAPIMachine, corev1.ConditionTrue,
+	// Use existingMAPIMachine to preserve lastTransitionTime of the Synchronized condition
+	// across steady-state reconciliations (when no status changes are detected,
+	// ensureMAPIMachineStatusUpdated returns early without copying existing conditions
+	// into convertedMAPIMachine, so convertedMAPIMachine.Status.Conditions is empty).
+	mapiMachineForCondition := convertedMAPIMachine
+	if existingMAPIMachine != nil {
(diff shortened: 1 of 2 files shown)


damdo:
@simkam in the end, have you figured out if this is also need for machinesetsync? TY

simkam:
> @simkam in the end, have you figured out if this is also need for machinesetsync? TY

@damdo not needed, machineset's sync already passes `existingMAPIMachineSet` to `applySynchronizedConditionWithPatch` https://github.com/openshift/cluster-capi-operator/blob/5eb434a71302783200c32b306da8d4b312b7cf15/pkg/controllers/machinesetsync/machineset_sync_controller.go#L617

damdo (commented):
/approve
/lgtm

damdo:
/override ci/prow/e2e-aws-capi-disconnected-techpreview

Known to be broken in teardown phase

damdo:
Openstack failures are at cluster install time and unrelated to the MAPI 2 CAPI migration, which is proven to be still working by the other providers e2es. Overriding

PR #659, opened 2026-08-27
OCPBUGS-114670: fix(capi2mapi): normalize empty SSHKeyName to nil for MAPI AWS conversion

simkam:
https://redhat.atlassian.net/browse/OCPBUGS-114670

CAPI's AWSMachine.Spec.SSHKeyName is tri-state: nil means "use cluster default", `ptr("")` means "no SSH key", and a non-empty pointer means a named key. IPI clusters installed with CAPA default to `ptr("")`.

MAPI's AWS actuator has no such tri-state handling — any non-nil KeyName, including an empty string, is passed straight to AWS RunInstances and rejected with "Invalid value '' for keyPairNames".

Normalize `ptr("")` to nil in the conversion so MAPI omits the field, matching the "no SSH key" intent.

It is equivalent of test helper in https://github.com/openshift/cluster-api-actuator-pkg/pull/491/changes/63074080db99e16eceb85ccb601310aac3ba3d29#diff-77afc90f80bdb30e5eb1d917d9def6557a9aca2994b361936df2f7f49a41ee39R175

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved AWS SSH key configuration handling during conversion.
  - Empty SSH key names are now treated as unset, while valid key names and omitted values remain unchanged.
  - This keeps provider configuration consistent across conversion paths and prevents invalid empty values from being carried forward.

### DIFF
--- a/pkg/conversion/capi2mapi/aws.go
+++ b/pkg/conversion/capi2mapi/aws.go
@@ -167,7 +167,7 @@ func (m machineAndAWSMachineAndAWSCluster) toProviderSpec() (*mapiv1beta1.AWSMac
 		},
 		// UserDataSecret - Populated below.
 		// CredentialsSecret - Handled below.
-		KeyName: m.awsMachine.Spec.SSHKeyName,
+		KeyName: convertAWSSSHKeyNameToMAPI(m.awsMachine.Spec.SSHKeyName),
 		// DeviceIndex - OCPCLOUD-2707: Value must always be zero. No other values are valid in MAPA even though the value is configurable.
 		PublicIP:             m.awsMachine.Spec.PublicIP,
 		SecurityGroups:       convertAWSSecurityGroupstoMAPI(m.awsMachine.Spec.AdditionalSecurityGroups), // This is the way we want to convert security groups, as the AdditionalSecurity Groups are what gets added to MAPI SGs.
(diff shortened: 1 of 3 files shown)


e2e/machineset_migration_capi_authoritative_test.go
```diff
@@ -224,6 +225,72 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:MachineAPIMigration] Ma
...
+
+				By("Scaling MAPI MachineSet to 1 to create a machine from the synced providerSpec")
+				Expect(mapiframework.ScaleMachineSetWithContext(ctx, name, 1)).To(Succeed(), "should scale up MAPI MachineSet")
+
+				By("Verifying the MAPI machine reaches Running — a launch failure indicates a field mistranslation in the CAPI→MAPI sync")
+				mapiMachine, err := mapiframework.GetLatestMachineFromMachineSet(ctx, cl, mapiMachineSet)
+				Expect(err).ToNot(HaveOccurred(), "should get MAPI machine from MachineSet")
+				verifyMachineRunning(cl, mapiMachine)
+			})
+		})
```
damdo: I don't think we want an e2e for this thing. I'd say fuzzing/unit is enough. Let's drop this

simkam: ok, e2e dropped.

damdo (commented):
/approve
/lgtm

/verified by ci

PR #669, opened 2026-09-11
OCPCLOUD-1501: use projected tokens for management kubeconfig

damdo:
## Summary

- Add a revision-managed `<InfrastructureName>-kubeconfig` profile using projected ServiceAccount token and CA files.
- Add InfrastructureName substitution, installer ownership, scoped Secret tracking/RBAC, and consumer Pod access.
- Stage ownership transfer, remove the legacy kubeconfig controller and `capi-controllers-token`, and add repair/E2E coverage.
- Document the resulting in-cluster-only kubeconfig architecture.

## Summary by CodeRabbit

- **New Features**
  - Added management-cluster kubeconfig Secrets using projected ServiceAccount credentials and infrastructure-specific API endpoints.
  - Added automatic restoration when the kubeconfig Secret is modified or deleted.
  - Added dedicated permissions for kubeconfig consumers and management components.

- **Changes**
  - Kubeconfig Secret ownership now belongs to the revision installer.
  - Revision generation now waits for complete infrastructure readiness and includes the infrastructure name.

- **Documentation**
  - Updated controller, RBAC, and managed-resource documentation.
  - Removed legacy kubeconfig controller and token architecture documentation.


capi-operator-manifests/kubeconfig/manifests.yaml
```diff
@@ -0,0 +1,12 @@
...
+  value: YXBpVmVyc2lvbjogdjEKY2x1c3RlcnM6Ci0gbmFtZTogbWFuYWdlbWVudC1jbHVzdGVyCiAgY2x1c3RlcjoKICAgIHNlcnZlcjogaHR0cHM6Ly9rdWJlcm5ldGVzLmRlZmF1bHQuc3ZjOjQ0MwogICAgY2VydGlmaWNhdGUtYXV0aG9yaXR5OiAvdmFyL3J1bi9zZWNyZXRzL2t1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvY2EuY3J0CmNvbnRleHRzOgotIG5hbWU6IG1hbmFnZW1lbnQtY2x1c3RlcgogIGNvbnRleHQ6CiAgICBjbHVzdGVyOiBtYW5hZ2VtZW50LWNsdXN0ZXIKICAgIG5hbWVzcGFjZTogb3BlbnNoaWZ0LWNsdXN0ZXItYXBpCiAgICB1c2VyOiBzZXJ2aWNlLWFjY291bnQKY3VycmVudC1jb250ZXh0OiBtYW5hZ2VtZW50LWNsdXN0ZXIKdXNlcnM6Ci0gbmFtZTogc2VydmljZS1hY2NvdW50CiAgdXNlcjoKICAgIHRva2VuRmlsZTogL3Zhci9ydW4vc2VjcmV0cy9rdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3Rva2VuCg==
+kind: Secret
+metadata:
+  annotations:
+    capi-operator.openshift.io/adopt-existing: always
+  labels:
+    cluster.x-k8s.io/cluster-name: ${INFRASTRUCTURE_NAME}
+  name: ${INFRASTRUCTURE_NAME}-kubeconfig
+  namespace: openshift-cluster-api
+type: cluster.x-k8s.io/secret
```
damdo, line 13: We should add a comment explaining the content of this secret

ocp-manifests-input/kubeconfig/management-cluster-kubeconfig-secret.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: openshift-cluster-api
+  annotations:
+    capi-operator.openshift.io/adopt-existing: "always"
+  labels:
+    cluster.x-k8s.io/cluster-name: ${INFRASTRUCTURE_NAME}
+type: cluster.x-k8s.io/secret
+# The value is a static in-cluster kubeconfig. It contains no bearer token or
+# embedded CA; consumers use their own projected ServiceAccount files.
+data:
+  value: YXBpVmVyc2lvbjogdjEKY2x1c3RlcnM6Ci0gbmFtZTogbWFuYWdlbWVudC1jbHVzdGVyCiAgY2x1c3RlcjoKICAgIHNlcnZlcjogaHR0cHM6Ly9rdWJlcm5ldGVzLmRlZmF1bHQuc3ZjOjQ0MwogICAgY2VydGlmaWNhdGUtYXV0aG9yaXR5OiAvdmFyL3J1bi9zZWNyZXRzL2t1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvY2EuY3J0CmNvbnRleHRzOgotIG5hbWU6IG1hbmFnZW1lbnQtY2x1c3RlcgogIGNvbnRleHQ6CiAgICBjbHVzdGVyOiBtYW5hZ2VtZW50LWNsdXN0ZXIKICAgIG5hbWVzcGFjZTogb3BlbnNoaWZ0LWNsdXN0ZXItYXBpCiAgICB1c2VyOiBzZXJ2aWNlLWFjY291bnQKY3VycmVudC1jb250ZXh0OiBtYW5hZ2VtZW50LWNsdXN0ZXIKdXNlcnM6Ci0gbmFtZTogc2VydmljZS1hY2NvdW50CiAgdXNlcjoKICAgIHRva2VuRmlsZTogL3Zhci9ydW4vc2VjcmV0cy9rdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3Rva2VuCg==
```
mdbooth: We need to make this less scary. We could make it `stringData`. We could also use kustomize SecretGenerator and just have the content in a plain text file. Slight preference for the latter.

pkg/controllers/revision/revision_controller.go
```diff
@@ -320,8 +332,14 @@ func (r *RevisionController) SetupWithManager(mgr ctrl.Manager, tlsOptions []fun
...
-					// Only enqueue if the infrastructure is ready and was not ready before
-					return isInfrastructureReady(e.ObjectNew) && !isInfrastructureReady(e.ObjectOld)
+					oldInfra, oldOK := e.ObjectOld.(*configv1.Infrastructure)
+					newInfra, newOK := e.ObjectNew.(*configv1.Infrastructure)
+
+					if !oldOK || !newOK || !isInfrastructureReady(newInfra) {
+						return false
+					}
+					// Enqueue when becoming ready or when the infrastructure name changes.
+					return !isInfrastructureReady(oldInfra) || oldInfra.Status.InfrastructureName != newInfra.Status.InfrastructureName
```
mdbooth: It's not a lot of additional complexity, but I'd still omit it. This will never change.

manifests/0000_30_cluster-api_03_rbac_roles.yaml
```diff
@@ -54,6 +54,34 @@ rules:
...
+# access as the legacy integration controller.
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+  - patch
```
mdbooth: I'm pretty sure `patch` is not required here. I know we had it before, but don't think CAPI is patching Nodes. Can you please at least try without and see if it explodes?

I'm pretty sure openshift-capi-controllers no longer needs any nodes permissions. Could you remove those, too?

capi-operator-manifests/kubeconfig/manifests.yaml
mdbooth, line 1: How about moving this to the existing `default` profile? I think it's capi-operators which creates the `Cluster` object this is required for anyway, right?

pkg/controllers/installer/installer_controller.go
```diff
@@ -150,6 +152,9 @@ func setupTrackingCache(mgr ctrl.Manager) (managedcache.TrackingCache, error) {
 			Scheme:               mgr.GetScheme(),
 			DefaultLabelSelector: labels.NewSelector().Add(*managedByReq),
 			Mapper:               mgr.GetRESTMapper(),
+			ByObject: map[client.Object]cache.ByObject{
+				&corev1.Secret{}: {Namespaces: map[string]cache.Config{controllers.DefaultCAPINamespace: {}}},
+			},
```
mdbooth: I'd prefer we don't make exceptions here, even for Secret. Note that we already have `DefaultLabelSelector` so we're already watching only Secrets we created. The namespace restriction isn't required and may be surprising in the future.

mdbooth: In fact, lets add a comment here explaining why so the bots aren't tempted to do it.

ocp-manifests-input/kubeconfig/management-cluster-kubeconfig-secret.yaml
damdo: Agree we can use `stringData` 👍 
I can have a look at the latter too

capi-operator-manifests/kubeconfig/manifests.yaml
damdo: Can try that yes

pkg/controllers/installer/installer_controller.go
damdo: Ack will do both

ocp-manifests-input/kubeconfig/management-cluster-kubeconfig-secret.yaml
damdo: Done. Secret changed to use readable stringData instead of an opaque base64 data.value. I chose stringData rather than SecretGenerator because it keeps the exact `${INFRASTRUCTURE_NAME}-kubeconfig` name, type, labels, and adoption annotation without introducing new transformations

pkg/controllers/revision/revision_controller.go
damdo: Done

manifests/0000_30_cluster-api_03_rbac_roles.yaml
damdo: Done

capi-operator-manifests/kubeconfig/manifests.yaml
damdo: Done

pkg/controllers/installer/installer_controller.go
damdo: Done

damdo: Upon further tests/research, I’d say this is an ugly but desired exception rather than an unnecessary one.

The installer manager is cluster-scoped, and its tracking cache defaults to cluster-wide informers. It does not infer namespaces from generated manifests. The `ByObject` entry is what makes the Secret informer use the namespaced endpoint for `openshift-cluster-api`.

The only retained Secret from this profile is the CAPI management kubeconfig, explicitly created in `openshift-cluster-api`. Removing the restriction caused the installer to request a cluster-wide Secret watch and fail with:

```text
cannot watch resource "secrets" at the cluster scope
```

This occurred in [the AWS CAPI TechPreview run](https://prow.ci.openshift.org/view/gs/test-platform-results-public/pr-logs/pull/openshift_cluster-capi-operator/669/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-capi-techpreview/2100193939368710144), while waiting for the `cluster-api` ClusterOperator to become available.

Keeping the restriction uses the existing namespace-scoped permission and avoids granting `capi-installer` cluster-wide Secret watch access. For this Secret lifecycle, I’d say this ugly exception is the narrower and more accurate RBAC design.

docs/rbac.md
```diff
@@ -43,6 +43,14 @@ The `machine-api-migration` ServiceAccount runs the `machine-api-migration` bina
 
 This SA also binds to ClusterRole `system:openshift:openshift-cluster-api:read-tls-configuration` for APIServer TLS profile reading.
 
+## Revision installer and kubeconfig access
+
+The revision installer is the sole owner of the `<InfrastructureName>-kubeconfig` Secret in `openshift-cluster-api`. Its installer RBAC is intentionally limited to creating, updating, and repairing the generated Secret; it does not grant the installer the broad runtime permissions used by CAPI controllers. The Secret is consumed only by in-cluster CAPI/provider Pods.
```
mdbooth, line 48: This isn't correct. I'm generally uncomfortable in including the kubeconfig secret under any specific controller: it's just a capi-installer owned artifact, same as everything else capi-installer installs.

The secret itself is probably interesting enough to warrant its own doc section, but it's unrelated to controller logic.

docs/controllers/revision-installer.md
```diff
@@ -0,0 +1,21 @@
+# Revision installer
+
+## Overview
+
+The revision controller renders the resources for an installed CAPI revision. The [revision installer reconciler](../../pkg/controllers/installer/revision_reconciler.go) applies those resources and is the sole owner of the management kubeconfig Secret named `<InfrastructureName>-kubeconfig` in `openshift-cluster-api`.
+
+## Management kubeconfig
```
mdbooth, line 7: Same as the other comment: this doesn't live here. Perhaps we need a separate docs section for artifacts we install, but it's not controller logic. This will only cause confusion.

docs/controllers/revision-installer.md
```diff
@@ -0,0 +1,21 @@
...
+server: https://kubernetes.default.svc:443
+certificate-authority: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+tokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+```
+
+The paths resolve inside each consumer Pod. Consequently, each consumer authenticates as its own ServiceAccount and must have the runtime RBAC required by the client. Kubernetes automatically rotates the projected token file; no Secret rewrite or legacy token-rotation controller is required.
+
+## Ownership and repair
+
+During revision reconciliation, the installer adopts the expected Secret when necessary and updates it when the generated profile changes. It also repairs drift or a stale legacy form. No separate kubeconfig controller or shared `capi-controllers-token` is used. Installer RBAC is limited to managing the generated Secret and is separate from the broader permissions granted to runtime consumers.
```
mdbooth, line 21: This is just because it's a managed artifact, same as any other managed artifact. I wouldn't bother mentioning this at all.

e2e/cluster_api_machine_management.go
```diff
@@ -93,14 +96,80 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManage
...
+			Expect(cluster.CertificateAuthorityData).To(BeEmpty())
+			user := kubeconfig.AuthInfos["service-account"]
+			Expect(user).NotTo(BeNil())
+			Expect(user.TokenFile).To(Equal("/var/run/secrets/kubernetes.io/serviceaccount/token"))
+			Expect(user.Token).To(BeEmpty())
+			context := kubeconfig.Contexts["management-cluster"]
+			Expect(context).NotTo(BeNil())
+			Expect(context.Cluster).To(Equal("management-cluster"))
+			Expect(context.AuthInfo).To(Equal("service-account"))
+			Expect(kubeconfig.CurrentContext).To(Equal("management-cluster"))
```
mdbooth, line 127: I wouldn't personally test this. The interesting questions here are:
* Does it exist
* Does it work

This is basically a diff test. It's an installer-managed artifact, and we already have tests that the installer writes what it's supposed to.

e2e/cluster_api_machine_management.go
```diff
@@ -93,14 +96,80 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManage
...
+
+			By("waiting for the revision installer to recreate the kubeconfig")
+			Eventually(func() []byte {
+				current := &corev1.Secret{}
+				if err := cl.Get(ctx, key, current); err != nil {
+					return nil
+				}
+				return current.Data["value"]
+			}).WithTimeout(framework.WaitMedium).WithPolling(framework.RetryMedium).Should(Equal(expectedValue))
+		})
```
mdbooth, line 163: I think these duplicate installer tests. I don't think we need extra coverage for one resource given that it's an ordinary managed resource.

e2e/cluster_api_machine_management.go
```diff
@@ -93,14 +96,80 @@ var _ = Describe("[sig-cluster-lifecycle][OCPFeatureGate:ClusterAPIMachineManage
...
+
+		It("should not have the removed capi-controllers token Secret", func() {
```
mdbooth, line 165: What does this mean?

manifests-gen/customizations_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	cluster := kubeconfig.Clusters["management-cluster"]
+	if cluster == nil || cluster.Server != "https://kubernetes.default.svc:443" || cluster.CertificateAuthority != "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt" || cluster.CertificateAuthorityData != nil {
+		t.Fatalf("unexpected cluster configuration: %#v", cluster)
+	}
+	user := kubeconfig.AuthInfos["service-account"]
+	if user == nil || user.TokenFile != "/var/run/secrets/kubernetes.io/serviceaccount/token" || user.Token != "" {
+		t.Fatalf("unexpected user configuration: %#v", user)
+	}
+}
```
mdbooth, line 94: This is just a diff test. I would prefer to remove it.

manifests-gen/customizations_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+func TestProcessObjectsDoesNotMutateNamespaceBehavior(t *testing.T) {
```
mdbooth, line 96: Can we rename this? Maybe "TestProcessObjectsDoesNotEmitNamespace", or did I misunderstand?

manifests-gen/generate.go
```diff
@@ -139,6 +140,11 @@ func writeManifests(opts cmdlineOptions, resources []client.Object) (err error)
 				return fmt.Errorf("error writing separator to manifests file: %w", err)
 			}
 		}
+		if resource.GetObjectKind().GroupVersionKind().Kind == "Secret" && strings.HasSuffix(resource.GetName(), "-kubeconfig") {
+			if _, err := writer.Write([]byte("# This Secret contains only static connection data; consumers use their own projected ServiceAccount token and CA files.\n")); err != nil {
+				return fmt.Errorf("error writing kubeconfig Secret comment: %w", err)
+			}
+		}
```
mdbooth, line 147: I don't think we need this in the generated manifests, especially as the contents are in plain text and the comment is already in the source file. This just makes this function harder to read.

manifests/0000_30_cluster-api_04_rbac_bindings.yaml
```diff
@@ -19,6 +19,44 @@ subjects:
...
+    release.openshift.io/feature-gate: "ClusterAPIMachineManagement"
+    capability.openshift.io/name: ClusterAPI
+roleRef:
+  kind: ClusterRole
+  name: openshift-capi-manager-local-cluster
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  namespace: openshift-cluster-api
+  name: capi-manager
```
mdbooth, line 38: It feels weird to be adding RBAC here for `capi-manager`. How would you feel about moving this and the referenced Roles to the cluster-api repo?

In fact... is there any chance it has these perms already? This is standard CAPI behaviour, so I would not be surprised if the CAPI default RBAC already contained this. UPDATE: Seems like it does not.

ocp-manifests-input/default/management-cluster-kubeconfig-secret.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ${INFRASTRUCTURE_NAME}-kubeconfig
+  namespace: openshift-cluster-api
+  annotations:
+    capi-operator.openshift.io/adopt-existing: "always"
```
mdbooth, line 7: Are you expecting it to exist already? I'd prefer to leave this off if there's no definite reason to add it because it overrides a safety mechanism. It's intended for cases where capi-installer is adopting a resource that was previously managed by CVO (e.g. the IPAM CRDs). I don't expect it to be used often.

pkg/controllers/installer/installer_controller_test.go
```diff
@@ -141,6 +150,40 @@ var _ = Describe("InstallerController", Serial, func() {
 		}, defaultNodeTimeout)
 	})
 
+	Context("Secret Object Management", func() {
```
mdbooth, line 153: Any idea why it added these tests? Seems like Secret is just like any other resource.

pkg/controllers/revision/revision_controller_test.go
```diff
@@ -597,11 +605,13 @@ var _ = Describe("RevisionController manifest substitutions", Serial, func() {
 		Expect(updatedClusterAPI.Status.Revisions).To(HaveLen(1))
 
 		rev := updatedClusterAPI.Status.Revisions[0]
-		Expect(rev.ManifestSubstitutions).To(HaveLen(2))
-		Expect(rev.ManifestSubstitutions[0].Key).To(Equal("TLS_CIPHER_SUITES"))
-		Expect(*rev.ManifestSubstitutions[0].Value).To(Equal("TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"))
-		Expect(rev.ManifestSubstitutions[1].Key).To(Equal("TLS_MIN_VERSION"))
-		Expect(*rev.ManifestSubstitutions[1].Value).To(Equal("VersionTLS12"))
+		Expect(rev.ManifestSubstitutions).To(HaveLen(3))
```
mdbooth, line 608: Maybe update this to ConsistOf? I should have done this originally. Not super important as I hope we'll replace it soon anyway.

(But we won't actually replace it soon, so maybe...)

pkg/revisiongenerator/revision_test.go
```diff
@@ -595,6 +595,26 @@ func TestComponents(t *testing.T) {
...
+kind: Secret
+metadata:
+  name: ${INFRASTRUCTURE_NAME}-kubeconfig
+  labels:
+    cluster.x-k8s.io/cluster-name: ${INFRASTRUCTURE_NAME}
+type: Opaque
+`
+	rev := must(NewRenderedRevision([]providerimages.ProviderImageManifests{
+		profile(t, "core", "img1", "default", manifest),
+	}, WithManifestSubstitutions(map[string]string{"INFRASTRUCTURE_NAME": "test-infra"})))(g)
```
mdbooth, line 610: This is just testing manifest substitution. It's not even testing the codepath which populates the manifests, which would be more interesting. I'd drop this.

AGENTS.md
```diff
@@ -33,13 +33,13 @@ The repository also includes:
 
 #### capi-operator Controllers
 - **CAPI Installer Controller** (`pkg/controllers/capiinstaller/`) - Handles installation of CAPI components and providers
+- **Revision installer** (`pkg/controllers/installer/`) - Owns the in-cluster `<InfrastructureName>-kubeconfig` Secret, including adoption, update, and repair.
```
mdbooth, line 36: No it doesn't.

AGENTS.md
```diff
@@ -51,6 +51,12 @@ The repository also includes:
 - **MAPI to CAPI Conversion** (`pkg/conversion/mapi2capi/`) - Library implementing Conversion of MAPI resources to CAPI
 - **CAPI to MAPI Conversion** (`pkg/conversion/capi2mapi/`) - Library implementing conversion of CAPI resources to MAPI
 
+### Management kubeconfig
+
+The revision installer generates `<InfrastructureName>-kubeconfig` in `openshift-cluster-api`. This Secret is for CAPI and infrastructure-provider Pods running in the cluster, not for workstation use. It contains static connection data and references the consumer Pod's projected files: `/var/run/secrets/kubernetes.io/serviceaccount/token` (`tokenFile`) and `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` (`certificate-authority`). Kubernetes rotates the token file automatically, without changing the Secret.
```
mdbooth, line 56: No it doesn't.

AGENTS.md
```diff
@@ -51,6 +51,12 @@ The repository also includes:
...
+
+Consumers authenticate as their own ServiceAccounts and must have the required runtime RBAC. The installer is the sole owner and can adopt, update, or repair the Secret. The former kubeconfig controller, shared `capi-controllers-token`, and legacy token-rotation path are no longer part of the architecture.
```
mdbooth, line 58: Seems like this is duplicating the stuff in the controllers directory (which also needed updates). Can we just reference it here with some front-matter?

README.md
```diff
@@ -15,7 +15,7 @@ and is later reconciled by the upstream operator.
 represents current cluster, it is treated as management and workload cluster at the same time.
 - [InfrastructureCluster](https://cluster-api.sigs.k8s.io/developer/providers/cluster-infrastructure.html) - CAPI Infrastructure Cluster CR that represents the infrastructure cluster.
 - Worker userdata secret - a secret that contains ignition configuration to be used by the worker nodes.
-- Kubeconfig secret - a secret that contains kubeconfig for the cluster.
+- Management kubeconfig Secret - `<InfrastructureName>-kubeconfig` in `openshift-cluster-api`, consumed by CAPI components running in-cluster. The revision installer owns this Secret and stores only the API endpoint plus projected ServiceAccount `tokenFile` and CA paths; it is not a workstation-usable kubeconfig.
```
mdbooth, line 18: The first sentence is correct and useful! The second sentence is not correct.

e2e/cluster_api_machine_management.go
damdo: It means the legacy secret is gone. It tests the tombstone works, but it's not needed. Will remove

manifests-gen/generate.go
damdo: Now that it is, I agree with you. Dropped

ocp-manifests-input/default/management-cluster-kubeconfig-secret.yaml
damdo: Woops, missed this, yes you are right, it won't exist as it is INFRASTRUCTURE_NAME rather than the previously hardcoded and now tombstoned one. Removing..

manifests/0000_30_cluster-api_04_rbac_bindings.yaml
damdo: Moving it to o/cluster-api with: https://github.com/openshift/cluster-api/pull/310
Will keep these in as TODO remove for now, so we can test the whole thing again without 310, until 310 merges.

ocp-manifests-input/default/management-cluster-kubeconfig-secret.yaml
damdo: This is what is generated by the kubeconfig-controller, which we are taking out of service.
So in micro-upgrades it will be there, we need to adopt it somehow or IIUC it'll block the installer from applying the new one.

pkg/controllers/installer/installer_controller_test.go
damdo: I agree, no point in keeping that

AGENTS.md
damdo: Agreed let's keep it only there

ocp-manifests-input/default/management-cluster-kubeconfig-secret.yaml
mdbooth: Ack. In that case adopt-existing is correct here.

manifests/0000_30_cluster-api_04_rbac_bindings.yaml
damdo: Merged. So I removed the RBAC content from here, in favour of that

